PR #1, opened 2021-05-12
added librispeech pipeline helper

JackTemaki:
This is some code I wrote down for a pipeline idea, just to have discussion material.

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -1,0 +1,181 @@
+"""
+This file contains helper functions for the (common) pipeline steps needed to use the LibriSpeech corpus
+"""
+import os
+
+from recipe.i6_core.audio.encoding import BlissChangeEncodingJob
+from recipe.i6_core.corpus.transform import MergeCorporaJob, MergeStrategy
+from recipe.i6_core.datasets.librispeech import *
+from recipe.i6_core.meta.system import CorpusObject
+
+
+durations = {
+    "train-clean-100": 100.6,
+    "train-clean-360": 363.6,
+    "train-other-500": 496.7,
+    "dev-clean": 5.4,
+    "dev-other": 5.3,
+    "test-clean": 5.4,
+    "test-other": 5.1,
+}
+
+durations["train-clean-460"] = (
+    durations["train-clean-100"] + durations["train-clean-360"]
+)
+durations["train-other-960"] = (
+    durations["train-clean-460"] + durations["train-other-500"]
+)
+
+
+def get_bliss_corpus_dict(audio_format="flac", create_alias_with_prefix=None):
+    """
+    Download and create a bliss corpus for each of the LibriSpeech training corpora and test sets,
+    and return all corpora as a single corpus dict.
+
+    :param str audio_format: flac (no re-encoding), wav or ogg
+    :param str|None create_alias_with_prefix:
+    :return: A corpus dict with the following entries:
+        - 'dev-clean'
+        - 'dev-other'
+        - 'test-clean'
+        - 'test-other'
+        - 'train-clean-100'
+        - 'train-clean-360'
+        - 'train-clean-460'
+        - 'train-other-500'
+        - 'train-other-960'
+    :rtype: dict[str, Path]
+    """
+
+    create_alias_with_prefix = os.path.join(create_alias_with_prefix, "LibriSpeech")
+
+    download_metadata_job = DownloadLibriSpeechMetadataJob()
+    if create_alias_with_prefix:
+        download_metadata_job.add_alias(
+            os.path.join(create_alias_with_prefix, "download", "metadata")
+        )
+
+    def _get_corpus(corpus_name):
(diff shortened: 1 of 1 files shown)


common/datasets/librispeech.py
```diff
@@ -0,0 +1,181 @@
...
+        - 'test-other'
+        - 'train-clean-100'
+        - 'train-clean-360'
+        - 'train-clean-460'
+        - 'train-other-500'
+        - 'train-other-960'
+    :rtype: dict[str, Path]
+    """
+
+    create_alias_with_prefix = os.path.join(create_alias_with_prefix, "LibriSpeech")
```
mmz33, line 50: this will return an error if `create_alias_with_prefix` is `None.`

common/datasets/librispeech.py
```diff
@@ -0,0 +1,181 @@
...
+
+    download_metadata_job = DownloadLibriSpeechMetadataJob()
+    if create_alias_with_prefix:
```
mmz33, line 53: but what if i want to create an alias without a prefix? i think this should be possible

mmz33 (changes requested):

common/datasets/librispeech.py
JackTemaki: Ah yes, this is wrong.

common/datasets/librispeech.py
JackTemaki: you can pass ""

mmz33: ah ok i thought that we always want to create `alias` dir and thus when it is None we set it to `""`. But yeah ok so with `None` it will not create the alias dir.

JackTemaki: we could say we always want an alias... I would always use one... but maybe there are some people who do not want this

common/datasets/librispeech.py
```diff
@@ -0,0 +1,181 @@
...
+        "test-other",
+        "train-clean-100",
+        "train-clean-360",
+        "train-other-500",
+    ]
+
+    bliss_corpus_dict = {
+        corpus_name: _get_corpus(corpus_name) for corpus_name in corpus_names
+    }
+
```
Marvin84, line 86: Here, I am giving a general comment for the whole implementation, by pointing out one concrete example. I would re-design the whole thing as a class, with its attributes, e.g., `corpus_names`, `create_alias_with_prefix`, `audio_format_options`, and so on. And then I would define functions. Say, I am a user who wants to run this corpora creation code only for a subset of corpus names, then what I can do is just to set the attribute of the class and then run the function that does the steps as you are doing here. In your case, you have a local variable that can not be set. You might argue that people can start to use random strings here, my answer would be: we can introduce an `ENUM` class and give the possibility to the user to choose which corpora should be included and in case of merging to make sure they included the ones we need.

common/datasets/librispeech.py
```diff
@@ -0,0 +1,181 @@
...
+            os.path.join(create_alias_with_prefix, "download", "metadata")
+        )
+
+    def _get_corpus(corpus_name):
+        download_corpus_job = DownloadLibriSpeechCorpusJob(corpus_key=corpus_name)
+        create_bliss_corpus_job = LibriSpeechCreateBlissCorpusJob(
+            corpus_folder=download_corpus_job.out_corpus_folder,
+            speaker_metadata=download_metadata_job.out_speakers,
+        )
+        if create_alias_with_prefix:
```
Marvin84, line 64: The variable name `create_alias_with_prefix` is misleading. It is a path, but the name reminds me of a boolean. And also Shouldn't we do a more specific check here? Just as an example, checking that it is not `None` and it has the substring `LibriSpeech` in it?

Marvin84 (commented):
Variable name and a general refactoring comment. See details in the code.

common/datasets/librispeech.py
JackTemaki: I do not understand, "LibriSpeech" is always added, and it does check for `None`

common/datasets/librispeech.py
christophmluscher: Maybe I understand the code incorrectly, but in the functions there is no `tk.register_output`, so although the sisyphus graph would include the required corpus creation steps and ignore the others, so it should not be necessary for the user to specify which corpora should be created.. or is this wrong?

Marvin84: Concretely, you are not wrong. I was just giving a design suggestion that is more flexible on the user side. Take it or leave it.

christophmluscher: ah ok that is true.
I have no strong feelings on this :)

JackTemaki:
@christophmluscher suggested to push this now so that we can create the jobs and start working with this. The current discussions have no effect on the Sisyphus graph, so we can easily make changes and fix this later on.

christophmluscher (approved):

mmz33 (approved):

PR #4, opened 2021-06-14
implements first version of GmmClass

christophmluscher:
This is mainly to have some discussions on this class. This is by no means finished. Please treat it this way :)


albertz:
I thought we don't need PRs for the users directory?


Christoph M. Lüscher ***@***.***> schrieb am Mo., 14. Juni
2021, 17:45:

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+                self.gt_features(c, **feat_args["gt"])
+                self.energy_features(c, **feat_args["energy"])
+            for t in self.train_corpora:
+                self.add_energy_to_features(t, "gt")
+
+        if "fb" in feat_args.keys():
+            for c in corpus_list:
+                self.fb_features(c, **feat_args["fb"])
+
+        unknown_features = set(feat_args.keys()).difference(
```
michelwi, line 208: Maybe we can allow for a generic feature type when the extraction flow is fully given. (E.g. NN-Bottleneck features)

JackTemaki:
> I thought we don't need PRs for the users directory?

I think this is just so that we get the notification and start discussing about it, because at some point this code should be shared.

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+            **kwargs,
+        )
+        self.jobs[corpus]["train_{}".format(name)].selected_alignment_jobs[
+            -1
+        ].add_alias("train_{}_alnt_last".format(name))
+        self.jobs[corpus]["train_{}".format(name)].selected_mixture_jobs[-1].add_alias(
+            "train_{}_mix_last".format(name)
+        )
+        tk.register_output(
+            "train_{}_{}_alnt_bundle_last".format(corpus, name),
```
michelwi, line 263: I don't like "alnt" as abbreviation for alignment and would prefer 'align' or full 'alignment'. But that's purely cosmetically.

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+        :param gmm_args: parameters for the different Gmm-HMM steps
+        :param train_data: dict(str: GmmInput)
+        :param dev_data: dict(str: GmmInput)
+        :param test_data: dict(str: GmmInput)
+        :return:
+        """
+        self.gmm_args = gmm_args
+        self._init_am(**gmm_args.am_args)
+        for name, v in sorted(train_data.items()):
+            self.corpora[name] = v.corpus_object
```
JackTemaki, line 110: This could be a separate function which could also work as external call to add corpora that are not yet in train or test.

so a `def add_corpus(name: str, corpus: meta.system.Corpus, add_lm: bool)`

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+
+                for dev_c in self.dev_corpora:
+                    pass
+
+                for tst_c in self.test_corpora:
+                    pass
+
+        # ---------- CaRT ----------
+        if "cart" in steps:
+            for c in self.train_corpora:
```
michelwi, line 742: Is estimating a separate CART for each training corpus the intended bahavior? I would discuss that we want the same cart for every corpus.

In fact, maybe we want to merge all (or a variable selection of all) training corpora for each training step.

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+            self._init_lm(name, **v.lm)
+            self.test_corpora.append(name)
+        self.cart_questions = gmm_args.cart_questions
+
+    @tk.block()
+    def _init_am(self, **kwargs):
+        self.crp["base"].acoustic_model_config = am.acoustic_model_config(**kwargs)
+
+    @tk.block()
+    def _init_corpora(self, name):
```
JackTemaki, line 138: I think this should be singular, as this is only for one corpus

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+        ##################################################
+        :param monophone_training_args:
+          'feature_flow': 'mfcc+deriv+norm',
+          'feature_energy_flow': 'energy,mfcc+deriv+norm',
+          'align_iter': 75,
+          'splits': 10,
+          'accs_per_split': 2,
+        :param monophone_recognition_args:
+          'eval_iter': [7, 8, 9, 10]
+          'pronunciation_scales': [10.0, 11.0, 12.0, 13.0, 14.0]
```
michelwi, line 154: Do we maybe want to set the pronunciation scale to 1.0?

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+                          'feature_flow': dev_corpus_name,
+                          'pronunciation_scale': pronunciation_scale,
+                          'lm_scale': lm_scale,
+                          'lm_lookahead': True,
+                          'lookahead_options': None,
+                          'create_lattice': True,
+                          'eval_single_best': True,
+                          'eval_best_in_lattice': True,
+                          'search_parameters': {
+                            'beam_pruning': 18.0,
```
michelwi, line 166: Please do NOT suggest 18 here. @curufinwe can provide a better value. (My guess would be 16 for "good" and 12 for "fast")

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+                            'beam-pruning-limit': 100000,
+                            'word-end-pruning': 0.75,
```
michelwi, line 168: Please do NOT suggest 0.75 here. @curufinwe can provide a better value. (My guess would be 0.5)

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+                            'word-end-pruning-limit': 15000
+                          },
+                          'best_path_algo': 'bellman-ford',  # options: bellman-ford, dijkstra
+                          'fill_empty_segments': False,
+                          'scorer': recog.Sclite,
+                          'scorer_args': {'ref': create_corpora.stm_files['dev-other']},
+                          'scorer_hyp_args': "hyp",
+                          'rtf': 50,
```
michelwi, line 176: with reasonable prunning settings then the rtf value can be decreased.

We should also add the temporal AM lookahead approximation value.

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+          early_recombination=False,
+          tdp_scale=1.0,
+          tdp_transition=(3.0, 0.0, 30.0, 0.0),  # loop, forward, skip, exit
+          tdp_silence=(0.0, 3.0, "infinity", 20.0),
+          tying_type="global",
+          nonword_phones="",
+          tdp_nonword=(0.0, 3.0, "infinity", 6.0)  # only used when tying_type = global-and-nonword
+        }
+        ##################################################
+        :param feature_extraction_args:
```
michelwi, line 316: maybe no need to repeat suggested defaults from above

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+          'recog_args': {
+                          'feature_flow': dev_corpus_name,
+                          'pronunciation_scale': pronunciation_scale,
+                          'lm_scale': lm_scale,
+                          'lm_lookahead': True,
+                          'lookahead_options': None,
+                          'create_lattice': True,
+                          'eval_single_best': True,
+                          'eval_best_in_lattice': True,
+                          'search_parameters': {
```
michelwi, line 394: please update or hide pruning settings before anybody sees them.

users/luescher/setups/hybrid/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .gmm_system import *
+from .util import *
```
albertz, line 2: Newline missing at end of file.

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+        gmm_args: GmmPipelineArgs,
+        train_data: Dict[str, GmmDataInput],
+        dev_data: Dict[str, GmmDataInput],
+        test_data: Dict[str, GmmDataInput],
+    ):
+        """
+        :param gmm_args: parameters for the different Gmm-HMM steps
+        :param train_data: dict(str: GmmInput)
+        :param dev_data: dict(str: GmmInput)
+        :param test_data: dict(str: GmmInput)
```
albertz, line 104: What docstring style is this supposed to be? The standard (PEP sth?) style would be:
```suggestion
        :param dict[str, GmmInput] train_data:
        :param dict[str, GmmInput] dev_data:
        :param dict[str, GmmInput] test_data:
```

users/luescher/setups/hybrid/util.py
```diff
@@ -0,0 +1,421 @@
...
+        :param lm: file: Path, type: str, scale: float
+        :param concurrent: concurrency for gmm hmm pipeline
+        """
+        self.corpus_object = corpus_object
+        self.lexicon = lexicon
+        self.lm = lm
+        self.concurrent = concurrent
+
+
+class GmmPipelineArgs:
```
JackTemaki, line 26: Maybe we also want to have each specific args block as own class? I think with this we could document this easier and split things a little bit more...

JackTemaki:
@christophmluscher can you move this PR to a branch in the main repo? Then we could also make edits, and you do not have to do this all alone.

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+        self.crp[name].language_model_config.file = file
+        self.crp[name].language_model_config.scale = scale
+
+    @tk.block()
+    def _init_lexica(self, name, file, normalize_pronunciation, **kwargs):
+        self.crp[name].lexicon_config = rasr.RasrConfig()
+        self.crp[name].lexicon_config.file = file
+        self.crp[name].lexicon_config.normalize_pronunciation = normalize_pronunciation
+
+    def allow_zero_weights(self, name: str = "base"):
```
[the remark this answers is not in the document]
christophmluscher: This can also be done with the training args for the individual training steps.. `{mono,tri,...}_args`
`init_am` should also be possible I think.. not 100% sure.. maybe @curufinwe knows?
so remove completely?

users/luescher/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,845 @@
...
+            },
+        )
+        self.feature_flows[corpus][name] = features.basic_cache_flow(feature_path)
+        self.feature_flows[corpus][f"uncached_{name}"] = feature_job.feature_flow
+
+    @tk.block()
+    def extract_features(self, feat_args: dict, **kwargs):
+        corpus_list = self.train_corpora + self.dev_corpora + self.test_corpora
+
+        if "mfcc" in feat_args.keys():
```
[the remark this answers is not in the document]
christophmluscher: I thought we always use the "+energy" in the GMM Pipeline when using MFCC, that is why I did not separate. This is more a query than an objection. I think it would be good to handle this consistent throughout the different SystemClasses (GMM, HybridNN, etc)

users/luescher/setups/hybrid/gmm_system.py
christophmluscher: so just add "generic"?
wouldn't the `extract_features` function just be extended by the user? or how do you want to handle this? Have a call to the FeatureExtractionJob?

users/luescher/setups/hybrid/util.py
christophmluscher: so 14? as a trade-off?

users/luescher/setups/hybrid/util.py
christophmluscher: I agree. What granularity would you suggest?
Basically mirroring the `run` function? `mono_train_args`, `mono_recog_args`, ...
since `recog_args` can be reused for different steps!?

christophmluscher:
> @christophmluscher can you move this PR to a branch in the main repo? Then we could also make edits, and you do not have to do this all alone.

this means just pushing to main repo!? not sure how this is done...

users/luescher/setups/hybrid/gmm_system.py
michelwi: I would instead of throwing an error if `unknown_features` are non-empty, loop over these and then call a `self.generic_features(c, feature_name, **feat_args[feature_name])` function (we can expand meta.system) that creates a generic feature extraction job.

the MfccJob is only a convenience function anyway.

users/luescher/setups/hybrid/gmm_system.py
christophmluscher: This is not what I wanted. If you have all corpora and audio files available you would merge all available corpora via MergeCorporaJob. Then you would set the merged single corpus file as a starting point. So you would actually only have one corpus within the class.
Later on, you might want to add corpora (for speed perturbation or TTS) and then redo for example the cart step. But this is not reflected in the "default' run function...

users/luescher/setups/hybrid/gmm_system.py
christophmluscher: I changed the pipeline for the feature extraction. It now allows also for generic feature extraction.

christophmluscher:
ToDos
- remove `allow_zero_weights`
- update search params (before merge!!!)
- add temporal AM lookahead

users/luescher/setups/hybrid/gmm_system.py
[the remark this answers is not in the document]
christophmluscher: merged into `_init_am` function. moved from `acoustic_model_post_config` to `acoustic_model_config`, so has to be set correctly from the start, otherwise set `allow_zero_weights` via `{mono,tri,...}_args`

common/setups/hybrid/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .gmm_system import *
+from .util import *
```
albertz, line 2: Is this really on purpose to have all symbols in this namespace?
Also, as you import all symbols from `sisyphus` below, those will also end up here.
(Just asking. I don't like this too much but if you think this is ok, then just leave it.)

common/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,878 @@
+__all__ = ["GmmSystem"]
+
+import itertools
+import sys
+from typing import Dict
+
+# -------------------- Sisyphus --------------------
+
+from sisyphus import *
```
albertz, line 9: Such imports of all symbols (`from <...> import *`) are generally not recommended.
But if this is standard code style here, or you think this is fine, then leave it.

common/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,878 @@
...
+        """
+        :param gmm_init_args: parameters for am and feature extraction
+        :param gmm_monophone_args: parameters for lin align and monophone
+        :param gmm_triphone_args: parameters for cart and triphone
+        :param gmm_vtln_args: parameters for VTLN
+        :param gmm_sat_args: parameters for SAT
+        :param gmm_vtln_sat_args: parameters for VTLN+SAT
+        :param dict[str, GmmInput] train_data:
+        :param dict[str, GmmInput] dev_data:
+        :param dict[str, GmmInput] test_data:
```
albertz, line 127: This is inconsistent to the type documentation above (`GmmDataInput` vs `GmmInput`), so sth is wrong.

But also, I don't think we need to have such information redundantly. I would suggest to only use type annotations.

Also, I don't really like documentation which doesn't say anything, like `gmm_vtln_args: parameters for VTLN`. Just leave it away, which reduces the time to read it, and thus makes it more clear in the end.

Also, if the type is already documented via type annotations, I think you don't need to mention the parameter at all in the docstring anymore (unless you want to really add further information/documentation for it).
(In RETURNN, we want to have all parameters listed, because this is always the way we document types, because RETURNN doesn't use type annotations yet, because Python 2 compatibility.)

common/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,878 @@
...
+
+    @tk.block()
+    def _init_lm(self, name, file, type, scale, **kwargs):
+        self.crp[name].language_model_config = rasr.RasrConfig()
+        self.crp[name].language_model_config.type = type
+        self.crp[name].language_model_config.file = file
+        self.crp[name].language_model_config.scale = scale
+
+    @tk.block()
+    def _init_lexicon(self, name, file, normalize_pronunciation, **kwargs):
```
albertz, line 183: I would prefer to more consistently document all types in parameters.

I assume `name` is `str` (this is most often the case).

I assume `normalize_pronunciation` is `bool`, although I'm not 100% sure.

But I have no idea what `file` is. Is this `io.TextIO` (accepting strings), or `io.RawIO` (accepting bytes), or a filename, or maybe a `Path`, or sth else? (Btw, for filenames, I usually would prefer `filename` as an argument name, to make that more clear.)

common/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,878 @@
...
+        self.crp[name].lexicon_config = rasr.RasrConfig()
+        self.crp[name].lexicon_config.file = file
+        self.crp[name].lexicon_config.normalize_pronunciation = normalize_pronunciation
+
+    def _add_features(self, name, corpus, feature_job, prefix=""):
```
albertz, line 188: Same comment about types. E.g. what is `corpus` and `feature_job`?

common/setups/hybrid/util.py
```diff
@@ -0,0 +1,399 @@
...
+class GmmInitArgs:
+    def __init__(
+        self,
+        costa_args: dict = None,
+        am_args: dict = None,
+        feature_extraction_args: dict = None,
+    ):
+        """
+        ##################################################
+        :param costa_args"
```
albertz, line 59:
```suggestion
        :param costa_args
```
?

common/setups/hybrid/util.py
```diff
@@ -0,0 +1,399 @@
...
+              'warping_function': "mel",
+              'filter_width': 268.258,  # 80
+              'normalize': True,
+              'normalization_options': None,
+              'without_samples': False,
+              'samples_options': {'audio_format': "wav",
+                                  'dc_detection': True,},
+              'cepstrum_options': {'normalize': False,
+                                   'outputs': 16,
+                                   'add_epsilon': False,},
```
albertz, line 96: I don't really like these hanging indents.
(But it's just the documentation, so maybe doesn't matter...)

common/setups/hybrid/util.py
```diff
@@ -0,0 +1,399 @@
...
+                                'dc_detection': True},
+            'fft_options': None,
+            'apply_log': True,
+            'add_epsilon': False,
+          }
+          'energy':
+          {
+              'without_samples': False,
+              'samples_options': {'audio_format': "wav",
+                          'dc_detection': True},
```
albertz, line 139: Indentation is completely off.

common/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,878 @@
...
+                    rtf=rtf,
+                    mem=mem,
+                    parallelize_conversion=parallelize_conversion,
+                    lattice_to_ctm_kwargs=lattice_to_ctm_kwargs,
+                    **kwargs,
+                )
+
+    # -------------------- run setup  --------------------
+
+    def run(self, steps=tuple("all")):
```
albertz, line 734: It is generally recommended to not use `tuple` explicitly but instead the specific syntax for it, like `('a', 'l', 'l')`. But actually, is this what you want here? I assume you want `("all",)` instead. So:
```suggestion
    def run(self, steps=("all",)):
```

Also, I miss some documentation here. What steps are allowed?

common/setups/hybrid/gmm_system.py
```diff
@@ -0,0 +1,878 @@
...
+        assert len(steps) > 0
+        if len(steps) == 1 and steps[0] == "all":
```
albertz, line 736: This can be simplified:
```suggestion
        if list(steps) == ["all"]:
```
Which I find more readable.

Or:
```suggestion
        if tuple(steps) == ("all",):
```

common/setups/hybrid/util.py
```diff
@@ -0,0 +1,399 @@
...
+    "GmmVtlnArgs",
+    "GmmSatArgs",
+    "GmmVtlnSatArgs",
+    "NnArgs",
+]
+
+import i6_core.meta as meta
+
+
+class GmmDataInput:
```
albertz, line 15: This class is not documented.

common/setups/hybrid/util.py
```diff
@@ -0,0 +1,399 @@
...
+extra_config = sprint.SprintConfig()
+extra_config.allow_zero_weights = True
+{accumulate,split,align}_extra_args = {'extra_config': extra_config}
+
+'{accumulate,split,align}_extra_rqmt': {'mem': 10, 'time': 8},
+
+vtln align time = 8
+
+if not using the run function -> name and corpus almost always to be added
+"""
```
albertz, line 47: What does this documentation belong to? Usually you would set it as a docstring to the module, some class or function.

And otherwise you would use comments (`#`) and not a string.

common/setups/hybrid/util.py
```diff
@@ -0,0 +1,399 @@
...
+        self.sat_recognition_args = sat_recognition_args
+        self.sdm_vtln_args = sdm_vtln_args
+        self.vtln_sat_training_args = vtln_sat_training_args
+        self.vtln_sat_recognition_args = vtln_sat_recognition_args
+
+
+class NnArgs:
+    def __init__(
+        self,
+        nn_ce_training_args: dict = None,
```
albertz, line 374: The type is wrong, if `None` is valid. It would be `Optional[dict]`.

albertz (approved):
Some generic comments. Decide for yourself what you use from those.
Also one small bug which should be fixed (`tuple("all") == ('a', 'l', 'l')`).
But otherwise, looks good to me, from a quick glance.

common/setups/hybrid/__init__.py
JackTemaki: as long as `__all__` is set correctly in each file this is fine I guess.

christophmluscher: this has been the default behaviour in combination with the`__all__`.

I am open to discussing this, but I would make this separate from this pull request and maybe do this in `i6_core`.

PR #6, opened 2021-07-09
restructure and fixes

christophmluscher:


common/setups/hybrid/base_system.py
```diff
@@ -0,0 +1,165 @@
...
+    - sat training
+    - sat recognition
+    - vtln+sat training
+    - vtln+sat recognition
+    """
+
+    def __init__(self):
+        super().__init__()
+
+        self.crp["base"].python_home = gs.RASR_PYTHON_HOME
```
michelwi, line 61: These settings are not necessary strictly speaking. Esp. if the RASR is compiled with system python it is fine to not set them and to not have `gs.RASR_PYTHON_HOME` set. I suggest to add somthing like
```if hasattr(gs, "SPRINT_PYTHON_HOME") and gs.SPRINT_PYTHON_HOME is not None:```

(or maybe we want to set `None` in this case?

File: common/setups/hybrid/base_system.py [hunk 2]
```diff
@@ -0,0 +1,165 @@
...
+    - corpora: name and i6_core.meta.system.Corpus
+    - lexicon
+    - lm
+
+    settings needed:
+    - am
+    - lm
+    - lexicon
+    - feature extraction
+    - linear alignement
```
michelwi, line 45: some of these settings are not needed in base class but in derived class `GmmSystem`

common/setups/hybrid/nn_system.py
```diff
@@ -0,0 +1,167 @@
...
+Path = tk.setup_path(__package__)
+
+# -------------------- System --------------------
+
+
+class NnSystem(BaseSystem):
+    """
+    - 5 corpora types: train, devtrain, cv, dev and test
+    - two training data settings: defined in returnn config or not
+    - 3 different types of decoding: returnn, rasr, rasr-wei
```
michelwi, line 35: maybe `rasr-wei` should be called rasr-label-sync in the public recipes

common/setups/hybrid/nn_system.py
```diff
@@ -0,0 +1,167 @@
...
+    - 3 different lm: count, lstm, trafo
```
michelwi, line 36: I would only differentiate 2 kinds: count, neural

where neural includes FFNN, LSTM, Trafo, ...

(edit:) See comment below, I think this distinction can/should be dropped.

common/setups/hybrid/nn_system.py
```diff
@@ -0,0 +1,167 @@
...
+# -------------------- Init --------------------
+
+Path = tk.setup_path(__package__)
+
+# -------------------- System --------------------
+
+
+class NnSystem(BaseSystem):
+    """
+    - 5 corpora types: train, devtrain, cv, dev and test
```
michelwi, line 33: `devtrain` not explained in comment

common/setups/hybrid/nn_system.py
```diff
@@ -0,0 +1,167 @@
...
+        self._init_am(**self.hybrid_init_args.am_args)
+
+        self._assert_corpus_name_unique(
+            train_data, devtrain_data, cv_data, dev_data, test_data
+        )
+
+        for name, v in sorted(train_data.items()):
+            self.add_corpus(name, data=v, add_lm=False)
+            self.train_corpora.append(name)
+            assert self.concurrent[name] == 1, "concurrency for train data should be 1"
```
michelwi, line 100: Why? For sequence training we would want to process the train data in parallel/

common/setups/hybrid/nn_system.py
```diff
@@ -0,0 +1,167 @@
...
+            feature_corpus=feature_corpus,
+            train_corpus=train_corpus,
+            dev_corpus=dev_corpus,
+            feature_flow=feature_flow,
+            alignment=alignment,
+            num_classes=num_classes,
+            returnn_config=returnn_config,
+        )
+
+    # -------------------- Count LM Recognition --------------------
```
michelwi, line 149: Maybe we don't  define separate steps for "count recognition", "LSTM recognition" etc. but allow for an arbitrary number of recognitions with different settings.

Different settings (as specified by a dict, or data object) can be other type of LM, can be other prior (or -scale), ...

common/setups/hybrid/base_system.py
christophmluscher: good point.

the default is None

common/setups/hybrid/nn_system.py
christophmluscher: but for NN training it should be, right? only for the lattice generation it should have higher concurrency. so turn into warning or drop completely?

common/setups/hybrid/nn_system.py
christophmluscher: Yes, I like that

common/setups/hybrid/nn_system.py
michelwi: depends on where/how this concurrency setting is used. I see two options:

a) force this to be 1 here and read it wherever needed. When higher concurrency is needed (e.g. lattices) then we create an overlay.
b) allow arbitrary number here and wherever a sequential processing (segment file for nn training) is required, hardcode a 1. and otherwise use the specified value

not sure which is the best way.

christophmluscher: I think I like b) better.. :D

common/setups/hybrid/base_system.py
```diff
@@ -42,24 +42,14 @@ class BaseSystem(meta.System):
...
     """
 
     def __init__(self):
         super().__init__()
 
-        self.crp["base"].python_home = gs.RASR_PYTHON_HOME
-        self.crp["base"].python_program_name = gs.RASR_PYTHON_EXE
+        if hasattr(gs, "RASR_PYTHON_HOME") and gs.RASR_PYTHON_HOME is not None:
+            self.crp["base"].python_home = gs.RASR_PYTHON_HOME
+            self.crp["base"].python_program_name = gs.RASR_PYTHON_EXE
```
michelwi, line 52: one might assume that setting `gs.RASR_PYTHON_HOME` but not setting `gs.RASR_PYTHON_EXE`  is a case that will never occur, but maybe check each separately?

JackTemaki (changes requested):

common/setups/hybrid/util.py
```diff
@@ -57,7 +57,7 @@ def __init__(
 # if not using the run function -> name and corpus almost always to be added
 
 
-class GmmInitArgs:
+class HybridInitArgs:
```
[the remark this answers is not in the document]
michelwi: Well CTC can also be interpreted as a hybrid HMM system with a special, simple hidden Markov model topology. 
Transducer.. here you are right.

But the name `RasrSystem` would then clash with the `GmmSystem` which also heavily uses RASR.

PR #9, opened 2021-07-30
added librispeech ARPA lm, lexicon and vocab

JackTemaki:


common/datasets/librispeech.py
```diff
@@ -255,14 +266,153 @@ def get_ogg_zip_dict(create_alias_with_prefix=None):
...
-    For i6-users: physical jobs are located in: `/work/common/asr/librispeech/data/sisyphus_work_dir/`
+    :return: the lexicon with special lemmas and phonemes
+    :rtype: lexicon.Lexicon
+    """
+    lex = lexicon.Lexicon()
+    lex.add_lemma(
+        lexicon.Lemma(
+            orth=["[SILENCE]", ""],
+            phon=["[SILENCE]"],
+            synt=[""],
```
michelwi, line 314: synt should be list of list 
(here it would work technically, but should nevertheless be consistent)

common/datasets/librispeech.py
```diff
@@ -255,14 +266,153 @@ def get_ogg_zip_dict(create_alias_with_prefix=None):
...
+            special="silence",
+            eval=[""],
```
michelwi, line 316: eval should be list of list 
(here it would work technically, but should nevertheless be consistent)

common/datasets/librispeech.py
mmz33: you could also leave it None as by default it will be an empty list.

mmz33 (approved):

common/datasets/librispeech.py
JackTemaki: list of list would give the wrong results, because I want to write `<synt/>`into the file, not `<synt><tok/></synth>` but `synt=[None]` might be more correct

JackTemaki: The above comment was meant for synth, but is of course valid here as well...

JackTemaki: And it was not my idea to write it like that, I took this from the `GraphemicLexiconFromWordListJob` Job. But well, just because it exist does not mean it is a good idea :D

christophmluscher (approved):

common/datasets/librispeech.py
```diff
@@ -8,21 +8,28 @@
...
+ - "3gram" for the non-pruned 3-gram LM
+ - "4gram" for the non-pruned 4-gram LM
 
 If you want to use other subsets (especially with .ogg zips),
 please consider to use segment lists to avoid creating new corpus files.
 
-For i6-users: physical jobs are located in: `/work/common/asr/librispeech/data/sisyphus_work_dir/`
+All alias and output paths will be under: ``<path_prefix>/LibriSpeech/....``
+
+For i6-users: physical jobs generate via the "export" functions
```
vieting, line 31:
```suggestion
For i6-users: physical jobs generated via the "export" functions
```

common/datasets/librispeech.py
```diff
@@ -255,14 +266,153 @@ def get_ogg_zip_dict(create_alias_with_prefix=None):
     return ogg_zip_dict
 
 
-def export_all_datasets(path_prefix):
+def get_arpa_lm_dict(create_alias_with_path_prefix=None):
+    """
+    Download the ARPA language models from OpenSLR,
+    valid keys are: "3gram" and "4gram".
+
+    :param create_alias_with_path_prefix:
```
vieting, line 274: I don't find it completely intuitive from the name what this argument does and what I should pass for it, so you might consider describing it here.

common/datasets/librispeech.py
```diff
@@ -255,14 +266,153 @@ def get_ogg_zip_dict(create_alias_with_prefix=None):
...
+    lm_dict["3gram"] = download_arpa_3gram_lm_job.out_file
+
+    if create_alias_with_path_prefix:
+        lm_prefix = os.path.join(create_alias_with_path_prefix, "LibriSpeech", "lm")
+        download_arpa_3gram_lm_job.add_alias(os.path.join(lm_prefix, "download_3gram_lm_job"))
+        download_arpa_4gram_lm_job.add_alias(os.path.join(lm_prefix, "download_4gram_lm_job"))
+    return lm_dict
+
+
+def get_static_lexicon():
```
vieting, line 300: Is "static" the best term here? The name could also make clear that it's about special lemmas/phonemes.

common/datasets/librispeech.py
```diff
@@ -255,14 +266,153 @@ def get_ogg_zip_dict(create_alias_with_prefix=None):
...
+        target_filename="librispeech-lexicon.txt",
+        checksum="d722bc29908cd338ae738edd70f61826a6fca29aaa704a9493f0006773f79d71")
+
+    convert_lexicon_job = LexiconFromTextFileJob(
+        text_file=download_lexicon_job.out_file, compressed=True
+    )
+
+    merge_lexicon_job = MergeLexiconJob(
+        bliss_lexica=[static_lexicon_job.out_bliss_lexicon,
+                      convert_lexicon_job.out_bliss_lexicon],
```
vieting, line 371: Is this consistent with the black style?

common/datasets/librispeech.py
```diff
@@ -255,14 +266,153 @@ def get_ogg_zip_dict(create_alias_with_prefix=None):
...
+        sort_lemmata=False,
+        compressed=True
+    )
 
+    if create_alias_with_path_prefix:
+        alias_path = os.path.join(
+            create_alias_with_path_prefix,
+            "LibriSpeech",
+            "lexicon"
+        )
```
vieting, line 382: I guess this fits in one line?

michelwi (approved):

vieting (approved):

PR #10, opened 2021-07-30
LJSpeech split g2p, dynamic lexicon and static lexicon

JackTemaki:


Atticus1806 (approved):

michelwi (commented):
the phonemes `'#'`, `'.'`, (and `+ - @ { }` ) should be avoided otherwise there might be problems in reading allophone files (esp. important for hybrid systems). You can use it with a different phon label and keep orth and synth.

JackTemaki:
> the phonemes `'#'`, `'.'`, (and `+ - @ { }` ) should be avoided otherwise there might be problems in reading allophone files (esp. important for hybrid systems). You can use it with a different phon label and keep orth and synth.

The question is if I really care about this here, as this lexicon will never get even close to RASR :D The lexicon will only be used for my TTS related G2P process. But yes, I could add `[PLACEHOLDER]` `[SPACE]` `[DOT]`  and so on instead...

@Atticus1806 what do you think? Then we would just need to change the RETURNN vocabulary accordingly... The only downside is that it will be weird if people look at a TTS only setup and wonder why we converted punctuation in such a way...

michelwi:
> The question is if I really care about this here, as this lexicon will never get even close to RASR

At some point we might be doing ASR-integraded pronunciation prediction. (@christophmluscher :P) and then it might be awkward to change the lexicon later or have a separate lexicon for ASR and TTS. (Or people 5 years from now do not know about this and might just use your lexicon)

But for the foreseeable future this should definitively be fine. Therefore this is just a comment and not a request for change.

Atticus1806:
>Then we would just need to change the RETURNN vocabulary accordingly... 

I guess it really is a tradeoff between what effort changing it now and the chance of it being used in a context where this becomes relevant combined with the effort with changing it then. In general I would argue for "ignoring until it happens" and not change it, but since we are currently restructuring / cleaning up anyways this might be worth to consider.

JackTemaki:
I will do something new here!

PR #12, opened 2021-08-05
refactor recognition

JackTemaki:
Moved some logic into `recognition`, streamlined some parameters.

### DIFF
--- a/common/setups/hybrid/gmm_system.py
+++ b/common/setups/hybrid/gmm_system.py
@@ -595,36 +595,44 @@ class GmmSystem(RasrSystem):
         self,
         name: str,
         iters: List[int],
+        lm_scales: List[int],
+        feature_scorer: Tuple[str, str],
+        optimize_am_lm_scale: bool,
+        # parameters just for passing through
         corpus: str,
         feature_flow: str,
-        feature_scorer,
-        pronunciation_scale: List[int],
-        lm_scale: List[int],
-        search_params: dict,
+        pronunciation_scale: int,
+        search_parameters: dict,
         rtf: int,
         mem: int,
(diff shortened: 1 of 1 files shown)


common/setups/hybrid/gmm_system.py
```diff
@@ -595,36 +595,44 @@ def recognition(
...
-                )
-            )
             recog_func = self.recog_and_optimize if optimize_am_lm_scale else self.recog
 
-            for it in iters:
+            for it, l in itertools.product(iters, lm_scales):
                 recog_func(
-                    name=f"{name}-{it:02d}",
-                    prefix=f"recog_{corpus}_{name}/",
+                    name=f"iter{it:02d}-lm{l}",
```
michelwi, line 628: `pronunciation_scale` if set, does not go into the alias/output name.

common/setups/hybrid/gmm_system.py
```diff
@@ -595,36 +595,44 @@ def recognition(
         self,
         name: str,
         iters: List[int],
+        lm_scales: List[int],
```
michelwi, line 598: all scales (and rtf) can be float

michelwi (changes requested):

common/setups/hybrid/gmm_system.py
JackTemaki: I did this on purpose, because if we add pronunciation_scale, we could discuss other stuff as well. I only added iteration and lm here, because this is stuff that is handled separately here. Also at the top I added this "only for passing through" comment. So everything behind there should be named specifically by the user.

michelwi (approved):

christophmluscher (approved):

PR #13, opened 2021-08-05
librispeech get_bliss_corpus_dict small fix

vieting:
In the default case, `subdir_prefix` is set to `None` and later used in `os.path.join()`, which does not work.

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -86,11 +86,7 @@ def get_bliss_corpus_dict(audio_format="flac", subdir_prefix=""):
     """
     assert audio_format in ["flac", "ogg", "wav"]
 
-    subdir_prefix = (
-        os.path.join(subdir_prefix, "LibriSpeech")
-        if subdir_prefix
-        else None
-    )
+    subdir_prefix = os.path.join(subdir_prefix, "LibriSpeech")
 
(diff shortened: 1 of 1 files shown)


common/datasets/librispeech.py
```diff
@@ -89,7 +89,7 @@ def get_bliss_corpus_dict(audio_format="flac", subdir_prefix=""):
     subdir_prefix = (
         os.path.join(subdir_prefix, "LibriSpeech")
         if subdir_prefix
-        else None
+        else ""
```
JackTemaki, line 92: Both is wrong, this if clause is not needed at all now that string is the default ;)

JackTemaki (changes requested):

common/datasets/librispeech.py
vieting: If we discard the addition of `"LibriSpeech"` yes. So should I remove it?

vieting: Or do you want to have it in any case?

vieting: I think it's nicer in any case, see the force-pushed commit.

JackTemaki (approved):

common/datasets/librispeech.py
JackTemaki: Yes this is exactly what i meant, removing the condition and just adding it per default

PR #14, opened 2021-08-08
Add switchboard and lexicon data preparation helper jobs

mmz33:


common/datasets/switchboard.py
```diff
@@ -0,0 +1,198 @@
...
+                phon=[tag_to_phon[tag]],
+                synt=[[]],
+                eval=[[]],
+            )
+        )
+
+    # create special lemmas
+    lex.add_lemma(
+        lexicon.Lemma(
+            orth=["[SENTENCE-END]"], synt=[["</s>"]], special="sentence-boundary"
```
michelwi, line 83: any special reason this has no `eval=[[]]`?

mmz33: hmm not sure why. I checked again the current lexicon we use and it is without `eval`.

mmz33: the `sentence-end` lemma has the same `synt` token and there we have empty `eval` tag.

common/datasets/switchboard.py
```diff
@@ -0,0 +1,198 @@
...
+        sort_phonemes=True,
+        sort_lemmata=False,
+        compressed=True,
+    )
+    merge_lexicon_job.add_alias(os.path.join(alias_name, "merge_lexicon_job"))
+
+    return merge_lexicon_job.out_bliss_lexicon
+
+
+def _export_train_bliss_coprus(subdir_prefix=""):
```
JackTemaki, line 169: typo `coprus`

common/datasets/switchboard.py
```diff
@@ -0,0 +1,198 @@
...
+from i6_core.lexicon.modification import WriteLexiconJob, MergeLexiconJob
+from i6_core.lib import lexicon
+
+
+def get_speakers_list(subdir_prefix=""):
+    """
+    Returns speakers list
+
+    :param str subdir_prefix: alias name prefix
+    :rtype: tk.Path
```
JackTemaki, line 12: if the `rtype` is only `Path`, please also specify with `:return:` in a line directly above what kind of file is returned.

JackTemaki (changes requested):

JackTemaki (approved):

PR #15, opened 2021-08-10
Setup for augmentation of the lexicon with g2p based OOVs pronunciation

Marvin84:
Note the following:
1. We need to decide whether this needs an `export()` and if it should be added into the `export_all()`
2. The job is done for all different train corpora, starting with the original lexicon **with stress**
3. The lexicon without stress is still not considered

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -36,13 +36,19 @@ import os
 from sisyphus import tk
 
 from i6_core.audio.encoding import BlissChangeEncodingJob
+from i6_core.corpus.stats import ExtractOovWordsFromCorpusJob
 from i6_core.corpus.transform import MergeCorporaJob, MergeStrategy
 from i6_core.datasets.librispeech import *
+from i6_core.g2p.apply import ApplyG2PModelJob
+from i6_core.g2p.convert import BlissLexiconToG2PLexiconJob, G2POutputToBlissLexiconJob
+from i6_core.g2p.train import TrainG2PModelJob
 from i6_core.lib import lexicon
 from i6_core.lexicon.conversion import LexiconFromTextFileJob
 from i6_core.lexicon.modification import WriteLexiconJob, MergeLexiconJob
 from i6_core.meta.system import CorpusObject
 from i6_core.tools.download import DownloadJob
+
+from i6_experiments.common.helpers.g2p import *
 
 
 durations = {
@@ -371,6 +377,60 @@ def get_bliss_lexicon(subdir_prefix=""):
     merge_lexicon_job.add_alias(os.path.join(alias_path, "merge_lexicon_job"))
 
     return merge_lexicon_job.out_bliss_lexicon
+
+
+def get_g2p_augmented_bliss_lexicon_dict(subdir_prefix=""):
+    """
+    Given the original LibriSpeech bliss lexicon, it is possible to estimate the pronunciation for
+    out of vocabulary (OOV) words for each of the LbriSpeech training corpora. Here, we create a dictionary
+    that has different train corpora as keys and the corresponding g2p augmented bliss lexicon as values
+
+    :param st subdir_prefix:
+    :return: dictionary of Paths to augmented bliss_lexicon
+    :rtype: dict[str, Path]
+    """
+    alias_path = os.path.join(subdir_prefix, "LibriSpeech", "lexicon")
+    augmented_bliss_corpora = {}
+
+    original_bliss_lexicon =  get_bliss_lexicon(subdir_prefix=subdir_prefix)
+    g2p_args = get_g2p_parameters()
+
+    g2p_lexicon_job = BlissLexiconToG2PLexiconJob(bliss_lexicon=original_bliss_lexicon)
+
+    g2p_train_job = TrainG2PModelJob(
+        g2p_lexicon=g2p_lexicon_job.out_g2p_lexicon,
+        **g2p_args["train"])
+    g2p_train_job.add_alias(os.path.join(alias_path, "train_g2p_model"))
+
+    def get_g2p_augmented_bliss_lexicon(corpus_name, bliss_corpus):
+        extract_oov_job = ExtractOovWordsFromCorpusJob(
+            bliss_corpus=bliss_corpus,
+            bliss_lexicon=original_bliss_lexicon
+        )
+        extract_oov_job.add_alias(os.path.join(alias_path, "extract-oov-from-{}".format(corpus_name)))
+
+        g2p_apply_job = ApplyG2PModelJob(
+            g2p_model=g2p_train_job.out_best_model,
+            word_list_file=extract_oov_job.out_oov_words,
+            **g2p_args["apply"]
+        )
+        g2p_apply_job.add_alias(os.path.join(alias_path, "apply-g2p-for-{}".format(corpus_name)))
+
+        g2p_final_lex_job = G2POutputToBlissLexiconJob(
+            iv_bliss_lexicon=original_bliss_lexicon,
+            g2p_lexicon=g2p_apply_job.out_g2p_lexicon
+        )
+        g2p_final_lex_job.add_alias(os.path.join(alias_path, "g2p-output-to-bliss-{}".format(corpus_name)))
+
+        return g2p_final_lex_job.out_oov_lexicon
+
+
+    bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
+    for corpus_name, bliss_corpus in bliss_corpus_dict.items():
+        if "train" in corpus_name:
+            augmented_bliss_corpora[corpus_name] = get_g2p_augmented_bliss_lexicon(corpus_name, bliss_corpus)
+
+    return augmented_bliss_corpora
 
 
 def get_lm_vocab(subdir_prefix=""):
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -378,17 +378,17 @@ def get_bliss_lexicon(subdir_prefix=""):
 def get_g2p_augmented_bliss_lexicon_dict(subdir_prefix=""):
     """
     Given the original LibriSpeech bliss lexicon, it is possible to estimate the pronunciation for
-    out of vocabulary (OOV) words for each of the LbriSpeech training corpora. Here, we create a dictionary
+    out of vocabulary (OOV) words for each of the LibriSpeech training corpora. Here, we create a dictionary
     that has different train corpora as keys and the corresponding g2p augmented bliss lexicon as values
 
-    :param st subdir_prefix:
+    :param str subdir_prefix:
     :return: dictionary of Paths to augmented bliss_lexicon
     :rtype: dict[str, Path]
     """
     alias_path = os.path.join(subdir_prefix, "LibriSpeech", "lexicon")
     augmented_bliss_corpora = {}
 
-    original_bliss_lexicon =  get_bliss_lexicon(subdir_prefix=subdir_prefix)
+    original_bliss_lexicon = get_bliss_lexicon(subdir_prefix=subdir_prefix)
     g2p_augmenter = G2PBasedOovAugmenter(original_bliss_lexicon=original_bliss_lexicon)
 
     bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -386,7 +386,7 @@ def get_g2p_augmented_bliss_lexicon_dict(subdir_prefix=""):
     :rtype: dict[str, Path]
     """
     alias_path = os.path.join(subdir_prefix, "LibriSpeech", "lexicon")
-    augmented_bliss_corpora = {}
+    augmented_bliss_lexica = {}
 
     original_bliss_lexicon = get_bliss_lexicon(subdir_prefix=subdir_prefix)
     g2p_augmenter = G2PBasedOovAugmenter(original_bliss_lexicon=original_bliss_lexicon)
@@ -394,13 +394,13 @@ def get_g2p_augmented_bliss_lexicon_dict(subdir_prefix=""):
     bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
     for corpus_name, bliss_corpus in bliss_corpus_dict.items():
         if "train" in corpus_name:
-            augmented_bliss_corpora[corpus_name] = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
+            augmented_bliss_lexica[corpus_name] = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
                 bliss_corpus=bliss_corpus,
                 corpus_name=corpus_name,
                 alias_path=alias_path
             )
 
-    return augmented_bliss_corpora
+    return augmented_bliss_lexica
 
 
 def get_lm_vocab(subdir_prefix=""):
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -36,19 +36,15 @@ import os
 from sisyphus import tk
 
 from i6_core.audio.encoding import BlissChangeEncodingJob
-from i6_core.corpus.stats import ExtractOovWordsFromCorpusJob
 from i6_core.corpus.transform import MergeCorporaJob, MergeStrategy
 from i6_core.datasets.librispeech import *
-from i6_core.g2p.apply import ApplyG2PModelJob
-from i6_core.g2p.convert import BlissLexiconToG2PLexiconJob, G2POutputToBlissLexiconJob
-from i6_core.g2p.train import TrainG2PModelJob
 from i6_core.lib import lexicon
 from i6_core.lexicon.conversion import LexiconFromTextFileJob
 from i6_core.lexicon.modification import WriteLexiconJob, MergeLexiconJob
 from i6_core.meta.system import CorpusObject
 from i6_core.tools.download import DownloadJob
 
-from i6_experiments.common.helpers.g2p import *
+from i6_experiments.common.helpers.g2p import G2PBasedOovAugmenter
 
 
 durations = {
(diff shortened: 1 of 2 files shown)


common/helpers/g2p.py
```diff
@@ -0,0 +1,18 @@
+
+def get_g2p_parameters():
```
JackTemaki, line 2: It seems I misunderstood your idea of having a g2p helper. I thought that you want to do a generic "pipeline" here, that can also be used for switchboard, and then call this pipeline for each librispeech or switchboard corpus with parameters that are passed from the outside.

Because why would you have the default args here again? You could remove this and get exactly the same pipeline.

JackTemaki (changes requested):

common/helpers/g2p.py
JackTemaki: Yes that would be good, because otherwise point (2) would be difficult. Do you think it makes sense to have a helper for training as well (so the conversion and train job)?

Marvin84: > Do you think it makes sense to have a helper for training

Definitely, yes.

michelwi: +1 from my side for a generic g2p pipeline helper. 
The helper pipeline then would then get the g2p_args dict (so the user can easily configure settings) together with a train lexicon, a seed lexicon and a OOV list to add to the train lex.

Then in the librispeech part you would call something like
```
g2p_args = get_g2p_parameters()
train_lex = get_bliss_lexicon()
seed_lex = train_lex

for corpus_name, bliss_corpus in bliss_corpus_dict.items():
    if "train" in corpus_name:
        oov_file = ExtractOovWordsFromCorpusJob(bliss_corpus, seed_lex)
        seed_lex = my_g2p_pipeline(
            train_lex=train_lex,
            seed_lex=seed_lex,
            oov_file=oov_file,
            **g2p_args)
        augmented_bliss_corpora[corpus_name] = seed_lex
````
i.e. always the same lexicon is used to train the g2p model s.t. Sisyphus will reuse the Job.
and the `seed_lex` is successively extended by the oovs of the respective next larger corpus.

**Review by michelwi — COMMENTED** [review]
Concerning you initial questions:
In my opinion, we should always use the non-stressed version of the lexicon and we should always include all training OOV's. Further, we should use the non-stressed lexicon to train the g2p model.

Then I am in favour of exporting the non-stressed-all-OOV-added lexicon as the canonical lexicon for Librispeech.
(I guess it wouldn't hurt to have also the 960's OOV in the 100h train lexicon)

**Comment by Marvin84** [comment]
> we should always use the non-stressed version

Yes I agree. We need to open another PR for that.

File: common/helpers/g2p.py (hunk 1 above)
Comment (Marvin84) [reply common/helpers/g2p.py]: I saw your comment too late. I already did some changes. Now we have a g2p_augmenter class that can be used for any lexicon. The class gets either a pre-trained model or in case no path is provided will train one model and then apply and give back the augmented lexicon when the relative is called.

File: common/datasets/librispeech.py [hunk 2]
```diff
@@ -373,6 +379,60 @@ def get_bliss_lexicon(subdir_prefix=""):
...
+    :return: dictionary of Paths to augmented bliss_lexicon
+    :rtype: dict[str, Path]
+    """
+    alias_path = os.path.join(subdir_prefix, "LibriSpeech", "lexicon")
+    augmented_bliss_corpora = {}
+
+    original_bliss_lexicon =  get_bliss_lexicon(subdir_prefix=subdir_prefix)
+    g2p_args = get_g2p_parameters()
+
+    g2p_lexicon_job = BlissLexiconToG2PLexiconJob(bliss_lexicon=original_bliss_lexicon)
```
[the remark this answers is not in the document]
Marvin84: The bliss lexicon is anyhow needed for the pipeline for the final augmentation, namely for the `G2POutputToBlissLexiconJob`. I would prefer to let the pipeline be transparent regarding this little redundancy.

common/helpers/g2p.py
michelwi: I see, this is already very close to my suggestion. 
I am only missing the option to have different lexica for training the g2p model and as seed for applying it.
(e.g. in cases when you continuously get more data and want to add new OOV at the end of the lexicon)

Marvin84: Done -> [here](https://github.com/rwth-i6/i6_experiments/pull/15/commits/57b9dfdcffa5caa6a23cb918c5eb5ccfdbe4edb5)

common/helpers/g2p.py
```diff
@@ -0,0 +1,102 @@
...
+        self.apply_args = {}
+
+        if train_args is not None:
+            self.train_args.update(train_args)
+
+        if apply_args is not None:
+            self.apply_args.update(apply_args)
+
+
+    def train_and_set_g2p_model(self, train_lexicon:str, alias_path:str):
```
JackTemaki, line 52: you could make this private and call it in the init if `g2p_model_path` is not set.

common/datasets/librispeech.py
```diff
@@ -373,6 +375,34 @@ def get_bliss_lexicon(subdir_prefix=""):
     return merge_lexicon_job.out_bliss_lexicon
 
 
+def get_g2p_augmented_bliss_lexicon_dict(subdir_prefix=""):
+    """
+    Given the original LibriSpeech bliss lexicon, it is possible to estimate the pronunciation for
+    out of vocabulary (OOV) words for each of the LbriSpeech training corpora. Here, we create a dictionary
```
mmz33, line 381: typo `LbriSpeech`

common/datasets/librispeech.py
```diff
@@ -373,6 +375,34 @@ def get_bliss_lexicon(subdir_prefix=""):
...
+    that has different train corpora as keys and the corresponding g2p augmented bliss lexicon as values
+
+    :param st subdir_prefix:
```
mmz33, line 384: typo `st`

common/datasets/librispeech.py
```diff
@@ -373,6 +375,34 @@ def get_bliss_lexicon(subdir_prefix=""):
...
+    :return: dictionary of Paths to augmented bliss_lexicon
+    :rtype: dict[str, Path]
+    """
+    alias_path = os.path.join(subdir_prefix, "LibriSpeech", "lexicon")
+    augmented_bliss_corpora = {}
+
+    original_bliss_lexicon =  get_bliss_lexicon(subdir_prefix=subdir_prefix)
```
mmz33, line 391: removing additional space after `=`

common/helpers/g2p.py
```diff
@@ -0,0 +1,102 @@
...
+from i6_core.corpus.stats import ExtractOovWordsFromCorpusJob
+from i6_core.g2p.apply import ApplyG2PModelJob
+from i6_core.g2p.convert import BlissLexiconToG2PLexiconJob, G2POutputToBlissLexiconJob
+from i6_core.g2p.train import TrainG2PModelJob
+
+
+class G2PBasedOovAugmenter():
+    def __init__(self,
+                 original_bliss_lexicon,
+                 g2p_model_path:Optional[str] = None,
```
mmz33, line 18: this should be None or `tk.Path` right? also there should be a space after `:`. Black formatting should fix this.

common/helpers/g2p.py
```diff
@@ -0,0 +1,102 @@
...
+                 train_args:Optional[dict] = None,
+                 apply_args:Optional[dict] = None,
+                 ):
+        """
+        :param original_bliss_lexicon: path to the original lexicon with OOV
+        :param g2p_model_path: path to the g2p model, if none a g2p model is trained
+        #######################################
+        :param train_args = {
```
mmz33, line 26: maybe add the comments about the default params under class directly

common/helpers/g2p.py
```diff
@@ -0,0 +1,102 @@
...
+
+class G2PBasedOovAugmenter():
+    def __init__(self,
+                 original_bliss_lexicon,
+                 g2p_model_path:Optional[str] = None,
+                 train_args:Optional[dict] = None,
+                 apply_args:Optional[dict] = None,
+                 ):
+        """
+        :param original_bliss_lexicon: path to the original lexicon with OOV
```
mmz33, line 23: missing params type here. also similar to the others. (to be consistent with the current params docs)

mmz33 (changes requested):

JackTemaki (changes requested):

common/helpers/g2p.py
JackTemaki: This is not needed, one variant (docstring or python) is fine.

JackTemaki: `original_bliss_lexicon` is missing a type though

common/helpers/g2p.py
```diff
@@ -0,0 +1,102 @@
...
+        g2p_train_job.add_alias(os.path.join(alias_path, "train_g2p_model"))
+        self.g2p_model_path = g2p_train_job.out_best_model
+
+
+    def get_g2p_augmented_bliss_lexicon(
+            self,
+            bliss_corpus:Path,
+            corpus_name: str,
+            alias_path:str,
+            train_lexicon:Optional[str] = None,
```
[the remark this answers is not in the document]
mmz33: yes i would agree with this.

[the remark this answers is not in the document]
Marvin84: I do not understand your comment. As you can see the class does not have separate attributes for `train_lexicon` and `original_bliss_lexicon` (or how @michelwi call seed_lexicon). The idea is that you instantiate the class for the lexicon with respect to which you want to identify your OOVs, when you call this function you can decide whether your model is trained with the same lexicon or another lexicon.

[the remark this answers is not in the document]
mmz33: > you can decide whether your model is trained with the same lexicon or another lexicon.
but this is not consistent. the lexicon used to get the OOVs should be the same used to train the g2p. I can not think of a case where you need to use another lexicon.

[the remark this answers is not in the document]
Marvin84: > lexicon used to get the OOVs should be the same used to train the g2p

Here, the idea is that you might want to use an already existing sis job, and instead of providing the path of the model, it is cleaner to use just the input lexicon and let the rest to the class. 
The option of providing the model path is for cases we use external pretrained models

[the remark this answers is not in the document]
Marvin84: > I can not think of a case where you need to use another lexicon

As @michelwi pointed out, you might have this case.

common/helpers/g2p.py
```diff
@@ -0,0 +1,102 @@
...
+
+from i6_core.corpus.stats import ExtractOovWordsFromCorpusJob
+from i6_core.g2p.apply import ApplyG2PModelJob
+from i6_core.g2p.convert import BlissLexiconToG2PLexiconJob, G2POutputToBlissLexiconJob
+from i6_core.g2p.train import TrainG2PModelJob
+
+
+class G2PBasedOovAugmenter():
+    def __init__(self,
+                 original_bliss_lexicon,
```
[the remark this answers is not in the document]
Marvin84: > used for training the G2P model

This is not necessarily the lexicon that is used for train. this is the lexicon you want to augment.

common/helpers/g2p.py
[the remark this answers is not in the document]
mmz33: but in @michelwi he still uses the `seed_lex` to get the OOVs so it is still consistent. In your case, you use `self.original_bliss_lexicon`.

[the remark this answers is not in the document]
JackTemaki: But then this is exactly the wrong way. The class should get the "training" lexicon, and this function can then get an  "reference" (or seed, although I do not get the term) lexicon for the OOVs that is then extended and returned

common/helpers/g2p.py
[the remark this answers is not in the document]
JackTemaki: I see the problem, but for me this direction makes no sense at all. As i commened below, then this should be swapped. Here you should have the training lexicon, and in the function were you augment the lexicon, you pass the lexicon to augment

common/helpers/g2p.py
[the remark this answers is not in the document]
Marvin84: > The class should get the "training" lexicon

If you see how the class is designed and named, you can see that it is not g2p-model-centered but lexicon-centered. This is why the attribute of the class from my point of view should not be the train lexicon but the lexicon you want to augment.

common/helpers/g2p.py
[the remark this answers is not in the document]
Marvin84: I see your point though. Passing the `train_lexicon` to the `get_g2p_augmented_bliss_lexicon` function is not a good idea. However, as I answered in the other thread, this class is not designed for training g2p model. So, I would just add an additional attribute for the train lexicon

common/helpers/g2p.py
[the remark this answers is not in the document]
JackTemaki: The class name definitely does not say this in my opinion... So in my head it is much more logical to have it g2p-model centered. Because otherwise the train and apply args would not make much sense.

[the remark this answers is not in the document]
Marvin84: > So in my head it is much more logical to have it g2p-model

Ok. Let's do it in this way.

common/datasets/librispeech.py
```diff
@@ -373,6 +375,34 @@ def get_bliss_lexicon(subdir_prefix=""):
...
+    bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
+    for corpus_name, bliss_corpus in bliss_corpus_dict.items():
+        if "train" in corpus_name:
+            augmented_bliss_corpora[corpus_name] = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
+                bliss_corpus=bliss_corpus,
+                corpus_name=corpus_name,
+                alias_path=alias_path
+            )
+
+    return augmented_bliss_corpora
```
JackTemaki, line 403: I just noticed, this should be lexicon instead of corpora or not?

common/helpers/g2p.py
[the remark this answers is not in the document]
michelwi: "seed" because it grows when you add more words to it.

Also note, that in my code in each iteration the seed_lexicon is overwritten with the updated lexicon. (See sorting of lemmata, s.t. we can reuse alignments made with the smaller corpus after more and more oov)

@Marvin84 I see that your code is centered around the lexicon-to-be-augmented. What I (we) are arguing that it should rather be around the g2p model, because in practice (possibility only my practice) I want to have only one model but continuously add (append) more words to it.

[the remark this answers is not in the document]
Marvin84: > What I (we) are arguing that it should rather be around the g2p model

Yes, but then that should be implemented differently. I am adapting this implementation to your idea. The idea is that there is an additional optional parameter for `train_lexicon`. In case it is not set then this is the same as the `original_bliss_lexicon`. Then if the model path is None, a model is trained.

common/helpers/g2p.py
```diff
@@ -0,0 +1,101 @@
...
+        "devel"          :"5%",
+        "size_constrains":"0,1,0,1"
+        }
+        #######################################
+        :param apply_args = {
+        "variants_mass"  :1.0,
+        "variants_number":1
+        }
+        """
+        super().__init__()
```
JackTemaki, line 41: There is no super class, so this can be removed.

common/helpers/g2p.py
```diff
@@ -0,0 +1,101 @@
...
+        self.train_lexicon = original_bliss_lexicon if train_lexicon is None else train_lexicon
+        self.g2p_model_path = g2p_model_path
+        self.train_args = {}
+        self.apply_args = {}
+
+        if train_args is not None:
+            self.train_args.update(train_args)
+
+        if apply_args is not None:
+            self.apply_args.update(apply_args)
```
JackTemaki, line 52: this was much shorter before and did the same thing, or did you want to add some defaults? If the dicts are empty anyway you do not need to call update, but just keep the one-line if/else statement.

common/helpers/g2p.py
```diff
@@ -0,0 +1,101 @@
...
+            corpus_name:str,
+            alias_path:str
+    ):
+        extract_oov_job = ExtractOovWordsFromCorpusJob(
+            bliss_corpus=bliss_corpus,
+            bliss_lexicon=self.original_bliss_lexicon
+        )
+        extract_oov_job.add_alias(os.path.join(alias_path, "extract-oov-from-{}".format(corpus_name)))
+
+        if self.g2p_model_path is None:
```
JackTemaki, line 77: This can be part of the init, so that you do not call this multiple times.

common/helpers/g2p.py
JackTemaki: I would support this, including a short sentence what this class should do.

JackTemaki (changes requested):

common/helpers/g2p.py
Marvin84: This has not been changed from the first commit of the class. I don't understand what you mean. Here, the jobs have their own default parameters. However, one might change a value. The default arguments in both cases are `None`, so I cannot initialize the `self.{train,apply}_args` with the passed parameters, however I would need to update the dict if the user passed something.

> keep the one-line if/else statement.
What do you mean online if/else statement?

common/helpers/g2p.py
Marvin84: This is done here, because it needs the `alias_path`. I do want the class having this path. If you prefer I can avoid adding the alias. Otherwise, this is where it should be called.

common/helpers/g2p.py
JackTemaki: Oh sorry, I thought it was different before. I mean:
`self.train_args = train_args if train_args else {}`

Marvin84: I can also do it like this, if you prefer.

JackTemaki: I am just saying it can be shorter, but you can decide

common/helpers/g2p.py
JackTemaki: Hmm I see, I am not fully happy with this but then it is okay.

common/helpers/g2p.py
Marvin84: ok. done

JackTemaki (approved):

common/helpers/g2p.py
```diff
@@ -0,0 +1,100 @@
...
+
+from i6_core.corpus.stats import ExtractOovWordsFromCorpusJob
+from i6_core.g2p.apply import ApplyG2PModelJob
+from i6_core.g2p.convert import BlissLexiconToG2PLexiconJob, G2POutputToBlissLexiconJob
+from i6_core.g2p.train import TrainG2PModelJob
+
+
+class G2PBasedOovAugmenter():
+    def __init__(self,
+                 original_bliss_lexicon:str,
```
mmz33, line 17: there should be space after `:` because this violates PEP8. Black formatting does not fix this?

mmz33 (changes requested):

common/helpers/g2p.py
```diff
@@ -0,0 +1,100 @@
...
+        }
+        """
+        self.original_bliss_lexicon = original_bliss_lexicon
+        self.train_lexicon = original_bliss_lexicon if train_lexicon is None else train_lexicon
+        self.g2p_model_path = g2p_model_path
+        self.train_args = train_args if train_args else {}
+        self.apply_args = apply_args if apply_args else {}
+
+
+    def _train_and_set_g2p_model(self, alias_path:str):
```
mmz33, line 54: also here

common/helpers/g2p.py
```diff
@@ -0,0 +1,100 @@
...
+        g2p_train_job = TrainG2PModelJob(
+            g2p_lexicon=g2p_lexicon_job.out_g2p_lexicon,
+            **self.train_args)
+        g2p_train_job.add_alias(os.path.join(alias_path, "train_g2p_model"))
+        self.g2p_model_path = g2p_train_job.out_best_model
+
+
+    def get_g2p_augmented_bliss_lexicon(
+            self,
+            bliss_corpus:Path,
```
mmz33, line 66: also here

mmz33 (changes requested):

mmz33 (approved):

common/datasets/librispeech.py
```diff
@@ -373,6 +375,34 @@ def get_bliss_lexicon(subdir_prefix=""):
...
+    original_bliss_lexicon = get_bliss_lexicon(subdir_prefix=subdir_prefix)
+    g2p_augmenter = G2PBasedOovAugmenter(original_bliss_lexicon=original_bliss_lexicon)
+
+    bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
+    for corpus_name, bliss_corpus in bliss_corpus_dict.items():
+        if "train" in corpus_name:
+            augmented_bliss_lexica[corpus_name] = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
+                bliss_corpus=bliss_corpus,
+                corpus_name=corpus_name,
+                alias_path=alias_path
```
michelwi, line 400: I would still suggest that we use the initial lexicon to train the g2p model and then extend it by the train-clean-100 OOV.
Then to create the train-clean-360-lexicon, we should augment the train-clean-100-lexicon by the remaining OOVs.

Not only do we save a little bit of computational effort on shared OOVs, but also would we be able to use the 100h alignment as input for 960h training jobs. (e.g. Pipeline: monophon -> 100h, triphon->460h, triphon->960h)
The only downside would be, that if we wanted to to 360h-only training, then a few unnecessary 100h OOV words would be added.

JackTemaki: > The only downside would be, that if we wanted to to 360h-only training, then a few unnecessary 100h OOV words would be added.

But this could be separated, so the 360h lexicon is "standalone" but the 460h lexicon is the 100h lexicon and then the additional OOVs from 360 added.

Marvin84: > I would still suggest 

Please do a separate PR for this.

PR #17, opened 2021-08-10
Create lexicon without stress

Marvin84:


common/datasets/librispeech.py
```diff
@@ -349,8 +349,18 @@ def get_bliss_lexicon(subdir_prefix=""):
         checksum="d722bc29908cd338ae738edd70f61826a6fca29aaa704a9493f0006773f79d71",
     )
 
+    text_lexicon = download_lexicon_job.out_file
+    if use_stress_marker:
+        from i6_core.text import PipelineJob
+        eliminate_stress_job = PipelineJob(
+        text_lexicon, ["sed 's/[0-9]//g'"],
```
JackTemaki, line 356: use black for indentation

common/datasets/librispeech.py
```diff
@@ -324,7 +324,7 @@ def get_special_lemma_lexicon():
     return lex
 
 
-def get_bliss_lexicon(subdir_prefix=""):
+def get_bliss_lexicon(subdir_prefix="", use_stress_marker=False):
```
JackTemaki, line 327: docstring for stress marker is missing, and this should be before the `subdir_prefix` (I think it is a good convention to have it last, but we actually need to discuss at some point)

JackTemaki (changes requested):

common/datasets/librispeech.py
mmz33, line 356: Are you sure it is safe to do this? what if there is something like: `<phon score=0.0>...`? and maybe there are other cases

mmz33 (changes requested):

common/datasets/librispeech.py
Marvin84: Please check the input to the Job. This is not a bliss lexicon, but a text file. Namely, the one you take after downloading the corpus.

mmz33: ah sorry i thought it is a bliss lexicon.

mmz33: I see. I thought about this because in SWB lexicon text file for example, there are orths with digits.

common/datasets/librispeech.py
```diff
@@ -349,8 +350,20 @@ def get_bliss_lexicon(subdir_prefix=""):
         checksum="d722bc29908cd338ae738edd70f61826a6fca29aaa704a9493f0006773f79d71",
     )
 
+    text_lexicon = download_lexicon_job.out_file
+    if use_stress_marker:
```
JackTemaki, line 354: This should be `if not`

common/datasets/librispeech.py
JackTemaki: I would still move `use_stress_marker` to be the first argument, because this is the one that influence the output.

common/datasets/librispeech.py
```diff
@@ -334,7 +334,8 @@ def get_bliss_lexicon(subdir_prefix=""):
     while the special lemmas come first. This way the result resembles the "legacy" lexicon closely, and all
     "special" entries are at one position.
 
-    :param st subdir_prefix:
+    :param str subdir_prefix:
+    :param bool use_stress_marker:
```
JackTemaki, line 338: maybe give some short explanation here

JackTemaki (changes requested):

JackTemaki (approved):

common/datasets/librispeech.py
michelwi: what if words contain digits? (e.g. "9/11" or "hdf5")

mmz33: ok i checked seems to be no words with digits.

Marvin84: > ok i checked seems to be no words with digits.

Glad you are now convinced, even though I already told you that this has been checked by me.

PR #18, opened 2021-08-11
allow for g2p augmented Librispeech lexicon with stress markers

michelwi:
This adds the option to use stress markers while g2p augmenting the librispeech lexicon if one so desires.

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -399,6 +399,7 @@ def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=
     out of vocabulary (OOV) words for each of the LibriSpeech training corpora. Here, we create a dictionary
     that has different train corpora as keys and the corresponding g2p augmented bliss lexicon as values
 
+    :param bool use_stress_marker: uses phonem symbols with stress markers
     :param str subdir_prefix:
     :return: dictionary of Paths to augmented bliss_lexicon
     :rtype: dict[str, Path]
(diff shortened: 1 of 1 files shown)


common/datasets/librispeech.py
```diff
@@ -393,7 +393,7 @@ def get_bliss_lexicon(use_stress_marker=False, subdir_prefix=""):
     return merge_lexicon_job.out_bliss_lexicon
 
 
-def get_g2p_augmented_bliss_lexicon_dict(subdir_prefix=""):
+def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=""):
```
JackTemaki, line 396: missing docstring!

common/datasets/librispeech.py
```diff
@@ -399,6 +399,7 @@ def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=
     out of vocabulary (OOV) words for each of the LibriSpeech training corpora. Here, we create a dictionary
     that has different train corpora as keys and the corresponding g2p augmented bliss lexicon as values
 
+    :param bool use_stress_marker: uses phonem symbols with stress markers
```
Marvin84, line 402: typo phonem -> phoneme

Marvin84 (changes requested):

common/datasets/librispeech.py
JackTemaki:
```suggestion
    :param bool use_stress_marker: uses phoneme symbols with stress markers
```

JackTemaki (approved):

PR #19, opened 2021-08-11
base g2p augmented lexicon for larger corpora on prev. lexicon

michelwi:
This implements my suggestion from https://github.com/rwth-i6/i6_experiments/pull/15

We would use the initial lexicon to train the (all) g2p models and then for each corpus the lexicon is continuously extended by the next larger corpus' OOVs. The order will be alphabetically i.e. "train-clean-100"->"train-clean-360"->"train-other-500"

Advantages:
 * we save a little bit of computational effort on shared OOVs
 * we will be able to use the 100h alignment as input for training jobs of larger corpora. (e.g. GMM Pipeline: monophone -> 100h, triphone->460h, triphone->960h)

Disadvantages:
 * if we wanted to do 360h-only training, then a few unnecessary 100h OOV words would be added.

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -410,18 +410,21 @@ def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=
     original_bliss_lexicon = get_bliss_lexicon(
         use_stress_marker=use_stress_marker, subdir_prefix=subdir_prefix
     )
-    g2p_augmenter = G2PBasedOovAugmenter(original_bliss_lexicon=original_bliss_lexicon)
+    current_bliss_lexicon = original_bliss_lexicon
 
     bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
-    for corpus_name, bliss_corpus in bliss_corpus_dict.items():
+    for corpus_name, bliss_corpus in sorted(bliss_corpus_dict.items()):
         if "train" in corpus_name:
-            augmented_bliss_lexica[
-                corpus_name
-            ] = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
+            g2p_augmenter = G2PBasedOovAugmenter(
+                original_bliss_lexicon=updated_bliss_lexicon,
+                train_lexicon=original_bliss_lexicon,
+            )
+            current_bliss_lexicon = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
                 bliss_corpus=bliss_corpus,
                 corpus_name=corpus_name,
                 alias_path=alias_path,
             )
+            augmented_bliss_lexica[corpus_name] = current_bliss_lexicon
 
     return augmented_bliss_lexica
 
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -415,16 +415,19 @@ def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=
     bliss_corpus_dict = get_bliss_corpus_dict(subdir_prefix=subdir_prefix)
     for corpus_name, bliss_corpus in sorted(bliss_corpus_dict.items()):
         if "train" in corpus_name:
-            g2p_augmenter = G2PBasedOovAugmenter(
-                original_bliss_lexicon=updated_bliss_lexicon,
-                train_lexicon=original_bliss_lexicon,
-            )
-            current_bliss_lexicon = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
-                bliss_corpus=bliss_corpus,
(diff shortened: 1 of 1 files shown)


JackTemaki (approved):
From my side this would be okay, but I think my opinion is here a little less important than for other people.

common/datasets/librispeech.py
```diff
@@ -410,18 +410,21 @@ def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=
...
+    for corpus_name, bliss_corpus in sorted(bliss_corpus_dict.items()):
         if "train" in corpus_name:
-            augmented_bliss_lexica[
-                corpus_name
-            ] = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
+            g2p_augmenter = G2PBasedOovAugmenter(
+                original_bliss_lexicon=updated_bliss_lexicon,
+                train_lexicon=original_bliss_lexicon,
+            )
+            current_bliss_lexicon = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
```
JackTemaki, line 422: the update step will be empty for "460" and "960" then, because we already have all OOVs in the earlier lexicas, so those two can be skipped (lex460 = lex360 and lex960 = lex500)

JackTemaki (changes requested):

JackTemaki:
> FYI: there is a g2p model available at openslr. the official lexicon already has g2p generated pronunciations

Yes this is true, but there are still OOVs.

Marvin84:
> FYI: there is a g2p model available at openslr. the official lexicon already has g2p generated pronunciations

The g2p augmented class proposed and merged from [here]( https://github.com/rwth-i6/i6_experiments/pull/15), has also the option of getting a g2p model path. We can decide to generate both augmented lexica, with our trained modela nd with official trained model

common/datasets/librispeech.py
michelwi: This is true, but what do you propose to change? Should there be exceptions like `if corpus_name=='train-clean-460': <copy 360>` or should I hard code the three relevant corpora again?

Just asking to find the most elegant solution.

christophmluscher:
> > FYI: there is a g2p model available at openslr. the official lexicon already has g2p generated pronunciations
> 
> The g2p augmented class proposed and merged from [here](https://github.com/rwth-i6/i6_experiments/pull/15), has also the option of getting a g2p model path. We can decide to generate both augmented lexica, with our trained model and with official trained model

that would leave us the option to compare the generated lexica. I like that

common/datasets/librispeech.py
JackTemaki:
```python3
if corpus_name in ['train-clean-460', 'train-other-960']:
    augmented_bliss_lexica[corpus_name] = current_bliss_lexicon
else:
    do stuff....
```

JackTemaki: So in those cases the current one is already the "correct" one, or not?

christophmluscher:
> > FYI: there is a g2p model available at openslr. the official lexicon already has g2p generated pronunciations
> 
> Yes this is true, but there are still OOVs.
> 
> EDIT: Or do you mean instead of training it ourself we should just download the existing g2p?

both :)
use pre existing and training ourselves.

JackTemaki (approved):

common/datasets/librispeech.py
```diff
@@ -415,16 +415,19 @@ def get_g2p_augmented_bliss_lexicon_dict(use_stress_marker=False, subdir_prefix=
...
-                original_bliss_lexicon=updated_bliss_lexicon,
-                train_lexicon=original_bliss_lexicon,
-            )
-            current_bliss_lexicon = g2p_augmenter.get_g2p_augmented_bliss_lexicon(
-                bliss_corpus=bliss_corpus,
-                corpus_name=corpus_name,
-                alias_path=alias_path,
-            )
-            augmented_bliss_lexica[corpus_name] = current_bliss_lexicon
+            if corpus_name in ["train-clean-460", "train-other-960"]:
```
Marvin84, line 418: I might be asking too much, but wouldn't the cleaner solution here be: run the OOV job, compare the resulting OOV set with the set of the previous step, and then decide?

JackTemaki: Sisyphus would support this kind of "Graph decision" only with the async workflows, and we do not want to introduce this here.

Marvin84: I see. My only concern is that this is based on prior knowledge. Then I suggest that we document it at least.

JackTemaki: Ok then I will accept this PR, and add documentation, do some testing, and if everything is fine add also the new lexicon stuff to /work/common

JackTemaki:
@christophmluscher Any comments?

PR #21, opened 2021-08-20
Fairseq hydra training job

JanEnglerRWTH:
Sisyphus job which starts a "fairseq-hydra-train" training for a given .yaml config data and (optional) command line arguments. 
The checkpoint directory is moved into the output folder of the job for re-starting the job.

### DIFF
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -1,0 +1,97 @@
+import os
+import subprocess as sp
+import yaml
+
+from sisyphus import *
+
+import recipe.i6_core.util as util
+
+
+class DictToYamlJob(Job):
+    """
+    Writes a dict into a .yaml file
+    """
+
+    def __init__(self, config_dict, hash_reset, *, yaml_prefix=""):
+        """
+        :param dict config_dict:
+        :param str hash_reset: Temporary way to reset the sisyphus hash for debugging purposes
+        :param str yaml_prefix: Prefix which should be written to the beginning, for example "# @package _group_" or "#!rnn.py"
+        """
+        assert isinstance(config_dict, dict)
+
+        self.config_dict = config_dict
+        self.yaml_prefix = yaml_prefix
+        self.out_config_file = self.output_path("fairseq_config.yaml")
+        self.file_out = self.output_path("out")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        yaml_dict = yaml.dump(self.config_dict)
+        # "# @package _group_" was written at the beginning in the example .yaml from fairseq:
+        if self.yaml_prefix != "":
+            yaml_dict = self.yaml_prefix + "\n" + yaml_dict
+        with open(self.out_config_file.get_path(), "w") as file:
+            file.write(yaml_dict)
+
+
+class FairseqHydraTrainingJob(Job):
+    """
+    Train a Fairseq model using fairseq-hydra-train
+    """
+
+    def __init__(
+        self,
+        python_exe,
+        fairseq_exe,
+        hash_reset,
+        *,  # args below are keyword only
+        additional_args=[],
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=2,
+        gpu_rqmt=1,
+    ):
+        """
+        :param Path|str python_exe: File path to the executable for running python
+        :param Path|str fairseq_exe: File path to the Fairseq executable: fairseq-hydra-train (usually in the same folder as python exe '.../bin/')
+        :param str hash_reset: temporary way to reset the sisyphus hash for debugging purposes
+        :param list additional_args: The arguments needed to configure the Fairseq task
+        :param int|float time_rqmt:
+        :param int|float mem_rqmt:
+        :param int cpu_rqmt:
+        :param int gpu_rqmt:
+        """
+        self.fairseq_python_exe = python_exe
+        self.fairseq_train = fairseq_exe
+        self.additional_args = additional_args
+        self.checkpoint_dir = self.output_path("checkpoints")
+        self.file_out = self.output_path("out")
+
+        self.rqmt = {
+            "gpu": gpu_rqmt,
+            "cpu": cpu_rqmt,
+            "mem": mem_rqmt,
+            "time": time_rqmt,
+        }
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def _get_run_cmd(self):
+        run_cmd = [
+            tk.uncached_path(self.fairseq_python_exe),
+            tk.uncached_path(self.fairseq_train),
+        ]
+        run_cmd += self.additional_args
+        run_cmd += ["checkpoint.save_dir=" + str(self.checkpoint_dir)]
+        return run_cmd
+
+    def create_files(self):
+        util.create_executable("fairseq.sh", self._get_run_cmd())
+
+    def run(self):
+        sp.check_call(self._get_run_cmd())
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -12,22 +12,22 @@ class FairseqHydraConfig:
     An object that manages a Fairseq hydra config (inspired by the ReturnnConfig).
     """
 
-    def __init__(self, fairseq_hydra_config_dict, *, yaml_prefix=""):
+    def __init__(self, config_dict, *, yaml_prefix=""):
         """
-        :param dict fairseq_hydra_config_dict: Contains the information which is needed for fairseq-hydra-train. Will be converted and dumped into a .yaml
+        :param dict config_dict: Contains the information which is needed for fairseq-hydra-train. Will be converted and dumped into a .yaml
         :param str yaml_prefix: Prefix which should be written to the beginning of the config, for example "# @package _group_"
         """
-        assert isinstance(fairseq_hydra_config_dict, dict)
-        self.fairseq_hydra_config_dict = fairseq_hydra_config_dict
+        assert isinstance(config_dict, dict)
+        self.config_dict = config_dict
         self.yaml_prefix = yaml_prefix
 
     def write(self, path):
-        yaml_dict = yaml.dump(self.fairseq_hydra_config_dict)
+        config_yaml = yaml.dump(self.config_dict)
         # "# @package _group_" was written at the beginning in the example .yaml from fairseq:
         if self.yaml_prefix != "":
-            yaml_dict = self.yaml_prefix + "\n" + yaml_dict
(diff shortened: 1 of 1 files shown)


users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+
+class DictToYamlJob(Job):
+    """
+    Writes a dict into a .yaml file
+    """
+
+    def __init__(self, config_dict, hash_reset, *, yaml_prefix=""):
+        """
+        :param dict config_dict:
+        :param str hash_reset: Temporary way to reset the sisyphus hash for debugging purposes
```
vieting, line 18: This should be removed.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class DictToYamlJob(Job):
+    """
+    Writes a dict into a .yaml file
+    """
+
+    def __init__(self, config_dict, hash_reset, *, yaml_prefix=""):
+        """
+        :param dict config_dict:
```
vieting, line 17: The job name is generic and not related to fairseq. However, this variable implies that what is written is always a config. I'd either make this a `FairseqConfig` or `FairseqYamlConfig` with a `write` method which is called in a `create_files` task of the `FairseqHydraTrainingJob` similar to the `ReturnnConfig` or leave this job generic.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        :param str hash_reset: Temporary way to reset the sisyphus hash for debugging purposes
+        :param str yaml_prefix: Prefix which should be written to the beginning, for example "# @package _group_" or "#!rnn.py"
+        """
+        assert isinstance(config_dict, dict)
+
+        self.config_dict = config_dict
+        self.yaml_prefix = yaml_prefix
+        self.out_config_file = self.output_path("fairseq_config.yaml")
```
vieting, line 25: See above regarding config/fairseq references.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        self.file_out = self.output_path("out")
```
vieting, line 26:
```suggestion
        self.out_file = self.output_path("out")
```
See https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md

But this doesn't seem to be used anyway, right?

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+
+class FairseqHydraTrainingJob(Job):
+    """
+    Train a Fairseq model using fairseq-hydra-train
+    """
+
+    def __init__(
+        self,
+        python_exe,
+        fairseq_exe,
```
vieting, line 48: These should go to the end, see https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md.

You can do this also similar as in the `ReturnnTrainingJob` with defaults from `gs` if `None` is given here. Maybe this is also fine for now here in your user folder, but at least if it should some time go to i6_core, then it should satisfy the contributing guidelines.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        additional_args=[],
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=2,
+        gpu_rqmt=1,
+    ):
+        """
+        :param Path|str python_exe: File path to the executable for running python
+        :param Path|str fairseq_exe: File path to the Fairseq executable: fairseq-hydra-train (usually in the same folder as python exe '.../bin/')
+        :param str hash_reset: temporary way to reset the sisyphus hash for debugging purposes
```
vieting, line 60: This should be removed.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        :param list additional_args: The arguments needed to configure the Fairseq task
```
vieting, line 61: You could make the name more descriptive, e.g. `command_line_args`

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+    Train a Fairseq model using fairseq-hydra-train
+    """
+
+    def __init__(
+        self,
+        python_exe,
+        fairseq_exe,
+        hash_reset,
+        *,  # args below are keyword only
+        additional_args=[],
```
vieting, line 51: Don't use empty lists as a function's default parameter. Use `None` instead and test for it in the function body, e.g., `self.additional_args = additional_args or []`.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        :param int|float time_rqmt:
+        :param int|float mem_rqmt:
+        :param int cpu_rqmt:
+        :param int gpu_rqmt:
+        """
+        self.fairseq_python_exe = python_exe
+        self.fairseq_train = fairseq_exe
+        self.additional_args = additional_args
+        self.checkpoint_dir = self.output_path("checkpoints")
+        self.file_out = self.output_path("out")
```
vieting, line 71:
```suggestion
        self.out_checkpoint_dir = self.output_path("checkpoints")
        self.out_file = self.output_path("out")
```

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def _get_run_cmd(self):
+        run_cmd = [
+            tk.uncached_path(self.fairseq_python_exe),
+            tk.uncached_path(self.fairseq_train),
```
vieting, line 87:
```suggestion
            self.fairseq_python_exe.get_path(),
            self.fairseq_train.get_path(),
```

users/engler/fairseq/training.py
vieting: Again, `self.file_out` is not used, right?

vieting (changes requested):

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        *,  # args below are keyword only
+        additional_args=[],
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=2,
+        gpu_rqmt=1,
+    ):
+        """
+        :param Path|str python_exe: File path to the executable for running python
+        :param Path|str fairseq_exe: File path to the Fairseq executable: fairseq-hydra-train (usually in the same folder as python exe '.../bin/')
```
vieting, line 59: Maybe rename to `fairseq_hydra_exe`?

vieting:
Do you always have the `config-dir` and `config-name` in the `additional_args`? You could also have a separate input argument for this where you can optionally provide the yaml config job and then automatically add the config in `_get_run_cmd` as this seems to be a very common case.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def _get_run_cmd(self):
+        run_cmd = [
+            tk.uncached_path(self.fairseq_python_exe),
+            tk.uncached_path(self.fairseq_train),
+        ]
+        run_cmd += self.additional_args
+        run_cmd += ["checkpoint.save_dir=" + str(self.checkpoint_dir)]
```
christophmluscher, line 90: Does fairseq throw an error if `checkpoint.save_dir` is defined twice? by job and by user in config for example?

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,97 @@
...
+        :param list additional_args: The arguments needed to configure the Fairseq task
+        :param int|float time_rqmt:
+        :param int|float mem_rqmt:
+        :param int cpu_rqmt:
+        :param int gpu_rqmt:
+        """
+        self.fairseq_python_exe = python_exe
+        self.fairseq_train = fairseq_exe
+        self.additional_args = additional_args
+        self.checkpoint_dir = self.output_path("checkpoints")
```
christophmluscher, line 70: this is an output directory, right?
`directory=True`

users/engler/fairseq/training.py
christophmluscher: when sisyphus builds the job graph we define via `register_output` which outputs we want. How is this applied to this job, so that sisyphus registers to run the job?

christophmluscher (changes requested):

users/engler/fairseq/training.py
vieting: I thought it was generally recommended not to use `tk.uncached_path` anymore (https://github.com/rwth-i6/i6_core/issues/32)

But yes, for `str` this suggested change would not work.

christophmluscher: yes generally not recommended but so far I don't think we can completely avoid this.. right!? @JackTemaki

JackTemaki: We can not completely avoid this, as you can not use "Path" objects in the `settings.py` (circular dependencies). I solved this by just not using the settings to define Paths, but specifying the Patsh explicitly for experiments.

users/engler/fairseq/training.py
JanEnglerRWTH: I have misinterpreted the guidelines^^. After looking at the `out_*` again, I understand it now
2. In a Sisyphus job class the output variables **end** in `out_*`

vieting (dismissed):

christophmluscher (approved):

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,128 @@
...
+        if keep_epochs is None:
+            self.keep_epochs = set(stored_epochs)
+        else:
+            self.keep_epochs = set(keep_epochs)
+        self.max_epoch = max_epoch
+        self.save_interval = save_interval
+        self.out_fairseq_hydra_yaml = self.output_path("fairseq_hydra_config.yaml")
+        self.out_checkpoint_dir = self.output_path("checkpoints", directory=True)
+        self.out_models = {
+            k: str(self.out_checkpoint_dir) + "/checkpoint{}.pt".format(k)
```
michelwi, line 82: Attention: this will write the absolute path of the model files in this dict. Therefore each subsequent Job using an item of this dict 
a) will have a hash depending on the location of your setup
b) will not know that the model file is linked to this training Job. So the dependency graph will be broken here

I suggest using a checkpoint class similar to https://github.com/rwth-i6/i6_core/blob/d27320747b0d7ebe40b8fec6a96cdacb7a961aa5/returnn/training.py#L38-L61

christophmluscher: good point! thank you

@JanEnglerRWTH this would be a very useful addition

PR #22, opened 2021-10-11
FairseqHydraTrainingJob: 2 major changes. 

JanEnglerRWTH:
1. Added ability to select the fairseq git that should be used for easier version-control.
2. Added a checkpoint-ish class according to michelwi comment on the previous PR

### DIFF
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -38,7 +38,6 @@ class PytorchHydraModel:
 
     def __init__(self, fairseq_hydra_config_file, model, epoch):
         """
-
         :param Path fairseq_hydra_config_file: Path to a returnn config file
         :param Path model: Path to a pytorch checkpoint
         :param int epoch:
@@ -104,12 +103,14 @@ class FairseqHydraTrainingJob(Job):
         self.fairseq_python_exe = (
             fairseq_python_exe
             if fairseq_python_exe is not None
-            else gs.FAIRSEQ_PYTHON_EXE
+            else getattr(gs, "FAIRSEQ_PYTHON_EXE", None)
         )
-        self.fairseq_hydra_exe = (
-            fairseq_hydra_exe if fairseq_hydra_exe is not None else gs.FAIRSEQ_HYDRA_EXE
-        )
+        self.fairseq_hydra_exe = fairseq_hydra_exe
         self.fairseq_root = fairseq_root
+        # We assume that only one of the two possible entry points is given as an input
+        assert (self.fairseq_root is not None) ^ (self.fairseq_hydra_exe is not None)
+        if self.fairseq_root is not None:
+            assert self.fairseq_python_exe is not None
 
         # Outputs:
         self.out_fairseq_hydra_yaml = self.output_path("fairseq_hydra_config.yaml")
@@ -161,12 +162,13 @@ class FairseqHydraTrainingJob(Job):
         run_cmd += ["checkpoint.save_dir=" + self.out_checkpoint_dir.get_path()]
         run_cmd += ["checkpoint.save_interval=" + str(self.save_interval)]
         run_cmd += ["optimization.max_epoch=" + str(self.max_epoch)]
+
         if self.fairseq_root is not None:
             sys.path.insert(0, self.fairseq_root)
             hydra_train_entry = self.fairseq_root + "fairseq_cli/hydra_train.py"
-            run_cmd.insert(0, tk.uncached_path(self.fairseq_python_exe))
-            run_cmd.insert(1, tk.uncached_path(hydra_train_entry))
+            run_cmd.insert(0, tk.uncached_path(hydra_train_entry))
         else:
+            run_cmd.insert(0, tk.uncached_path(self.fairseq_hydra_exe))
+        if self.fairseq_python_exe is not None:
             run_cmd.insert(0, tk.uncached_path(self.fairseq_python_exe))
-            run_cmd.insert(1, tk.uncached_path(self.fairseq_hydra_exe))
         return run_cmd


users/engler/fairseq/training.py
```diff
@@ -0,0 +1,162 @@
...
+        max_epoch=1,
+        save_interval=1,
+        keep_epochs=None,
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=2,
+        gpu_rqmt=1,
+        fairseq_python_exe=None,
+        fairseq_hydra_exe=None,
+        fairseq_git=None,
```
JackTemaki, line 70: the consistent name to the other jobs would be `fairseq_root`, because in theory the source folder can come from non-git sources.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def _get_run_cmd(self):
+        run_cmd = [
+            "--config-dir",
+            os.path.dirname(self.out_fairseq_hydra_yaml.get_path()),
+            "--config-name",
+            os.path.basename(self.out_fairseq_hydra_yaml.get_path()),
+        ]
+        run_cmd += self.command_line_args
+        run_cmd += ["checkpoint.save_dir=" + str(self.out_checkpoint_dir)]
```
JackTemaki, line 150: please use `.get_path()` instead of string conversion for all variables which are definitely `Path` or `Variable` objects you want to resolve. This is just a convention we want to introduce to make clear that something is a Path object, and to prohibit any caching mistakes. The implicit conversion `__str__` representation was indended for different purposes orignally, and resolves to `get_cached_path()` and not `get_path()`, which can potentially lead to errors.

JackTemaki: Of course in the cases you definitely want to have the caching, use `get_cached_path`.

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,162 @@
...
+            k: PytorchHydraModel(
+                self.out_fairseq_hydra_yaml,
+                self.output_path("checkpoints/checkpoint{}.pt".format(k)),
+                k,
+            )
+            for k in stored_epochs
+            if k in self.keep_epochs
+        }
+        self.gpu_rqmt = gpu_rqmt
+        self.fairseq_python_exe = fairseq_python_exe
```
JackTemaki, line 111: This is also not a problem but only a convention: The inputs should come first before all ouputs, see https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTING.md

users/engler/fairseq/training.py
```diff
@@ -0,0 +1,162 @@
...
+        :param FairseqHydraConfig fairseq_hydra_config:
+        :param list command_line_args: Additional command line arguments (starting with "--*") to configure the Fairseq-hydra task
+        :param int max_epoch: maximum number of epochs to run. Note that this value IS currently HASHED.
+        :param int save_interval: save a checkpoint each n-th epoch
+        :param list[int]|set[int]|None keep_epochs: specify which checkpoints are kept in self.out_models, use None for each save_interval-th epoch
+        :param int|float time_rqmt: Overall time requirements
+        :param int|float mem_rqmt: Memory requirements (per GPU)
+        :param int cpu_rqmt: Required number of CPUs (per GPU)
+        :param int gpu_rqmt: Number of required GPUs
+        :param Path|str python_exe: File path to the executable for running python
```
michelwi, line 82: Did you mean
```suggestion
        :param Path|str fairseq_python_exe: File path to the python executable for running fairseq
```

users/engler/fairseq/training.py
michelwi, line 111: Maybe also allow to set a default in the global settings
```suggestion
        self.fairseq_python_exe = (
            fairseq_python_exe if fairseq_python_exe is not None else gs.FAIRSEQ_PYTHON_EXE
        )
```

vieting:
If I look at files changed, it seems like everything in the file is new. Maybe that's because this branch has conflicts with the main branch. Could you please resolve them, so that we can see the differential changes of this PR?

christophmluscher (approved):

users/engler/fairseq/training.py
```diff
@@ -30,6 +31,23 @@ def write(self, path):
...
 
 
+class PytorchHydraModel:
+    """
+    Defines a Pytorch hydra model as yaml config, pytorch checkpoint file and epoch
+    """
+
+    def __init__(self, fairseq_hydra_config_file, model, epoch):
+        """
+
```
vieting, line 41:
```suggestion
```

users/engler/fairseq/training.py
```diff
@@ -109,20 +145,28 @@ def create_files(self):
...
         run_cmd += ["checkpoint.save_interval=" + str(self.save_interval)]
         run_cmd += ["optimization.max_epoch=" + str(self.max_epoch)]
-        if self.fairseq_python_exe is not None:
+        if self.fairseq_root is not None:
+            sys.path.insert(0, self.fairseq_root)
+            hydra_train_entry = self.fairseq_root + "fairseq_cli/hydra_train.py"
+            run_cmd.insert(0, tk.uncached_path(self.fairseq_python_exe))
+            run_cmd.insert(1, tk.uncached_path(hydra_train_entry))
+        else:
             run_cmd.insert(0, tk.uncached_path(self.fairseq_python_exe))
```
vieting, line 170: We should only add this if it's not `None`, right?

users/engler/fairseq/training.py
```diff
@@ -109,20 +145,28 @@ def create_files(self):
...
         ]
         run_cmd += self.command_line_args
-        run_cmd += ["checkpoint.save_dir=" + str(self.out_checkpoint_dir)]
+        run_cmd += ["checkpoint.save_dir=" + self.out_checkpoint_dir.get_path()]
         run_cmd += ["checkpoint.save_interval=" + str(self.save_interval)]
         run_cmd += ["optimization.max_epoch=" + str(self.max_epoch)]
-        if self.fairseq_python_exe is not None:
+        if self.fairseq_root is not None:
+            sys.path.insert(0, self.fairseq_root)
+            hydra_train_entry = self.fairseq_root + "fairseq_cli/hydra_train.py"
```
vieting, line 166: I would add this after uncashing `self.fairseq_root`.

users/engler/fairseq/training.py
```diff
@@ -109,20 +145,28 @@ def create_files(self):
...
+            run_cmd.insert(0, tk.uncached_path(self.fairseq_python_exe))
+            run_cmd.insert(1, tk.uncached_path(hydra_train_entry))
```
vieting, line 168: I think `run_cmd = [tk.uncached_path(...), tk.uncached_path(...)] + run_cmd` is easier to read, but that's a matter of taste.

users/engler/fairseq/training.py
```diff
@@ -76,26 +101,37 @@ def __init__(
             self.keep_epochs = set(keep_epochs)
         self.max_epoch = max_epoch
         self.save_interval = save_interval
+        self.fairseq_python_exe = (
+            fairseq_python_exe
+            if fairseq_python_exe is not None
+            else gs.FAIRSEQ_PYTHON_EXE
```
vieting, line 107: Actually it should be okay not to have a `fairseq_python_exe`, right? If `fairseq_hydra_exe` is given, this is not necessary, is it? If not, we should allow that it is `None` and `gs.FAIRSEQ_PYTHON_EXE` is not defined.

Then, if it actually is necessary later, we should assert that it is not `None`.
```suggestion
            else getattr(gs, "FAIRSEQ_PYTHON_EXE", None)
```

users/engler/fairseq/training.py
```diff
@@ -76,26 +101,37 @@ def __init__(
...
+        )
+        self.fairseq_hydra_exe = (
+            fairseq_hydra_exe if fairseq_hydra_exe is not None else gs.FAIRSEQ_HYDRA_EXE
```
vieting, line 110: See above, might also not be needed.

users/engler/fairseq/training.py
```diff
@@ -76,26 +101,37 @@ def __init__(
...
+        )
+        self.fairseq_root = fairseq_root
```
vieting, line 112: I think we should assert that either of `fairseq_hydra_exe` and `fairseq_root` is `None` because then we clearly know which version to use in `_get_run_cmd()`.

vieting (changes requested):

users/engler/fairseq/training.py
vieting: Yes exactly, it should be XOR.

users/engler/fairseq/training.py
JanEnglerRWTH: Okay

PR #23, opened 2021-10-18
hybrid rasr system update

christophmluscher:
updates:
- RasrSystem
- GmmSystem

Please look into the RasrSteps mechanism! In the run function of the GmmSystem


common/setups/hybrid/gmm_system.py
```diff
@@ -234,8 +249,8 @@ def cart_and_lda(
             generalized_eigenvalue_args=generalized_eigenvalue_args,
         )
         self.jobs[corpus]["cart_and_lda_{}_{}".format(corpus, name)] = cart_lda
-        self.lda_matrices["{}_{}".format(corpus, name)] = cart_lda.last_lda_matrix
-        self.cart_trees["{}_{}".format(corpus, name)] = cart_lda.last_cart_tree
+        self.lda_matrices[corpus][name] = cart_lda.last_lda_matrix
```
JackTemaki, line 252: This seems to be a very reasonable change, but can we also change `corpus` to `corpus_key` everywhere this is a string? This would distinguish this much better from other code, where `corpus` is an actual corpus.

common/setups/hybrid/gmm_system.py
```diff
@@ -320,21 +335,25 @@ def triphone_training(
     def vtln_feature_flow(
         self,
         name: str,
-        corpus: str,
+        train_corpora: str,
```
JackTemaki, line 338: `train_corpus` or `train_corpus_key`?  this should not be plural I guess...

common/setups/hybrid/gmm_system.py
```diff
@@ -489,16 +509,16 @@ def estimate_cmllr(
 
         overlay = "%s_cmllr_%s" % (corpus, name) if overlay is None else overlay
         self.add_overlay(corpus, overlay)
-        self.crp[overlay].segment_path = new_segments.segment_path
+        self.crp[overlay].out_segment_path = new_segments.out_segment_path
```
JackTemaki, line 512: this is a crp object, I am quite sure the `out_` prefix is wrong then.

common/setups/hybrid/gmm_system.py
```diff
@@ -646,210 +682,535 @@ def recognition(
...
+                map_file=speaker_seq.out_cluster_map_file,
+                transform_dir=self.jobs[overlay]["cmllr"].transforms,
+            )
+
+            with tk.block(f"{name}_recognition"):
+                recog_func = (
+                    self.recog_and_optimize if optimize_am_lm_scale else self.recog
+                )
+
+                for l in lm_scales:
```
JackTemaki, line 774: You are already looping over `l` in the product at the top

common/setups/hybrid/gmm_system.py
```diff
@@ -646,210 +682,535 @@ def recognition(
...
                 )
 
-            for c in self.dev_corpora:
-                # recog and optimize
-                pass
-
-            for c in self.test_corpora:
-                pass
+        for step_idx, (step_name, step_args) in enumerate(steps.get_step_iter()):
+            # ---------- Feature Extraction ----------
```
JackTemaki, line 860: I think every step definition should be a function, this way it is easier when working with class inheritance.

common/setups/hybrid/gmm_system.py
```diff
@@ -646,210 +682,535 @@ def recognition(
...
+            if step_name.startswith("extract"):
+                self.extract_features(feat_args=step_args)
+                if (
+                    "voiced" in self.hybrid_init_args.feature_extraction_args.keys()
+                    and "tone" in self.hybrid_init_args.feature_extraction_args.keys()
+                ):
```
JackTemaki, line 866: is the `voiced+tone` case only relevant for the GmmModel? Otherwise I would recommend to move this to the `RasrSystem`

common/setups/hybrid/gmm_system.py
```diff
@@ -603,12 +623,13 @@ def recognition(
...
         corpus: str,
         feature_flow: str,
-        pronunciation_scale: float,
+        pronunciation_scale: Union[float, List[float]],
         search_parameters: dict,
         rtf: float,
         mem: float,
         parallelize_conversion: bool,
         lattice_to_ctm_kwargs: dict,
+        reestimate_prior=False,
```
michelwi, line 632: Is this function for general recognition or only GMM?

for GMM we have no prior. so no need to (re)estimate it.

common/setups/hybrid/rasr_system.py
```diff
@@ -153,6 +160,50 @@ def single_density_mixtures(
...
-            alignment=meta.select_element(self.alignments, corpus, alignment),
+            alignment=meta.select_element(
+                self.alignments, corpus, (corpus, alignment, -1)
+            ),
             split_first=False,
         )
+
+    # -------------------- Forced Alignment --------------------
+
+    def forced_align(self, name, corpus, target_corpus, flow, feature_scorer, **kwargs):
```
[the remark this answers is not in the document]
michelwi: This is possible duplication of `meta.system.align`.  Can it be reused/extended?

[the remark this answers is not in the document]
michelwi: maybe add a flag `dump_alignment` so the user can choose to not have the alignment dumped to save disk space.

Or is the use case of this function explicitly to create and dump arbitrary alignments? then maybe rename to `align_and_dump`.
You can maybe internally call the `meta.system.align`

common/setups/hybrid/gmm_system.py
christophmluscher: I agree. This is also inconsistent in `i6_core`.
One way to define this would be:
`corpus_key` -> `str` (basically corpus identifier, `corpus_name` would also be an option)
`corpus_object` -> `CorpusObject`
`corpus_file` -> `str` (would be filepath)
`corpus` -> `Path`
I am open for suggestions on this.
@curufinwe your opinion?

common/setups/hybrid/gmm_system.py
```diff
@@ -603,12 +623,13 @@ def recognition(
         # parameters just for passing through
         corpus: str,
         feature_flow: str,
-        pronunciation_scale: float,
+        pronunciation_scale: Union[float, List[float]],
```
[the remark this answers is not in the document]
christophmluscher: you mean `Union[float, List[float]]` -> `List[float]`?

common/setups/hybrid/gmm_system.py
```diff
@@ -646,210 +682,535 @@ def recognition(
...
+            )
+            self.crp[overlay].corpus_config.file = recognized_corpus.output_corpus_path
+            self.crp[overlay].segment_path = copy.deepcopy(
+                self.crp[corpus].segment_path
+            )
+
+            self.corpora[overlay] = copy.deepcopy(self.corpora[corpus])
+            self.corpora[overlay].corpus_file = recognized_corpus.output_corpus_path
+
+            alignment = mm.AlignmentJob(
```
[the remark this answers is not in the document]
christophmluscher: at the moment I use one loop over `iters`, `pronunciation_scales`, and `lm_scales`. So move the complete `add_overlay` and `AlignmentJob` out of the loop?

common/setups/hybrid/gmm_system.py
```diff
@@ -646,210 +682,535 @@ def recognition(
...
-
-        # ---------- Triphone ----------
-        if "tri" in steps:
-            for c in self.train_corpora:
-                self.triphone_training(
-                    "tri",
-                    c,
-                    initial_alignment="train_mono",
-                    **self.gmm_triphone_args.triphone_training_args,
+            else:
```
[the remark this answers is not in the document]
christophmluscher: added

common/setups/hybrid/gmm_system.py
christophmluscher: only for GMM -> removed

common/setups/hybrid/gmm_system.py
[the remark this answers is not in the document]
michelwi: just consistent with `lm_scales`. I have no strong preference if both should be `Union[float, List[float]]` or `List[float]`?

common/setups/hybrid/gmm_system.py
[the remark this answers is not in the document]
michelwi: I have been re-reading the code and it I noticed I was wrong. While indeed the alignment would be Independent of the LM scale, the LM scale determines which transcription (i.e. ctm file of previous recognition run with LM_scale=l) is used for the alignment (cf. line 717). So currently the implementation is correct.

But now I wonder if the overlay you are creating is correct. Because for every `it, p, l` you are creating the same overlay `"%s_sat" % corpus` despite the fact that the corpus file is different. 
Probably the correct thing happens, but after the loop the older overlays are inaccessible/overwritten.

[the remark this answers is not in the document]
christophmluscher: I added the `it, p, l` to the `overlay_key`

common/setups/hybrid/util.py
```diff
@@ -376,34 +407,81 @@ def __init__(
...
-        self.vtln_sat_training_args = vtln_sat_training_args
-        self.vtln_sat_recognition_args = vtln_sat_recognition_args
-        self.sdm_vtln_sat_args = sdm_vtln_sat_args
+        self.training_args = training_args
+        self.recognition_args = recognition_args
+        self.sdm_args = sdm_args
+
+
+class ForcedAlignmentArgs:
+    def __init__(self, name, target_corpus, flow, feature_scorer):
```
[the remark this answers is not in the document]
christophmluscher: No, because atm in the `run` function the feature scorer would be taken from all `train_corpora` if feature_scorer is not a FeatureScorer Object.
Idea would be to apply all feature scorers from all train_corpora to the new corpus.

common/setups/hybrid/util.py
```diff
@@ -376,34 +407,81 @@ def __init__(
...
     def __init__(
         self,
-        nn_training_args: Optional[dict] = None,
-        nn_count_recognition_args: Optional[dict] = None,
-        nn_nn_recognition_args: Optional[dict] = None,
-        nn_rescoring_args: Optional[dict] = None,
+        returnn_configs: Dict[str, returnn.ReturnnConfig],
+        training_args: Optional[dict] = None,
+        count_recognition_args: Optional[dict] = None,
+        lm_recognition_args: Optional[dict] = None,
```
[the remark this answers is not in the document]
christophmluscher: > both are lm_recognition_args, I suggest to rename lm_recognition_args to neural_recognition_args.

yes

> But then why make this distinction? we can have a list/dict/tuple of recognition_args and then some of them might have a neural lm configured and some have a count lm configured.

My reasoning was that normally we want to do count and neural recognition for a given model. That is why I wanted to have both. But not sure if this applies to all users..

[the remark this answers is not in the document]
michelwi: Some users would only do count based LM recognitions, other users might want to do [count_2gram, count_4gram, ffnn_4gram, lstm_small, lstm_big, trafo] recognitions. 

As far as I can see, there would be no special different handling of count vs. neural recognition args except maybe for a different alias.

I would implement it e.g. as a dict of dicts with `lm_name` -> `recognition_args_dict` so that we can have an arbitrary number of recognition settings.

common/setups/hybrid/util.py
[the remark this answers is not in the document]
michelwi: I was asking because in your suggested change `rasr_system.py:171` in function
```
   def forced_align(self, name, corpus, target_corpus, flow, feature_scorer, **kwargs):
```

there is the additional `corpus` argument. So I was wondering if by symmetry we need this also here, or could omit it there.

common/setups/hybrid/util.py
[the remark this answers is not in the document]
christophmluscher: OK. Agreed.

common/setups/hybrid/util.py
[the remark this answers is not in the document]
christophmluscher: Ah OK. the function parameter `corpus` is needed but specified in the run function.

christophmluscher:
I agree that the voiced+tone extraction should be at a different place. Since this is quite uncommon, I think I ll maybe set it up as a generic feature extraction and not a specific setup. Not sure yet. I ll think about it. For now removed from this pull request.

common/setups/hybrid/gmm_system.py
christophmluscher: https://github.com/rwth-i6/i6_experiments/pull/23#issuecomment-961280375

common/setups/hybrid/gmm_system.py
```diff
@@ -204,7 +224,7 @@ def monophone_training(
     def cart_and_lda(
         self,
         name: str,
-        corpus: str,
+        corpus_key: str,
         initial_flow: str,
```
JackTemaki, line 228: Here we also have the `object` vs `str` issue, as those are not real flows but flow keys. We should also rename this...

common/setups/hybrid/gmm_system.py
```diff
@@ -264,7 +287,7 @@ def cart_and_lda(
     def triphone_training(
         self,
         name: str,
-        corpus: str,
+        corpus_key: str,
         feature_flow: str,
         initial_alignment: str,
```
JackTemaki, line 292: Maybe we use the `_key` suffix for all string variables that look into the member dicts?

common/setups/hybrid/gmm_system.py
JackTemaki: Yes I think this looks really good now!

common/setups/hybrid/gmm_system.py
christophmluscher: fine with me.

christophmluscher: I think this makes it clearer.

christophmluscher: I just looked into the code again and no this is not perfect.. as soon as we use the `meta.select_element` we can use both.. 

`feature_flow` being the selector.. so actually the type is not correct..

```
def select_element(collection, default_corpus, selector, default_index=-1):
    if not isinstance(selector, (list, tuple, str)):
        return selector
```

michelwi: I agree.

But sometimes a function accepts an `Object` or a (`str`,`str`,`int`) tupel of (corpus_key, name, index) and gets the correct `Object` from the member dicts via `i6_core.meta.system.select_element`.
How would we handle/name this?

christophmluscher: so depending on how we process this selector down the line it can be `str` or `str|rasr.SOMETHING`

christophmluscher: so I ll check the code and adapt where possible but in some cases adding the key would limit our choices artificially..

christophmluscher: maybe as an example to see what I mean, check `monophone_training` function, how `feature_energy_flow` and `feature_flow` are handled. Former has to be a `str`, latter does not.

christophmluscher: > But sometimes a function accepts an Object or a (str,str,int) tupel of (corpus_key, name, index) and gets the correct Object from the member dicts via i6_core.meta.system.select_element.
How would we handle/name this?

not sure if this is the best approach. I would leave it as `feature_flow` but add a type as: `Union[str, tuple[str, str,str], rasr.FlagDependentFlowAttribute`.
But maybe that gets too unwieldy?

JackTemaki (approved):
Just some minor stuff. I think we should merge this soon so that we can start testing and also work on possibly own PRs (like me adding some CTC related stuff).

common/setups/rasr/rasr_system.py
```diff
@@ -0,0 +1,225 @@
...
+                        self.add_energy_to_features(t, "mfcc+deriv")
+                        self.add_energy_to_features(t, "mfcc+deriv+norm")
+                    if kk == "gt":
+                        self.add_energy_to_features(t, "gt")
+                    if kk == "fb":
+                        self.add_energy_to_features(t, "fb")
+
+    # -------------------- Single Density Mixtures --------------------
+
+    def single_density_mixtures(
```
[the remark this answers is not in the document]
michelwi: This is used to estimate the prior from the alignment in Hybrid DNN-HMM systems when _the Povey method_ is not used. So Yes.

michelwi (approved):

common/setups/rasr/rasr_system.py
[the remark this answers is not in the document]
michelwi: > Or is this needed ONLY for DNN-HMM and can/should be moved to the GMM system?

This is needed for all _Hybrid_ systems, where a prior p(s) is used via the Bayes rule
``` p( x | s ) = p( s | x ) / p( s )  *  p( x ) ```
to replace the GMM emission prob. by a neural network.

Not sure if **only** needed here, but restricting it to the Gaussian Mixture Model (GMM) System is imho wrong, because it is also used in the Neural Network (NN) System.

>Is this needed for CTC/Transducer or some random new approaches? 

Transducer is a discriminative approach, so in principle no prior is needed. CTC can be both, a special case of a hybrid system or a direct discriminative model. Factored-Hybrid will definitively use it. Also currently we are supposed to do some Experiments that include priors into Transducers - but here I would strongly suggest to not use single-density-mixture but the softmax prior.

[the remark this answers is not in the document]
Comment (JackTemaki) [reply common/setups/rasr/rasr_system.py:167]: Ok I see, then it can stay here. Thanks for the explanation!

File: common/setups/hybrid/util.py [hunk 17]
```diff
@@ -376,34 +407,81 @@ def __init__(
...
+
+class ForcedAlignmentArgs:
+    def __init__(self, name, target_corpus, flow, feature_scorer):
+        self.name = name
+        self.target_corpus = target_corpus
+        self.flow = flow
+        self.feature_scorer = feature_scorer
+
+
+class ReturnnRasrDataInput(RasrDataInput):
```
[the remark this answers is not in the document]
christophmluscher: agreed. this pull request is for GMM. so I ll add as todo but address in a different pull request

PR #24, opened 2021-11-09
restructure feature extraction

JackTemaki:


christophmluscher (approved):
loop var naming not a must. otherwise ok

michelwi (approved):

PR #25, opened 2021-11-18
fairseq training: avoid config overwrite by default

vieting:
In the `FairseqHydraTrainingJob`, `max_epoch` and `save_interval` are currently set to 1 by default and values possibly stored in `fairseq_hydra_config` are overwritten by default, which I think is quite unintuitive. Therefore, this PR changes the behavior to use values from the config by default and to only overwrite them, if a value is explicitly given in the init.

### DIFF
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import subprocess as sp
 import yaml
@@ -57,8 +58,8 @@ class FairseqHydraTrainingJob(Job):
         fairseq_hydra_config,
         *,  # args below are keyword only
         command_line_args=None,
-        max_epoch=1,
-        save_interval=1,
+        max_epoch=None,
+        save_interval=None,
         keep_epochs=None,
         time_rqmt=4,
         mem_rqmt=4,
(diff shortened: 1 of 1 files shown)


michelwi (approved):
Maybe also ask @JanEnglerRWTH

vieting:
I didn't ask him because he's not at i6 anymore.

@JanEnglerRWTH you're of course welcome to give feedback anyway if you like.

users/engler/fairseq/training.py
```diff
@@ -91,15 +91,28 @@ def __init__(
...
+        ).get("save_interval", None)
+        assert (
+            self.save_interval is not None
+        ), "save_interval has to be set explicitly or via fairseq_hydra_config"
+        self.max_epoch = max_epoch or fairseq_hydra_config.config_dict.get(
+            "optimization", {}
+        ).get("max_epoch", None)
+        assert (
+            self.save_interval is not None
+        ), "save_interval has to be set explicitly or via fairseq_hydra_config"
```
christophmluscher, line 105: why two times `assert self.save_interval ...`? L98 and L103

users/engler/fairseq/training.py
```diff
@@ -91,15 +91,28 @@ def __init__(
         # Inputs:
         self.fairseq_hydra_config = fairseq_hydra_config
         self.command_line_args = command_line_args or []
-        stored_epochs = list(range(save_interval, max_epoch, save_interval)) + [
-            max_epoch
-        ]
+        self.save_interval = save_interval or fairseq_hydra_config.config_dict.get(
+            "checkpoint", {}
+        ).get("save_interval", None)
```
christophmluscher, line 96: is it possible to do something like!?

`self.save_interval = fairseq_hydra_config.config_dict.get("checkpoint", {}).get("save_interval", save_interval)`

then the default would be save_interval if the fairseq_hydra_config does not set the value. I think that would also save a bit of the logic in `_get_run_cmd`.

same for `max_epoch`

vieting: But this would change the default behavior. Now, if `save_interval` is not `None` in the init, we use this and overwrite what is specified in the config. Your suggestion would be the other way around. I think it is more intuitive to take the input argument if both are not `None`.

christophmluscher: thats true. I think this is disambiguous.. mabye add a warning if both are not `None`?

christophmluscher (approved):
> mabye add a warning if both are not None?

I think adding a warning if values are set via input param and via config file would help the user, but not a must

users/engler/fairseq/training.py
vieting: Yes, I did that now.

PR #26, opened 2021-11-19
fix: no automatic upper casing

christophmluscher:
fixes automatic upper casing in g2p pipeline (i6_core job update)

issue: rwth-i6/i6_core#177
pull: rwth-i6/i6_core#184

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -478,6 +478,7 @@ def get_g2p_augmented_bliss_lexicon_dict(
                     bliss_corpus=bliss_corpus,
                     corpus_name=corpus_name,
                     alias_path=alias_path,
+                    casing="upper",
                 )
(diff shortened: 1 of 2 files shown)


JackTemaki (changes requested):
This would break the existing LibriSpeech jobs, which is especially problematic for the ones that are already exportet to `/work/common`. 

The casing needs to be configurable, and set to "upper" for the LibriSpeech lexicon pipeline.

christophmluscher:
the casing is now configurable. for librispeech set to "upper" which should not break the hashes. the corresponding pull request for `i6_core` is rwth-i6/i6_core#184

I think the hash test is failing because in `i6_core` the corresponding pull request is not merged yet.. @JackTemaki yes??

christophmluscher:
I would only merge, after merging the pull request in `i6_core` and the hash test does not fail.

christophmluscher:
@JackTemaki ping :)

PR #27, opened 2021-11-23
adapt to new Lemma interface

michelwi:
The RASR Lexicon only supports a single syntactic token sequence per lemma while our implementation in i6_core supports multiple. The PR https://github.com/rwth-i6/i6_core/pull/185 fixes this.

This PR is to adapt the current recipes to the new interface and should land together with the PR in i6_core.


christophmluscher (approved):
overlap to #28

JackTemaki (approved):
Can now be accepted as https://github.com/rwth-i6/i6_core/pull/188 is merged

PR #29, opened 2021-11-29
add option for unknown phoneme

christophmluscher:
adds a flag to add an option for unknown phoneme mapping

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -304,7 +304,7 @@ def get_arpa_lm_dict(output_prefix="datasets"):
 
 
 @lru_cache()
-def get_special_lemma_lexicon():
+def get_special_lemma_lexicon(add_unknown_phoneme_and_mapping=True):
     """
     Generate the special lemmas for LibriSpeech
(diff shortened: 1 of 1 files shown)


michelwi (commented):
How wil RASR behave in training if the unknown lemma has no phonemic representation and an unknown word appears in the transcript?

JackTemaki:
> How wil RASR behave in training if the unknown lemma has no phonemic representation and an unknown word appears in the transcript?

Good questions, but I just found out that I need this flag for the CTC setup as well, so thanks @christophmluscher for adding this.

JackTemaki (approved):

christophmluscher:
> How wil RASR behave in training if the unknown lemma has no phonemic representation and an unknown word appears in the transcript?

I have no unknown lemma in training. And the unknown symbol got zero weights. Thats why a I added to check if there are performance differences. You basically dont have to consider the unknown phoneme in CART and so on

JackTemaki:
I just saw that the docstrings are missing, I will add them now.

PR #30, opened 2021-12-02
add hybrid nn system [wip]

christophmluscher:
add a hybrid nn system

still work in progress. very untested!!! but general feedback and input welcome

features are still missing but will be extended (seq disc training and restoring for example)

### DIFF
--- a/common/setups/rasr/gmm_system.py
+++ b/common/setups/rasr/gmm_system.py
@@ -923,7 +923,7 @@ class GmmSystem(RasrSystem):
         lm_scales = [lm_scales] if isinstance(lm_scales, float) else lm_scales
 
         for it, p, l in itertools.product(iters, pronunciation_scales, lm_scales):
-            prev_ctm_key = f"recog_{train_corpus_key}-{prev_ctm}-{corpus}-ps{p:02.2f}-lm{l:02.2f}-iter{it:02d}"
+            prev_ctm_key = f"recog_{train_corpus_key}-{prev_ctm[0]}-{corpus}-ps{prev_ctm[1]:02.2f}-lm{prev_ctm[2]:02.2f}-iter{prev_ctm[3]:02d}{prev_ctm[4]}"
             assert prev_ctm_key in self.ctm_files[corpus], (
                 "the previous recognition stage '%s' did not provide the required recognition: %s"
                 % (prev_ctm, prev_ctm_key)
@@ -936,7 +936,7 @@ class GmmSystem(RasrSystem):
                 self.corpora[corpus].corpus_file
             )
 
-            overlay_key = f"{corpus}_it{it}_ps{p}_lm{l}_sat"
+            overlay_key = f"{corpus}_{name}_it{it}_ps{p}_lm{l}_sat"
             self.add_overlay(corpus, overlay_key)
             self.crp[overlay_key].corpus_config = copy.deepcopy(
                 self.crp[corpus].corpus_config
@@ -1358,7 +1358,7 @@ class GmmSystem(RasrSystem):
             self.store_allophones(trn_c)
 
         for eval_c in self.dev_corpora + self.test_corpora:
-            self.create_stm_from_corpus(eval_c)
+            self.create_stm_from_corpus(eval_c, **self.hybrid_init_args.stm_args)
             self._set_scorer_for_corpus(eval_c)
 
         for step_idx, (step_name, step_args) in enumerate(steps.get_step_iter()):
--- a/common/setups/rasr/nn_system.py
+++ b/common/setups/rasr/nn_system.py
@@ -1,0 +1,668 @@
+__all__ = ["NnArgs", "NnSystem"]
+
+import copy
+import itertools
+import sys
+from typing import Dict, List, Optional, Tuple, Union
+
+# -------------------- Sisyphus --------------------
+
+import sisyphus.toolkit as tk
+import sisyphus.global_settings as gs
+
+from sisyphus.delayed_ops import DelayedFormat
+
+# -------------------- Recipes --------------------
+
+import i6_core.features as features
+import i6_core.rasr as rasr
+import i6_core.returnn as returnn
+
+from i6_core.util import MultiPath, MultiOutputPath
+
+from .rasr_system import RasrSystem
+
+from .util import (
+    RasrInitArgs,
+    ReturnnRasrDataInput,
+    OggZipHdfDataInput,
+    NnArgs,
+    RasrSteps,
+)
+
+# -------------------- Init --------------------
+
+Path = tk.setup_path(__package__)
+
+# -------------------- System --------------------
+
+
+class NnSystem(RasrSystem):
+    """
+    - 5 corpora types: train, devtrain, cv, dev and test
+        devtrain is a small split from the train set which is evaluated like
+        the cv but not used for error calculating. Since we can have different
+        datasubsets per subepoch, we do not caculate the tran score/error on
+        a consistent datasubset
+    - two training data settings: defined in returnn config or not
+    - 3 different types of decoding: returnn, rasr, rasr-label-sync
+    - 2 different lm: count, neural
+    - cv is dev for returnn training
+    - dev for lm param tuning
+    - test corpora for final eval
+
+    settings needed:
+    - am
+    - lm
+    - lexicon
+    - ce training
+    - ce recognition
+    - ce rescoring
+    - smbr training
+    - smbr recognition
+    - smbr rescoring
+    """
+
+    def __init__(
+        self,
+        returnn_root: Optional[str] = None,
+        returnn_python_home: Optional[str] = None,
+        returnn_python_exe: Optional[str] = None,
+    ):
+        super().__init__()
+
+        self.returnn_root = returnn_root or (
+            gs.RETURNN_ROOT if hasattr(gs, "RETURNN_ROOT") else None
+        )
+        self.returnn_python_home = returnn_python_home or (
+            gs.RETURNN_PYTHON_HOME if hasattr(gs, "RETURNN_PYTHON_HOME") else None
+        )
+        self.return_python_exe = returnn_python_exe or (
+            gs.RETURNN_PYTHON_EXE if hasattr(gs, "RETURNN_PYTHON_EXE") else None
+        )
+
+        self.tf_fwd_input_name = "tf-fwd-input"
+
+        self.cv_corpora = []
+        self.devtrain_corpora = []
+
+        self.train_input_data = None
+        self.cv_input_data = None
+        self.devtrain_input_data = None
+        self.dev_input_data = None
+        self.test_input_data = None
+
+        self.train_cv_pairing = None
+
+        self.datasets = {}
+
+        self.oggzips = {}  # TODO remove?
+        self.hdfs = {}  # TODO remove?
+        self.extern_rasrs = {}  # TODO remove?
+
+        self.nn_configs = {}
+        self.nn_models = {}  # TODO remove?
+        self.nn_checkpoints = {}
+
+    # -------------------- Helpers --------------------
+    @staticmethod
+    def adapt_returnn_config_for_recog(returnn_config: returnn.ReturnnConfig):
+        """
+        Adapt a RETURNN config for recognition, e.g., remove loss and use log softmax activation in last layer
+
+        :param ReturnnConfig config:
+        :rtype ReturnnConfig:
+        """
+        assert isinstance(returnn_config, returnn.ReturnnConfig)
+        config = copy.deepcopy(returnn_config)
+        forward_output_layer = config.config.get("forward_output_layer", "output")
+        network = config.config.get("network")
+        for layer_name, layer in network.items():
+            if layer.get("unit", None) in {"lstmp"}:
+                layer["unit"] = "nativelstm2"
+            if layer.get("target", None):
+                layer.pop("target")
+                layer.pop("loss", None)
+                layer.pop("loss_scale", None)
+                layer.pop("loss_opts", None)
+        if network[forward_output_layer]["class"] == "softmax":
+            network[forward_output_layer]["class"] = "linear"
+            network[forward_output_layer]["activation"] = "log_softmax"
+        elif network[forward_output_layer]["class"] == "linear":
+            if network[forward_output_layer]["activation"] == "softmax":
+                network[forward_output_layer]["activation"] = "log_softmax"
+            elif network[forward_output_layer]["activation"] == "sigmoid":
+                network[forward_output_layer]["activation"] = "log_sigmoid"
+            elif network[forward_output_layer]["activation"] == "exp":
+                network[forward_output_layer]["activation"] = None
+            elif network[forward_output_layer]["activation"] is None:
+                network[forward_output_layer]["activation"] = "log"
+        # target = 'classes'
+        if "cropped" in network:
+            if network["output"]["from"] == ["cropped"]:
+                network["output"]["from"] = "upsample"
+            network.pop("cropped")
+            if "lstm_bwd_1" in network:
+                network["lstm_bwd_1"]["from"] = "upsample"
+                network["lstm_fwd_1"]["from"] = "upsample"
+            if "lstm_fwd_1_no_init" in network:
+                network["lstm_bwd_1_no_init"]["from"] = "upsample"
+                network["lstm_fwd_1_no_init"]["from"] = "upsample"
+
+        return config
+
+    @staticmethod
+    def get_tf_flow(
+        checkpoint_path: Union[Path, returnn.Checkpoint],
+        tf_graph_path: Path,
+        returnn_op_path: Path,
+        forward_output_layer: str = "output",
+        tf_fwd_input_name: str = "tf-fwd-input",
+    ):
+        """
+        Create flow network and config for the tf-fwd node
+
+        :param Path checkpoint_path: RETURNN model checkpoint which should be loaded
+        :param Path tf_graph_path: compiled tf graph for the model
+        :param Path returnn_op_path: path to native lstm library
+        :param str forward_output_layer: name of layer whose output is used
+        :param str tf_fwd_input_name: tf flow node input name. see: add_tf_flow_base_flow()
+        :rtype: FlowNetwork
+        """
+        input_name = tf_fwd_input_name
+
+        tf_flow = rasr.FlowNetwork()
+        tf_flow.add_input(input_name)
+        tf_flow.add_output("features")
+        tf_flow.add_param("id")
+        tf_fwd = tf_flow.add_node("tensorflow-forward", "tf-fwd", {"id": "$(id)"})
+        tf_flow.link(f"network:{input_name}", tf_fwd + ":input")
+        tf_flow.link(tf_fwd + ":log-posteriors", "network:features")
+
+        tf_flow.config = rasr.RasrConfig()
+
+        tf_flow.config[tf_fwd].input_map.info_0.param_name = "input"
+        tf_flow.config[
+            tf_fwd
+        ].input_map.info_0.tensor_name = "extern_data/placeholders/data/data"
+        tf_flow.config[
+            tf_fwd
+        ].input_map.info_0.seq_length_tensor_name = (
+            "extern_data/placeholders/data/data_dim0_size"
+        )
+
+        tf_flow.config[tf_fwd].output_map.info_0.param_name = "log-posteriors"
+        tf_flow.config[
+            tf_fwd
+        ].output_map.info_0.tensor_name = f"{forward_output_layer}/output_batch_major"
+
+        tf_flow.config[tf_fwd].loader.type = "meta"
+        tf_flow.config[tf_fwd].loader.meta_graph_file = tf_graph_path
+        tf_flow.config[tf_fwd].loader.saved_model_file = checkpoint_path
+
+        tf_flow.config[tf_fwd].loader.required_libraries = returnn_op_path
+
+        return tf_flow
+
+    @staticmethod
+    def add_tf_flow_to_base_flow(
+        base_flow: rasr.FlowNetwork,
+        tf_flow: rasr.FlowNetwork,
+        tf_fwd_input_name: str = "tf-fwd-input",
+    ):
+        """
+        Integrate tf-fwd node into the regular flow network
+
+        :param FlowNetwork base_flow:
+        :param FlowNetwork tf_flow:
+        :param str tf_fwd_input_name: see: get_tf_flow()
+        :rtype: FlowNetwork
+        """
+        assert (
+            len(base_flow.outputs) == 1
+        ), "Not implemented otherwise"  # see hard coded tf-fwd input
+        base_output = list(base_flow.outputs)[0]
+
+        input_name = tf_fwd_input_name
+
+        feature_flow = rasr.FlowNetwork()
+        base_mapping = feature_flow.add_net(base_flow)
+        tf_mapping = feature_flow.add_net(tf_flow)
+        feature_flow.interconnect_inputs(base_flow, base_mapping)
+        feature_flow.interconnect(
+            base_flow, base_mapping, tf_flow, tf_mapping, {base_output: input_name}
+        )
+        feature_flow.interconnect_outputs(tf_flow, tf_mapping)
+
+        return feature_flow
+
+    def _add_output_alias_for_train_job(
+        self,
+        train_job: Union[returnn.ReturnnTrainingJob, returnn.ReturnnRasrTrainingJob],
+        train_corpus_key: str,
+        cv_corpus_key: str,
+        name: str,
+    ):
+        train_job.add_alias(f"train_nn/{train_corpus_key}_{cv_corpus_key}/{name}_train")
+        self.jobs[f"{train_corpus_key}_{cv_corpus_key}"][name] = train_job
+        self.nn_models[f"{train_corpus_key}_{cv_corpus_key}"][
+            name
+        ] = train_job.out_models
+        self.nn_checkpoints[f"{train_corpus_key}_{cv_corpus_key}"][
+            name
+        ] = train_job.out_checkpoints
+        self.nn_configs[f"{train_corpus_key}_{cv_corpus_key}"][
+            name
+        ] = train_job.out_returnn_config_file
+        tk.register_output(
+            f"train_nn/{train_corpus_key}_{cv_corpus_key}/{name}_learning_rate.png",
+            train_job.out_plot_lr,
+        )
+
+    # -------------------- Setup --------------------
+    def init_system(
+        self,
+        hybrid_init_args: RasrInitArgs,
+        train_data: Dict[str, Union[ReturnnRasrDataInput, OggZipHdfDataInput]],
+        cv_data: Dict[str, Union[ReturnnRasrDataInput, OggZipHdfDataInput]],
+        devtrain_data: Optional[
+            Dict[str, Union[ReturnnRasrDataInput, OggZipHdfDataInput]]
+        ] = None,
+        dev_data: Optional[Dict[str, ReturnnRasrDataInput]] = None,
+        test_data: Optional[Dict[str, ReturnnRasrDataInput]] = None,
+        train_cv_pairing: Optional[
+            List[Tuple[str]]
+        ] = None,  # List[Tuple[trn_c, cv_c, name, dvtr_c]]
+    ):
+        self.hybrid_init_args = hybrid_init_args
+
+        self._init_am(**self.hybrid_init_args.am_args)
+
+        devtrain_data = devtrain_data if devtrain_data is not None else {}
+        dev_data = dev_data if dev_data is not None else {}
+        test_data = test_data if test_data is not None else {}
+
+        self._assert_corpus_name_unique(
+            train_data, cv_data, devtrain_data, dev_data, test_data
+        )
+
+        self.train_input_data = train_data
+        self.cv_input_data = cv_data
+        self.devtrain_input_data = devtrain_data
+        self.dev_input_data = dev_data
+        self.test_input_data = test_data
+
+        self.train_corpora.extend(list(train_data.keys()))
+        self.cv_corpora.extend(list(cv_data.keys()))
+        self.devtrain_corpora.extend(list(devtrain_data.keys()))
+        self.dev_corpora.extend(list(dev_data.keys()))
+        self.test_corpora.extend(list(test_data.keys()))
+
+        self._set_eval_data(dev_data)
+        self._set_eval_data(test_data)
+
+        self.train_cv_pairing = (
+            list(itertools.product(self.train_corpora, self.cv_corpora))
+            if train_cv_pairing is None
+            else train_cv_pairing
+        )
+
+        for pairing in self.train_cv_pairing:
+            trn_c = pairing[0]
+            cv_c = pairing[1]
+
+            self.jobs[f"{trn_c}_{cv_c}"] = {}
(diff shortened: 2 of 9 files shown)


common/setups/rasr/gmm_system.py
```diff
@@ -826,6 +830,46 @@ def sat_recognition(
...
+        gmm_output.lexicon = self.crp[corpus_key].lexicon_config
+        gmm_output.cart_tree = self.crp[
+            corpus_key
+        ].acoustic_model_config.state_tying.file
+        gmm_output.feature_flows = self.feature_flows[corpus_key]
+        gmm_output.features = self.feature_caches[corpus_key]
+        gmm_output.segment_path = self.crp[corpus_key].segment_path
+        gmm_output.allophone_file = self.allophone_files["base"]
+
+        if corpus_type == "dev" or corpus_type == "test":
```
michelwi, line 859: Why only set LM for test sets? E.g. for sequence training we also have a LM for the train set.

common/setups/rasr/gmm_system.py
```diff
@@ -826,6 +830,46 @@ def sat_recognition(
...
+            gmm_output.lm_file = self.crp[corpus_key].language_model_config.file
+            gmm_output.lm = self.crp[corpus_key].language_model_config
+            gmm_output.acoustic_mixtures = self.mixtures[corpus_key][
+                f"train_{steps.get_step_names_as_list()[step_idx - 1]}"
+            ][-1]
```
michelwi, line 864: It always used to be the case, that the feature scorer is associated to a dev/test set. This makes sense, as there might be different settigns per test set.
But for the mixtures themselves it would make more sense to me to associate them to the respective train set they were trained on.

Actually, I think `self.mixtures[corpus_key][name]` only exists for train corpora anyway.

common/setups/rasr/gmm_system.py
```diff
@@ -826,6 +830,46 @@ def sat_recognition(
...
+
+        if corpus_type == "train":
+            gmm_output.alignments = self.alignments[corpus_key][
+                f"train_{steps.get_step_names_as_list()[step_idx - 1]}"
+            ][-1]
```
michelwi, line 869: Also dev sets can have alignments, e.g. if we use them for cross validation of a ce loss.

common/setups/rasr/gmm_system.py
```diff
@@ -851,11 +895,17 @@ def run_monophone_step(self, step_args):
                 name = step_args.training_args["name"]
                 feature_scorer = (trn_c, f"train_{name}")
 
+                recog_args = copy.deepcopy(step_args.recognition_args)
+                if step_args.test_recognition_args is None:
+                    break
+                recog_args.update(step_args.test_recognition_args)
```
michelwi, line 901: Maybe we do want to use the same recognition args for test and dev corpora?
```suggestion
                if step_args.test_recognition_args is not None:
                    recog_args.update(step_args.test_recognition_args)
```
or would you rather that
`test_recognition_args = None` -> no test
`test_recognition_args = {}` -> test with same config

common/setups/rasr/gmm_system.py
```diff
@@ -887,11 +937,17 @@ def run_triphone_step(self, step_args):
                 name = step_args.training_args["name"]
                 feature_scorer = (trn_c, f"train_{name}")
 
+                recog_args = copy.deepcopy(step_args.recognition_args)
+                if step_args.test_recognition_args is None:
```
michelwi, line 941: see above

common/setups/rasr/gmm_system.py
```diff
@@ -940,7 +996,21 @@ def run_vtln_step(self, step_args, step_idx, steps):
                 )
 
             for tst_c in self.test_corpora:
-                pass
+                name = step_args.training_args["train"]["name"]
+                feature_scorer = (trn_c, f"train_{name}")
+
+                recog_args = copy.deepcopy(step_args.recognition_args)
+                if step_args.test_recognition_args is None:
+                    break
```
michelwi, line 1004: see above

common/setups/rasr/gmm_system.py
```diff
@@ -969,7 +1039,21 @@ def run_sat_step(self, step_args):
                 )
 
             for tst_c in self.test_corpora:
-                pass
+                name = step_args.training_args["name"]
+                feature_scorer = (trn_c, f"train_{name}")
+
+                recog_args = copy.deepcopy(step_args.recognition_args)
+                if step_args.test_recognition_args is None:
+                    break
```
michelwi, line 1047: above

common/setups/rasr/gmm_system.py
```diff
@@ -998,7 +1082,21 @@ def run_vtln_sat_step(self, step_args):
...
```
michelwi, line 1090: a

common/setups/rasr/gmm_system.py
```diff
@@ -1007,6 +1105,37 @@ def run_vtln_sat_step(self, step_args):
                     **step_args.sdm_args,
                 )
 
+    def run_recognition_step(self, step_args):
```
michelwi, line 1108: Ive been writing "see above" now so many times that I would suggest to remove code duplication by moving out the two loops and calls to `self.recognition` into a separate function. Or maybe this `run_recognition_step` is already sufficient and you just call this in each training step.

common/setups/rasr/gmm_system.py
```diff
@@ -998,7 +1082,21 @@ def run_vtln_sat_step(self, step_args):
...
+                name = step_args.training_args["name"]
+                feature_scorer = (trn_c, f"train_{name}")
+
+                recog_args = copy.deepcopy(step_args.recognition_args)
+                if step_args.test_recognition_args is None:
+                    break
+                recog_args.update(step_args.test_recognition_args)
+                recog_args["optimize_am_lm_scale"] = False
+
+                self.recognition(
```
michelwi, line 1094:
```suggestion
                self.sat_recognition(
```

common/setups/rasr/gmm_system.py
```diff
@@ -969,7 +1039,21 @@ def run_sat_step(self, step_args):
...
```
michelwi, line 1051:
```suggestion
                self.sat_recognition(
```

common/setups/rasr/util.py
```diff
@@ -244,10 +251,15 @@ def __init__(
             }
         }
         ##################################################
+        test_recognition_args
```
michelwi, line 254: If you keep it this way, add comment that the default `None` leads to no test recognition
```suggestion
        :param test_recognition_args:
```

common/setups/rasr/util.py
```diff
@@ -382,9 +407,14 @@ def __init__(
         ##################################################
         :param recognition_args:
         ##################################################
+        test_recognition_args
```
michelwi, line 410:
```suggestion
        :param test_recognition_args:
```

common/setups/rasr/util.py
```diff
@@ -349,6 +368,10 @@ def __init__(
...
```
michelwi, line 371:
```suggestion
        :param test_recognition_args:
```

common/setups/rasr/util.py
```diff
@@ -309,6 +322,10 @@ def __init__(
...
```
michelwi, line 325:
```suggestion
        :param test_recognition_args:
```

common/setups/rasr/util.py
```diff
@@ -429,47 +461,235 @@ def __init__(
...
+        self.recognition_args = recognition_args
+
+
+class OutputArgs:
+    def __init__(self, name):
+        self.name = name
+        self.corpus_key_type_pair = {}
+
+    def add_corpus_key_type_pair(self, corpus_key, corpus_type):
+        self.corpus_key_type_pair[corpus_key] = corpus_type
```
michelwi, line 476: optional cosmetic change:
```suggestion
        self.corpus_type_mapping = {}

    def define_corpus_type(self, corpus_key, corpus_type):
        self.corpus_type_mapping[corpus_key] = corpus_type
```

michelwi (commented):
some comments on everything but the new NN system.

common/setups/rasr/gmm_system.py
christophmluscher: Yes. But this is the GMM output. I thought that for sequence training we normally want a different LM for lattice generation. For example, Librispeech count LM for decoding is 4gram. For sequence training, this is a 2gram. or not?
That is why I did not want to automatically set the LM for training.. but I am open to handle it differently

common/setups/rasr/gmm_system.py
christophmluscher: I agree. I ll add this

common/setups/rasr/gmm_system.py
christophmluscher: yes. I want to have no decoding for the `None` case.
Idea is to first evaluate on the `dev` set. Find a good set of parameters and then pick/update the recog args accordingly

common/setups/rasr/gmm_system.py
michelwi: Yes, so for train set a 2gram LM would be exported.

This is not about automatically setting it, but about exporting it if set.

common/setups/rasr/util.py
```diff
@@ -8,8 +8,13 @@
     "GmmSatArgs",
     "GmmVtlnSatArgs",
     "ForcedAlignmentArgs",
+    "RecognitionArgs",
```
JackTemaki, line 11: Do you think it would make sense to separate this further? I imagine that GmmRecogntionArgs, HybridRecognitionArgs, CTCRegognitionArgs ect.... might be difficult.

common/setups/rasr/util.py
```diff
@@ -429,47 +470,239 @@ def __init__(
...
+    ):
+        self.oggzip_files = oggzip_files
+        self.alignments = alignments
+        self.context_window = (
+            context_window
+            if context_window is not None
+            else {"classes": 1, "data": 242}
+        )
+        self.audio = (
+            audio if audio is not None else {"features": "raw", "sample_rate": 16000}
```
JackTemaki, line 641: I think it is not good to have hardcoded values somehwere here.

common/setups/rasr/util.py
```diff
@@ -429,47 +470,239 @@ def __init__(
...
+
+class OggZipHdfDataInput:
+    def __init__(
+        self,
+        oggzip_files: tk.Path,
+        alignments: tk.Path,
+        context_window: Optional[Dict] = None,
+        audio: Optional[Dict] = None,
+        partition_epoch: int = 1,
+        seq_ordering: str = "laplace:226",
```
JackTemaki, line 630: this is a bad default, sequence ordering should be data size independent, so something like `laplace:.1000` (the dot says that this should be the bin size instead of the number of bins)

common/setups/rasr/util.py
```diff
@@ -429,47 +470,239 @@ def __init__(
...
+        targets: Optional[str] = None,
+    ):
+        self.oggzip_files = oggzip_files
+        self.alignments = alignments
+        self.context_window = (
+            context_window
+            if context_window is not None
+            else {"classes": 1, "data": 242}
```
JackTemaki, line 638: I do not understand what this is about.

common/setups/rasr/util.py
```diff
@@ -429,47 +470,239 @@ def __init__(
...
+            crp.language_model_config.file = self.lm["filename"]
+            crp.language_model_config.scale = self.lm["scale"]
+
+        if self.allophone_file is not None:
+            crp.acoustic_model_config.allophones.add_from_file = self.allophone_file
+
+        return crp
+
+
+class OggZipHdfDataInput:
```
JackTemaki, line 622: Can you already add docstrings to the new stuff? Then it would be much easier for me to understand why you are actually introducing this, and then it is also already better documented.

common/setups/rasr/util.py
```diff
@@ -309,6 +323,11 @@ def __init__(
         ##################################################
         :param recognition_args:
         ##################################################
+        :param test_recognition_args:
+        decoding parameters might change depending on whether dev or test sets are decoded.
+        setting test_recognition_args to None means to test recognition!
```
JackTemaki, line 328: to -> no?

JackTemaki (changes requested):
Maybe it is a good idea to split "Util" a little bit into the parts for which is actually needed. But this can be discussed.

common/setups/rasr/util.py
christophmluscher: I am thinking that maybe we could merge the `Gmm*Args` to `GmmStepArgs` because basically it is always the same steps and then make an documentation at the beginning of the file.

at the moment RecognitionArgs only holds the name plus a dict with the arguments.. I dont see a need to split this..

I generally agree that the `util.py` is a bit big

common/setups/rasr/util.py
christophmluscher: the values for the OggZipHdfData are taken from a setup by Peter. So for raw waveform setups. I can remove the defaults for `audio` and `context_window`

vieting: Yes, I think this default value should be removed.

common/setups/rasr/util.py
vieting: This one is more likely to be helpful but still, I agree we shouldn't hardcode it here.

common/setups/rasr/util.py
christophmluscher: yes I can

common/setups/rasr/util.py
christophmluscher: split maybe into `gmm_util`, `nn_util`, `ctc_util` or so?

common/setups/rasr/util.py
```diff
@@ -429,47 +470,242 @@ def __init__(
...
+        audio: Dict,
+        targets: str,
+        partition_epoch: int = 1,
+        seq_ordering: str = "laplace:.1000",
+    ):
+        """
+        :param oggzip_files:
+        :param alignments:
+        :param context_window: {"classes": 1, "data": 242}
+        :param audio: {"features": "raw", "sample_rate": 16000}
```
vieting, line 637: Maybe explain what this would mean
```suggestion
        :param audio: e.g. {"features": "raw", "sample_rate": 16000} for raw waveform input with a sample rate of 16 kHz
```

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        if "cropped" in network:
+            if network["output"]["from"] == ["cropped"]:
+                network["output"]["from"] = "upsample"
+            network.pop("cropped")
+            if "lstm_bwd_1" in network:
+                network["lstm_bwd_1"]["from"] = "upsample"
+                network["lstm_fwd_1"]["from"] = "upsample"
+            if "lstm_fwd_1_no_init" in network:
+                network["lstm_bwd_1_no_init"]["from"] = "upsample"
+                network["lstm_fwd_1_no_init"]["from"] = "upsample"
```
michelwi, line 136: this is ugly special handling with hardcoded layer names.. 

maybe rather allow for arbitrary user definable cleanup functions that are called on the network..

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+
+        self._init_am(**self.hybrid_init_args.am_args)
+
+        devtrain_data = devtrain_data if devtrain_data is not None else {}
+        dev_data = dev_data if dev_data is not None else {}
+        test_data = test_data if test_data is not None else {}
+
+        self._assert_corpus_name_unique(
+            train_data, devtrain_data, cv_data, dev_data, test_data
+        )
```
michelwi, line 248: does this mean cv_data == dev_data is not supportet?

michelwi: Also reordering for consistency reasons
```suggestion
        self._assert_corpus_name_unique(
            train_data, cv_data, devtrain_data, dev_data, test_data
        )
```

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+                raw_sample_rate=raw_sampling_rate,
+                feat_sample_rate=feature_sampling_rate,
+            )
+            self.oggzips[name] = j.out_ogg_zip
+            j.add_alias(f"oggzip/{name}")
+
+            # TODO self.jobs[name]["hdf_full"] = j = returnn.ReturnnDumpHDFJob()
+
+    def generate_lattices(self):
+        pass
```
michelwi, line 287: technically we could even do sequence training for GMMs. Just nobody since Zoltan has done this and I don't even have an example config for this.

So while in general this function could go into the `RasrSystem`, we can also leave it here.

michelwi: I can do a PR to add the lattices if you want. Just remind me some time ;)

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        self.nn_models[f"{train_corpus_key}_{cv_corpus_key}"][
+            name
+        ] = train_job.out_models
+        self.nn_configs[f"{train_corpus_key}_{cv_corpus_key}"][
+            name
+        ] = train_job.out_returnn_config_file
+
+        return train_job
+
+    def returnn_rasr_training(
```
michelwi, line 329: This function shares almost the same signature as `returnn_training` and does very similar things. 
Would it be possible / desirable to only have one function, which then internally decides on `ReturnnTrainingJob` vs `ReturnnRasrTrainingJob` and does the handling based on `*_corpus_key`.

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        self,
+        name: str,
+        returnn_config: returnn.ReturnnConfig,
+        checkpoints: List[returnn.Checkpoint],
+        acoustic_mixture_path: tk.Path,
+        prior_scale: float,
+        pronunciation_scales: List[float],
+        lm_scales: List[float],
+        optimize_am_lm_scale: bool,
+        corpus_key: str,
```
michelwi, line 380:
```suggestion
        recognition_corpus_key: str,
```
?

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        return train_job
+
+    # -------------------- Recognition --------------------
+
+    def nn_recognition(
+        self,
+        name: str,
+        returnn_config: returnn.ReturnnConfig,
+        checkpoints: List[returnn.Checkpoint],
+        acoustic_mixture_path: tk.Path,
```
michelwi, line 375: `acoustic_mixture_path` could be optional, if a prior file is provided.
(ok, not really.. we would still need a dummy mixture)

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        prior_scale: float,
```
michelwi, line 376: mh.. maybe `prior_scales` also deserves to be a list to be iterated over?

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+
+    def run_nn_step(self, step_args):
+        for trn_c, cv_c, dvtr_c, name in itertools.product(
+            self.train_corpora,
+            self.cv_corpora,
+            self.devtrain_corpora,
+            step_args.returnn_configs,
+        ):
+            if (trn_c, cv_c) in self.train_cv_pairing or self.train_cv_pairing is None:
+                continue
```
michelwi, line 455: do you really mean to skip over all combinations defined in the pairing or is there a `not` missing?

michelwi: I am wondering if iterating over `itertools.product(..)` is the right thing to do, or if in the step args a list of `(trn_c, cv_c, dvtr_c, name)` tuples can be defined that is iterated over.

michelwi (commented):
some comments on `rasr/nn_system.py`

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        self.datasets = {}
+
+        self.oggzips = {}
+
+        self.nn_configs = {}
+        self.nn_models = {}
+
+    # -------------------- Helpers --------------------
+    @staticmethod
+    def adapt_returnn_config_for_recog(returnn_config: returnn.ReturnnConfig):
```
JackTemaki, line 105: I think if this function is needed at all there are severe flaws in how we are building our configs. I am trying to solve the same issues right now for the CTC setup. I will give you am update when I have a clean solution for this.

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+            if "lstm_bwd_1" in network:
+                network["lstm_bwd_1"]["from"] = "upsample"
+                network["lstm_fwd_1"]["from"] = "upsample"
+            if "lstm_fwd_1_no_init" in network:
+                network["lstm_bwd_1_no_init"]["from"] = "upsample"
+                network["lstm_fwd_1_no_init"]["from"] = "upsample"
+
+        return config
+
+    @staticmethod
```
JackTemaki, line 140: I think this helper can be external at some point. Or actually I see a lot of overlap potential in my current CTCSystem vs the NNSystem here. I think it would be nice if the CTCSystem can fully extend from it. Then this would be fine here, but still people might want to build flows without the System class. All other flows are also defined standalone. (But there is also no problem in moving this later)

common/setups/rasr/nn_system.py
JackTemaki: > does this mean cv_data == dev_data is not supportet?

Unfortunately this is still impossible with the current Training jobs :/

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        self,
+        returnn_root: Optional[str] = None,
+        returnn_python_home: Optional[str] = None,
+        returnn_python_exe: Optional[str] = None,
+    ):
+        super().__init__()
+
+        self.returnn_root = returnn_root or gs.RETURNN_ROOT
+        self.returnn_python_home = returnn_python_home or gs.RETURNN_PYTHON_HOME
+        self.return_python_exe = returnn_python_exe or gs.RETURNN_PYTHON_EXE
```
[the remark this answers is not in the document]
christophmluscher: My thinking was that `ReturnnTrainingJob` and `ReturnnRasrTrainingJob` also both require setting `RETURNN_*` in `gs`. Then I would rather have the setup fail as soon as possible.
But I did not think about setting `returnn_` at a later stage, e.g. via `step_args`.
I think you are right to leave this open so that the user has flexibility.

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        self.devtrain_corpora = []
+        self.cv_corpora = []
+
+        self.train_input_data = None
+        self.cv_input_data = None
+        self.devtrain_input_data = None
+        self.dev_input_data = None
+        self.test_input_data = None
+
+        self.train_cv_pairing = None
```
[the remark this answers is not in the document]
christophmluscher: At the moment, you can add two train datasets and two cv datasets. This would result in 4 training jobs. If you want to fix train/cv pairs, this variable holds this information.

common/setups/rasr/nn_system.py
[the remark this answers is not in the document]
michelwi: There is no harm in having it fail here if it is unset. But then rather than throwing an `AttributeError` we should in this case assert that either is set.

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+
+        self.datasets = {}
+
+        self.oggzips = {}
```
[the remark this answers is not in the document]
christophmluscher: I am not sure how to handle this yet. WIP. At some point, I would like to have the possibility to automatically convert a dataset based on CRPs to a setup with OggZips and/or ... I think that would need the extra variables, as you said, but not fully thought through yet.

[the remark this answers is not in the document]
christophmluscher: One idea: use `datasets` for `ReturnnRasrTrainingJob`. Everything else for `ReturnnTrainingJob`. Or so

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+        Adapt a RETURNN config for recognition, e.g., remove loss and use log softmax activation in last layer
+
+        :param ReturnnConfig config:
+        :rtype ReturnnConfig:
+        """
+        assert isinstance(returnn_config, returnn.ReturnnConfig)
+        config = copy.deepcopy(returnn_config)
+        forward_output_layer = config.config.get("forward_output_layer", "output")
+        network = config.config.get("network")
+        for layer_name, layer in network.items():
```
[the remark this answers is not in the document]
christophmluscher: I agree. Based on comments by Nick. I ll wait to see his solution which should hopefully mean that we can remove this function

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,554 @@
...
+            if layer.get("unit", None) in {"lstmp"}:
+                layer["unit"] = "nativelstm2"
+            if layer.get("target", None) and layer_name != forward_output_layer:
+                layer.pop("target")
+                layer.pop("loss", None)
+                layer.pop("loss_scale", None)
+        network[forward_output_layer]["class"] = "linear"
+        network[forward_output_layer]["activation"] = "log_softmax"
+        network[forward_output_layer].pop("target", None)
```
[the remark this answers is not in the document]
christophmluscher: I dont know. Nick hopefully makes this function obsolete :)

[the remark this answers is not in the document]
christophmluscher: @JackTemaki no pressure :D

common/setups/rasr/nn_system.py
christophmluscher: true

common/setups/rasr/nn_system.py
christophmluscher: Ah then I would prefer to leave this function here :)

> I can do a PR to add the lattices if you want. Just remind me some time ;)

yes please :)

common/setups/rasr/nn_system.py
christophmluscher: maybe not sure :)

common/setups/rasr/nn_system.py
christophmluscher: yes. good point. that could be nicer and maybe more user-friendly..

common/setups/rasr/nn_system.py
christophmluscher: Yes, we should maybe discuss this on Tuesday, how CTCSystem and NNSystem are going to relate to each other.

maybe a `flow` dir in `setups`??

JackTemaki: maybe this could be in i6_core/returnn/flow.py or so...? I am not sure.

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,564 @@
...
+        self.test_input_data = None
+
+        self.train_cv_pairing = None
+
+        self.datasets = {}
+
+        self.oggzips = {}
+
+        self.nn_configs = {}
+        self.nn_models = {}
```
JackTemaki, line 97: I just saw that there is a big inconsistency with respect to "Models" and "Checkpoints". We should only use one, so that all stuff like "getting best epoch", "checkpoint averaging" and so on will be working. I think you switched to `Checkpoint` in your further code, and I would suggest to do this everywhere, also here.

common/setups/rasr/nn_system.py
```diff
@@ -0,0 +1,564 @@
...
+        )
+
+        tf_flow.config[tf_fwd].output_map.info_0.param_name = "log-posteriors"
+        tf_flow.config[
+            tf_fwd
+        ].output_map.info_0.tensor_name = f"{forward_output_layer}/output_batch_major"
+
+        tf_flow.config[tf_fwd].loader.type = "meta"
+        tf_flow.config[tf_fwd].loader.meta_graph_file = tf_graph_path
+        tf_flow.config[tf_fwd].loader.saved_model_file = checkpoint_path
```
JackTemaki, line 193: Are you sure this worked for you? I got some problems when directly passing the checkpoint, as the path had the `*.index` extension which RASR did not accept, so I needed to add a "delayed cutting" function to solve this.

michelwi: I think this was the problem with `Model` and we specifically designed `Checkpoint` such that this works..
Here (https://github.com/rwth-i6/i6_core/blob/deb2bbe5748e9f9e4e9c6740290928beb5f1edf9/returnn/training.py#L55) the additional `".index"` should be removed.

michelwi: >  I will set `ReturnnModel` to deprecated

I don't think I meant `ReturnnModel` (which contains a path, a config, and an epoch) but a different class that was effectively doing the same as `Checkpoint` but was hashing the absolute path to the model. (And was already removed)

> The `ckpt_path` can not be used here, because this already calls get_path()

But we would only set the `Checkpoint` object here. The call to `Checkpoint.str` (and subsequently `ckpt_path`) would only happen at runtime when the `create_files` Task of Recognition was executed; Therefore we _can_ use it here, right?

common/setups/rasr/nn_system.py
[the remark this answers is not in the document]
JackTemaki: Okay, so we can not completely drop this function I think, but there are some things that can be removed.

 - Targets do not to be dropped in RETURNN, for me I can always keep them.
 - I would consider using lstmp an user error, and not something to fix here
 - For the log outputs, they can be part of the network in the first place, and then we can tell the tf_flow helper what output to use. This would be more consistent and flexible (and safe heuristics code).
 
 For the "cropped" thing I would refer to @michelwi solution, as this is way too specific to be included here.

common/setups/rasr/nn_system.py
```diff
@@ -454,7 +454,7 @@ def nn_recognition(
         self,
         name: str,
         returnn_config: returnn.ReturnnConfig,
-        checkpoints: List[returnn.Checkpoint],
+        returnn_models: List[returnn.ReturnnModel],
```
JackTemaki: Why did you change the Checkpoints back to Model? Is there something the Checkpoint can not do? Because I remember that we discussed at some point that ReturnnModel should be deprecated because it is not supported by some jobs (e.g. CheckpointAveraging or GetBestCheckpoint).

christophmluscher: No, I wrote that down incorrectly. I ll revert back then

JackTemaki: But why was that incorrectly? I am using `Checkpoint` for my CTCSystem and this seems to be more consistent with the rest of the setups.

christophmluscher: I wrote down that Checkpoint was deprecated. I didn't recheck that. My mistake

christophmluscher: That was incorrect of me

JackTemaki (approved):
Lets merge this and then continue!

JackTemaki:
@christophmluscher you need to rebase the branch once on the master, there are merge conflicts!

michelwi:
> Lets merge this and then continue!

There are many changes in user folders of other users (rossenbach/dierkes). Is this intended or a rebase-bug?

christophmluscher:
> you need to rebase the branch once on the master, there are merge conflicts!

yes. will do

> There are many changes in user folders of other users (rossenbach/dierkes). Is this intended or a rebase-bug?

I did a merge from main, that was maybe not optimal. I ll look into this

PR #32, opened 2022-01-20
add _get_raw_bliss_lexicon

JackTemaki:
I did some code restructuring in preparation for an updated Librispeech Lexicon pipeline.

After building some setups it seems it was a bad idea to first merge the static symbols and then do the G2P, it would save a lot of job cases when this would be the other way around. This PR does not introduce this yet (a second get_g2p_augmented_lexicon), but just does some initial reformatting.

### DIFF
--- a/common/datasets/librispeech.py
+++ b/common/datasets/librispeech.py
@@ -304,7 +304,7 @@ def get_arpa_lm_dict(output_prefix="datasets"):
 
 
 @lru_cache()
-def get_special_lemma_lexicon(add_unknown_phoneme_and_mapping=True):
+def _get_special_lemma_lexicon(add_unknown_phoneme_and_mapping=True):
     """
     Generate the special lemmas for LibriSpeech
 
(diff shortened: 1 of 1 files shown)


JackTemaki:
@christophmluscher @Atticus1806 

Ping ;)

common/datasets/librispeech.py
```diff
@@ -354,6 +354,50 @@ def get_special_lemma_lexicon(add_unknown_phoneme_and_mapping=True):
     return lex
 
 
+@lru_cache()
+def _get_raw_bliss_lexicon(
+    use_stress_marker,
```
Atticus1806, line 359: Is there a reasonable default for this bool? Or is this intentionally not set? `get_bliss_lexicon` has `False` as default.

Atticus1806 (commented):
Looks good, besides this default everything is fine I think.

>Ping ;)
Pong.

christophmluscher (approved):

common/datasets/librispeech.py
christophmluscher: If you want to use this private function, you should think about what you want to do. so I think there is no reasonable default, IMHO

PR #33, opened 2022-01-26
Additions to GMM pipeline

christophmluscher:
- adds ngram recognition
- adds output creation for later NN training steps

### DIFF
--- a/common/setups/rasr/gmm_system.py
+++ b/common/setups/rasr/gmm_system.py
@@ -912,27 +912,14 @@ class GmmSystem(RasrSystem):
                 **step_args.training_args,
             )
 
-            for dev_c in self.dev_corpora:
-                name = step_args.training_args["name"]
-                feature_scorer = (trn_c, f"train_{name}")
+            name = step_args.training_args["name"]
+            feature_scorer = (trn_c, f"train_{name}")
 
-                self.recognition(
-                    name=f"{trn_c}-{name}",
-                    corpus_key=dev_c,
-                    feature_scorer_key=feature_scorer,
-                    **step_args.recognition_args,
-                )
-
-            for tst_c in self.test_corpora:
-                name = step_args.training_args["name"]
-                feature_scorer = (trn_c, f"train_{name}")
-
-                self.recognition(
-                    name=f"{trn_c}-{name}",
-                    corpus_key=tst_c,
-                    feature_scorer_key=feature_scorer,
-                    **step_args.recognition_args,
-                )
+            self.run_recognition_step(
+                step_args=step_args,
+                name=f"{trn_c}-{name}",
+                feature_scorer=feature_scorer,
+            )
 
             # ---------- SDM Mono ----------
             if step_args.sdm_args is not None:
@@ -948,27 +935,14 @@ class GmmSystem(RasrSystem):
                 **step_args.training_args,
             )
 
-            for dev_c in self.dev_corpora:
-                name = step_args.training_args["name"]
-                feature_scorer = (trn_c, f"train_{name}")
+            name = step_args.training_args["name"]
+            feature_scorer = (trn_c, f"train_{name}")
 
-                self.recognition(
-                    f"{trn_c}-{name}",
-                    corpus_key=dev_c,
-                    feature_scorer_key=feature_scorer,
-                    **step_args.recognition_args,
-                )
-
-            for tst_c in self.test_corpora:
-                name = step_args.training_args["name"]
-                feature_scorer = (trn_c, f"train_{name}")
-
-                self.recognition(
-                    name=f"{trn_c}-{name}",
-                    corpus_key=tst_c,
-                    feature_scorer_key=feature_scorer,
-                    **step_args.recognition_args,
-                )
+            self.run_recognition_step(
+                step_args=step_args,
+                name=f"{trn_c}-{name}",
+                feature_scorer=feature_scorer,
+            )
 
             # ---------- SDM Tri ----------
             if step_args.sdm_args is not None:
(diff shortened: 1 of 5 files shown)


common/setups/rasr/gmm_system.py
```diff
@@ -1007,6 +1074,63 @@ def run_vtln_sat_step(self, step_args):
...
+        assert (
+            name is None
+            and feature_scorer is None
+            and isinstance(step_args, RecognitionArgs)
+        ) ^ (
+            isinstance(name, str)
+            and isinstance(feature_scorer, Tuple)
+            and not isinstance(step_args, RecognitionArgs)
+        ), (
+            "please check that vairables are not specified in two places. type (name, feature_scorer, step_args):",
```
michelwi, line 1092: Typo
```suggestion
            "please check that variables are not specified in two places. type (name, feature_scorer, step_args):",
```

common/setups/rasr/gmm_system.py
```diff
@@ -1007,6 +1074,63 @@ def run_vtln_sat_step(self, step_args):
                     **step_args.sdm_args,
                 )
 
+    def run_recognition_step(
```
michelwi, line 1077: Can we maybe expand this step to incorporate also sat_recognition?

E.g. have a flag in the step_args  and then do smth like
```
recog_funktion = self.sat_recognition if step_args["sat"] else self.recognition
```

This way we could simplify / deduplify some code above in the `run_sat_step`

common/setups/rasr/gmm_system.py
```diff
@@ -826,6 +831,77 @@ def sat_recognition(
...
+        corpus_key: str,
+        corpus_type: str,
+        step_idx: int,
+        steps: RasrSteps,
+        extract_features: List,
+    ):
+        """
+        :param corpus_key: corpus name identifier
+        :param corpus_type: corpus used for: train, dev or test
+        :param step_idx: which iteration to take within one step
```
michelwi, line 847: Please clarify: "which iteration to take _within one step_" 
I would interpret it as e.g. select this iteration of align_split_acumulate_sequence of one specified GMM-Training-Step.

But below I see the code
```
steps.get_step_names_as_list()[step_idx - 1]
```
which reads to me like we select a specific step from the list of all defined steps.
(e.g. step_idx=1 -> monophone GMM; step_idx=3 -> triphone_sat GMM)

Which of these is true?

common/setups/rasr/util/rasr.py
```diff
@@ -0,0 +1,204 @@
...
+    - lm: file, scale, type
+    - concurrency
+    """
+
+    def __init__(
+        self,
+        corpus_object: meta.CorpusObject,
+        lexicon: dict,
+        lm: Optional[dict] = None,
+        concurrent: int = 10,
```
michelwi, line 29: Is 10 a reasonable default?

also
```suggestion
        concurrent: Optional[int] = 10,
```
right?

common/setups/rasr/util/rasr.py
```diff
@@ -0,0 +1,204 @@
...
+        self.corpus_object = corpus_object
+        self.lexicon = lexicon
+        self.lm = lm
+        self.concurrent = concurrent
+
+
+# when getting an zero weights error, set:
+#
+# extra_config = sprint.SprintConfig()
+# extra_config.allow_zero_weights = True
```
michelwi, line 46: I think you must set
```suggestion
# extra_config["*"].allow_zero_weights = True
```
otherwise it will be put in the very top ordering unit of the config and never be queried.
Bonus points if you find the correct identifier(s).

common/setups/rasr/util/rasr.py
```diff
@@ -0,0 +1,204 @@
...
+        :param lm: filename: Path, type: str, scale: float
+        :param concurrent: concurrency for gmm hmm pipeline
+        """
+        self.corpus_object = corpus_object
+        self.lexicon = lexicon
+        self.lm = lm
+        self.concurrent = concurrent
+
+
+# when getting an zero weights error, set:
```
michelwi, line 43: Is this even the right place for this information?
Maybe make this part of a doc string.

common/setups/rasr/util/rasr.py
christophmluscher: > Is 10 a reasonable default?

for a eval set, I think so. for training set maybe not..

> right?

No, I don't think so. concurrent variable needs to be set to an int. It is used directly in for example `SegmentCorpusJob` which does not accept a `None` value.

common/setups/rasr/util/rasr.py
christophmluscher: I think I verified that this works. But I'll double-check and it is definitively cleaner with the correct identifiers

common/setups/rasr/util/rasr.py
christophmluscher: no it is not :)

common/setups/rasr/util/rasr.py
michelwi: > for a eval set, I think so. 

ok.

> No, I don't think so. [...] which does not accept a None value

yes, it was a misunderstanding on my side about the meaning of `Optional`. pls. disregard comment. :)

common/setups/rasr/util/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .rasr import *
+from .gmm import *
+from .nn import *
```
JackTemaki, line 3: This does not exist yet

common/setups/rasr/util/gmm.py
```diff
@@ -1,186 +1,27 @@
...
-        :param scorer:
-        "kaldi", "sclite", default is sclite
-        ##################################################
-        """
-        self.costa_args = costa_args
-        self.default_mixture_scorer_args = default_mixture_scorer_args
-        self.scorer = scorer
-        self.am_args = am_args
-        self.feature_extraction_args = feature_extraction_args
+from .nn import ReturnnRasrDataInput
```
JackTemaki, line 24: Does not exist yet.

christophmluscher:
@JackTemaki can you approve and merge, please? :)

JackTemaki (approved):
There are still some small things that will break the current code, but I can fix this directly after merging.

PR #36, opened 2022-01-31
added job for cutting speech from audio based on transcription

dierkes-j:
@vieting This is my job for cutting speech from audio files

### DIFF
--- a/users/dierkes/preprocessing/segment_cutting.py
+++ b/users/dierkes/preprocessing/segment_cutting.py
@@ -1,10 +1,11 @@
 import soundfile as sf
 import numpy as np
-from i6_core.lib import corpus
 import multiprocessing
-import tqdm
+import logging
 
 from sisyphus import Job, Task, tk
+
+from i6_core.lib import corpus
 
 class CutAndStitchSpeechSegmentsFromCorpusJob(Job):
     """
@@ -79,7 +80,7 @@ class CutAndStitchSpeechSegmentsFromCorpusJob(Job):
         print(f"launching {self.n_workers} processes")
 
         tasks = [(r, self.out_audio_path, self.target_length, self.file_extension) for r in recordings]
-
         with multiprocessing.Pool(processes=self.n_workers) as pool:
-            for _ in tqdm.tqdm(pool.imap_unordered(self.cut_file, tasks), total=len(tasks)):
-                pass
+            for i, _ in enumerate(pool.imap_unordered(self.cut_file, tasks)):
+                if i % 100 == 0:
+                    logging.info(f"{i} of {len(tasks)} files done")
--- a/users/dierkes/preprocessing/segment_cutting.py
+++ b/users/dierkes/preprocessing/segment_cutting.py
@@ -6,15 +6,15 @@ import tqdm
 
 from sisyphus import Job, Task, tk
 
-class CutSpeechSegmentsFromAudio(Job):
+class CutAndStitchSpeechSegmentsFromCorpusJob(Job):
     """
-    This Job uses a transcription file to cut all segments containing speech from the audio files
+    This Job uses a bliss corpus to cut all segments containing speech from the audio files
     and stitches them together to files of a target length. Segments that are
     longer than the target_length are stored as individual files.
 
 
     """
-    def __init__(self, bliss_corpus, target_length=20, n_workers=16, file_extension='.wav'):
+    def __init__(self, bliss_corpus_file, target_length=20, n_workers=16, file_extension='wav'):
         """
 
         :param tk.Path bliss_corpus:
@@ -22,17 +22,17 @@ class CutSpeechSegmentsFromAudio(Job):
         :param int n_workers: number of threads for work on files in parallel
         :param str file_extension: currently supported are wav, flac and mp3
         """
-        self.bliss_corpus = bliss_corpus
+        self.bliss_corpus_file = bliss_corpus_file
         self.target_length = target_length
         self.n_workers = n_workers
         self.file_extension = file_extension
-        assert file_extension in ['.wav', '.flac', '.mp3']
+        assert file_extension in ['wav', 'flac', 'mp3']
 
         self.out_audio_path = self.output_path("audio/", directory=True)
-        self.cut_rqmt = {'time': 8, 'mem': 16, 'cpu': self.n_workers}
+        self.rqmt = {'time': 8, 'mem': 16, 'cpu': self.n_workers}
 
     def tasks(self):
-        yield Task("run", rqmt=self.cut_rqmt)
+        yield Task("run", rqmt=self.rqmt)
 
     def cut_file(self, task):
         recording, root_out, target_len_sec, extension = task
@@ -57,7 +57,7 @@ class CutSpeechSegmentsFromAudio(Job):
 
             # if a slice is longer than target_len_sec, we put it entirely in it's own piece
             if length_accumulated + (end - start) > target_len_sec and length_accumulated > 0:
-                file_out = f"{root_out}/{recording_name}_{i}{extension}"
+                file_out = f"{root_out}/{recording_name}_{i}.{extension}"
                 sf.write(file_out, np.hstack(to_stitch), samplerate=samplerate)
                 to_stitch = []
                 i += 1
@@ -67,13 +67,13 @@ class CutSpeechSegmentsFromAudio(Job):
             length_accumulated += end - start
 
         if to_stitch:
-            file_out = f"{root_out}/{recording_name}_{i}{extension}"
+            file_out = f"{root_out}/{recording_name}_{i}.{extension}"
             sf.write(file_out, np.hstack(to_stitch), samplerate=samplerate)
 
     def run(self):
-        self.corpus = corpus.Corpus()
-        self.corpus.load(self.bliss_corpus.get_path())
-        recordings = list(self.corpus.all_recordings())
+        self.corpus_object = corpus.Corpus()
+        self.corpus_object.load(self.bliss_corpus_file.get_path())
+        recordings = list(self.corpus_object.all_recordings())
 
         print(f"{len(recordings)} recordings detected")
         print(f"launching {self.n_workers} processes")
--- a/users/dierkes/preprocessing/segment_cutting.py
+++ b/users/dierkes/preprocessing/segment_cutting.py
@@ -1,0 +1,107 @@
+import pathlib
+import soundfile as sf
+import numpy as np
+import json
+import multiprocessing
+import argparse
+import tqdm
+import os
+import shutil
+import logging
+
+from sisyphus import Job, Task, tk
+
+class CutSpeechSegmentsFromAudio(Job):
+    """
+    This Job uses a transcription file to cut all segments containing speech from the audio files
+    and stitches them together to files of a target length. Segments that are
+    longer than the target_length are stored as individual files.
+
+
+    """
+    def __init__(self, audio_dir, segments_dir, target_length=20, n_workers=32, file_extension='.wav'):
+        """
+
+        :param audio_dir: path to all the audio files. Subfolders are not yet supported in audio
+            directory
+        :param segments_dir: path to the transcriptions. Subfolders are not yet supported in
+            transcription directory
+        :param target_length: target length of the stitched extracted speech
+        :param n_workers:
+        :param file_extension: currently supported are wav, flac and mp3
+        """
+        self.audio_dir = audio_dir
+        self.segments_dir = segments_dir
+        self.target_length = target_length
+        self.n_workers = n_workers
+        self.file_extension = file_extension
+        assert file_extension in ['.wav', '.flac', '.mp3']
+
+        self.out_audio_folder = self.output_path("audio/", directory=True)
+        self.cut_rqmt = {'time': 8, 'mem': 8, 'cpu': 4}
+
+    def tasks(self):
+        yield Task("cut_speech_segments", rqmt=self.cut_rqmt, mini_task=True)
+
+    def cut_file(self, task):
+        path_file, root_out, target_len_sec, extension, segments = task
+        file_name = path_file.stem
+        meta_file_path = f"{segments}/{file_name}.json"
+
+        with open(meta_file_path, 'r') as f:
+            meta = json.loads(f.read())["value"]["segments"]
+
+        to_cut = []
+        for seg in meta:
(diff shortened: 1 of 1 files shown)


users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if to_stitch:
+            file_out = f"{root_out}/{file_name}_{i}{extension}"
+            sf.write(file_out, np.hstack(to_stitch), samplerate=8000)
+
+    def cut(self, input_dir_audio, input_dir_segments, output_dir_audio,
+            target_len_sec=20, n_process=32, out_extension='.wav'):
+
+        list_dir = pathlib.Path(input_dir_audio).glob(f"**/*{out_extension}")
+        list_dir = [x for x in list_dir]
```
vieting, line 90:
```suggestion
        list_dir = list(list_dir)
```

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,107 @@
...
+
+            to_stitch.append(slice)
+            length_accumulated += end - start
+
+        if to_stitch:
+            file_out = f"{root_out}/{file_name}_{i}{extension}"
+            sf.write(file_out, np.hstack(to_stitch), samplerate=8000)
+
+    def cut(self, input_dir_audio, input_dir_segments, output_dir_audio,
+            target_len_sec=20, n_process=32, out_extension='.wav'):
```
vieting, line 87: I wouldn't introduce any defaults here if you already have them in the `__init__`.

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,107 @@
...
+        with open(meta_file_path, 'r') as f:
+            meta = json.loads(f.read())["value"]["segments"]
+
+        to_cut = []
+        for seg in meta:
+            to_cut += [(seg["start"], seg["end"])]
+
+        data, samplerate = sf.read(path_file)
+        assert len(data.shape) == 1
+        assert samplerate == 8000
```
vieting, line 60: In general, it would be good to have this configurable in the __init__ and then `assert samplerate == self.samplerate`.

vieting: Or actually not sure if we need to assert this at all. Does it really matter?

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,107 @@
...
+        length_accumulated = 0.0
+
+        i = 0
+        for start, end in to_cut:
+            start_index = int(start * samplerate)
+            end_index = int(end * samplerate)
+            slice = data[start_index:end_index]
+
+            # if a slice is longer than target_len_sec, we put it entirely in it's own piece
+            if length_accumulated + (end - start) > target_len_sec and length_accumulated > 0:
```
vieting, line 72: With this logic of dumping everything if we would exceed the target length, we might still get short segments if a very long segment follows. Maybe this is not important, just as a note.

vieting (changes requested):
Thanks for adding this. I noticed this is more specific to your data structure than I expected. It might be more helpful for other cases or users, to make this more generic in two ways:

1. Use a bliss corpus as input. I guess you have this in your pipeline later on anyway. Then you split all recordings based on the segments inside.
2. The job does two things: splitting and stitching. We could have two separate jobs for this. One just splits, the other one then stitches given the splits in the first job.

dierkes-j:
> The job does two things: splitting and stitching. We could have two separate jobs for this. One just splits, the other one then stitches given the splits in the first job.
But then we need to store all the splits of the first split job additionally, instead of just instantly combining them into the stitched files on the fly. Of course I could remove the stored files of the first split job when running the stitching job, but I am not sure if this is necessarily better.

vieting:
> > The job does two things: splitting and stitching. We could have two separate jobs for this. One just splits, the other one then stitches given the splits in the first job.
> 
> But then we need to store all the splits of the first split job additionally, instead of just instantly combining them into the stitched files on the fly. Of course I could remove the stored files of the first split job when running the stitching job, but I am not sure if this is necessarily better than doing it on the fly.

That's true, but as you say the unneeded files could be deleted. In general, sisyphus jobs should be atomic, so splitting here would make sense. However, for me it's ok if you leave it in one job here as it's experimental anyway.

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,107 @@
...
+import tqdm
+import os
+import shutil
+import logging
+
+from sisyphus import Job, Task, tk
+
+class CutSpeechSegmentsFromAudio(Job):
+    """
+    This Job uses a transcription file to cut all segments containing speech from the audio files
```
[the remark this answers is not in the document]
dierkes-j: I adjusted the job to work with a bliss corpus instead

users/dierkes/preprocessing/segment_cutting.py
dierkes-j: this is meant to be this way so that segments longer than the target_length will always be in an individual file

dierkes-j: It is actually meant to be this way so that segments longer than the target_length will always be in an individual file

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        self.audio_dir = audio_dir
+        self.segments_dir = segments_dir
+        self.target_length = target_length
+        self.n_workers = n_workers
+        self.file_extension = file_extension
+        assert file_extension in ['.wav', '.flac', '.mp3']
+
+        self.out_audio_folder = self.output_path("audio/", directory=True)
+        self.cut_rqmt = {'time': 8, 'mem': 8, 'cpu': 4}
```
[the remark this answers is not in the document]
dierkes-j: I am confused about this: If I schedule jobs with more than 4 cpu it tells me that those requirements cannot be met by the cluster. What does cpu mean here? _Actual_  cpu's or number of cores of a single cpu? I am using multiple processes in my code, so one cpu with multiple cores would already be sufficient to achieve parallelism.

dierkes-j:
I adjusted my code to work with a bliss corpus instead. 

> That's true, but as you say the unneeded files could be deleted. In general, sisyphus jobs should be atomic, so splitting here would make sense. However, for me it's ok if you leave it in one job here as it's experimental anyway.

Ok, then I would currently leave the logic as it is. When we consider pushing the code to a less experimental repository we can discuss the two job approach again.

users/dierkes/preprocessing/segment_cutting.py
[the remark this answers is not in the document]
JackTemaki: In that case you have wrong Sisyphus settings with respect to the SGE. Please check your `settings.py` file of your setup:
I have something like this:
```python
    return EngineSelector(engines={'short': LocalEngine(cpus=4),
                                   'long': SonOfGridEngine(default_rqmt={'cpu' : 1, 'mem' : 2, 'gpu' : 0, 'time' : 1})},
                                  default_engine='long')
```
so if for any reason you run the job on "short", it will not have more than 4 cpus. But the SGE is usually unlimited.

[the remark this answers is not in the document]
vieting: Regarding the number of cores per CPU, I'm not sure what a reasonabe assumption would be. Do you know @JackTemaki?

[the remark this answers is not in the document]
dierkes-j: I figured out the problem with the requirements not being met and fixed it

users/dierkes/preprocessing/segment_cutting.py
vieting: But they will only be in an individual file, if the buffer `to_stitch` is empty before, right?

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,85 @@
...
+        :param str file_extension: currently supported are wav, flac and mp3
+        """
+        self.bliss_corpus = bliss_corpus
+        self.target_length = target_length
+        self.n_workers = n_workers
+        self.file_extension = file_extension
+        assert file_extension in ['.wav', '.flac', '.mp3']
+
+        self.out_audio_path = self.output_path("audio/", directory=True)
+        self.cut_rqmt = {'time': 8, 'mem': 16, 'cpu': self.n_workers}
```
vieting, line 32:
```suggestion
        self.rqmt = {'time': 8, 'mem': 16, 'cpu': self.n_workers}
```

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def tasks(self):
+        yield Task("run", rqmt=self.cut_rqmt)
```
vieting, line 35:
```suggestion
        yield Task("run", rqmt=self.rqmt)
```

users/dierkes/preprocessing/segment_cutting.py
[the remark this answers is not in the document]
vieting: Please update to bliss corpus input

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,85 @@
...
+        :param tk.Path bliss_corpus:
+        :param int target_length: target length of the stitched extracted speech in seconds
+        :param int n_workers: number of threads for work on files in parallel
+        :param str file_extension: currently supported are wav, flac and mp3
+        """
+        self.bliss_corpus = bliss_corpus
+        self.target_length = target_length
+        self.n_workers = n_workers
+        self.file_extension = file_extension
+        assert file_extension in ['.wav', '.flac', '.mp3']
```
vieting, line 29: From other jobs, it seems like the convention would be to rather use `"wav"` instead of `".wav"`, see e.g. https://github.com/rwth-i6/i6_core/blob/main/corpus/transform.py

users/dierkes/preprocessing/segment_cutting.py
dierkes-j: Yes, but it will always be empty because we empty it before adding the new segment. If a segment exceeds the `target_length` we first store everything in `to_stitch` in a file and empty `to_stitch`. Then the new segmented is added to `to_stitch`. Now in case this segment is already larger than the `target_length` it will be stored in its own file in the next iteration of the loop. So segments longer than the `target_length` will always be in their own file.

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,85 @@
+import soundfile as sf
+import numpy as np
+from i6_core.lib import corpus
+import multiprocessing
+import tqdm
+
+from sisyphus import Job, Task, tk
+
+class CutSpeechSegmentsFromAudio(Job):
```
vieting, line 9: Maybe the name could reflect a bit better what the job does and it should end on `"Job"`, see https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTING.md

Maybe `CutAndStitchSpeechSegmentsFromCorpusJob`?

vieting (changes requested):
Some smaller comments remaining

users/dierkes/preprocessing/segment_cutting.py
vieting: Ah yes, I see.

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,85 @@
+import soundfile as sf
+import numpy as np
+from i6_core.lib import corpus
+import multiprocessing
+import tqdm
+
+from sisyphus import Job, Task, tk
```
vieting, line 7:
```suggestion
import multiprocessing
import numpy as np
import soundfile as sf
import tqdm

from sisyphus import Job, Task, tk

from i6_core.lib import corpus
```
See https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTING.md

users/dierkes/preprocessing/segment_cutting.py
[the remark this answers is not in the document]
vieting: @JackTemaki do you think this is alright? If so, I'd merge the PR.

users/dierkes/preprocessing/segment_cutting.py
```diff
@@ -0,0 +1,86 @@
...
+
+class CutAndStitchSpeechSegmentsFromCorpusJob(Job):
+    """
+    This Job uses a bliss corpus to cut all segments containing speech from the audio files
+    and stitches them together to files of a target length. Segments that are
+    longer than the target_length are stored as individual files.
+
+
+    """
+    def __init__(self, bliss_corpus_file, target_length=20, n_workers=16, file_extension='wav'):
```
JackTemaki, line 18: `n_workers` should not be hashed, as this has no effect on the output of the job. You can check other jobs (e.g. ReturnnTraining or so) on how to write custom hash functions.

JackTemaki (changes requested):

users/dierkes/preprocessing/segment_cutting.py
vieting: @JackTemaki also suggested offline to reduce the default `n_workers` to 4 to reduce the file server load.

vieting:
```suggestion
    def __init__(self, bliss_corpus_file, target_length=20, n_workers=4, file_extension='wav'):
```

dierkes-j: I am not sure about this change. Even if we are only using 4 cpus we can benefit of more workers running on multiple cores of the cpus, can't we? Is this also problematic because of the server load?

JackTemaki: The `cpu` entry for rqmt is about the cores (or more exactly, allowed number of simultaneously running threads), not about the number of actual CPUs. It is just that you do not want to make the job so fast that you have IO as bottleneck.

JackTemaki: So of you specify more workers than you have as allowed threads, you program will actually be slower and not faster, because there is an increased overhead in thread swapping.

vieting: Then we should also change `cpu` in `rqmt` to `n_workers`, right?

PR #44, opened 2022-02-14
Caching for FairseqHydraTrainingJob

dierkes-j:
@vieting @christophmluscher Here are my adjustments to support caching for the `FairseqTrainingJob`. A running job with the changes can be found here: `/work/asr3/vieting/hiwis/dierkes/hykist/i6_experiments/users/engler/fairseq/training/FairseqHydraTrainingJob.Oe5PeZuRXNcy`

### DIFF
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -193,7 +193,7 @@ class FairseqHydraTrainingJob(Job):
                 try:
                     cached_audio_fn = sp.check_output(["cf", f"{name}.bundle"]).strip().decode("utf8")
                 except sp.CalledProcessError:
-                    print(f"Cache manager: Error occurred for {name}, using local file")
+                    print(f"Cache manager: Error occurred for files in {name}")
                     raise
 
                 with open(cached_audio_fn) as local_bundle:
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -242,5 +242,10 @@ class FairseqHydraTrainingJob(Job):
     @classmethod
     def hash(cls, kwargs):
         d = copy.copy(kwargs)
-        d.pop("cached", None)
+        d.pop("use_cache_manager", None)
+        d.pop("time_rqmt", None)
+        d.pop("mem_rqmt", None)
+        d.pop("cpu_rqmt", None)
+        d.pop("gpu_rqmt", None)
         return super().hash(d)
+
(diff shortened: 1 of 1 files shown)


users/engler/fairseq/training.py
```diff
@@ -166,6 +181,27 @@ def create_files(self):
...
+                with open(f"{manifest_path}/{name}", "r") as manifest_file:
+                    manifest_lines = manifest_file.read().splitlines()
+                audio_path = manifest_lines[0]
+                bundle_lines = map(lambda line: audio_path + "/" + line.split("\t")[0], manifest_lines[1:])
+                with open(f"{name}.bundle", 'w') as bundle_file:
+                    bundle_file.write("\n".join(bundle_lines))
+                try:
+                    cached_audio_fn = sp.check_output(["cf", f"{name}.bundle"]).strip().decode("utf8")
+                except sp.CalledProcessError:
+                    print(f"Cache manager: Error occurred for {name}, using local file")
```
vieting, line 196: Does this make sense? If the caching fails, you'd still have the path of the cached file in the new manifest, so you'd actually not use the local file, right?

users/engler/fairseq/training.py
```diff
@@ -86,6 +98,7 @@ def __init__(
             (usually in the same folder as python exe '.../bin/')
         :param Path|str fairseq_root: File path to the fairseq git for alternative call of fairseq-hydra-train
             (no need to install fairseq here)
+        :param bool cached: enables caching of the data given in the manifest
```
vieting, line 101: Maybe make it explicit that the i6 cache manager is used. The variable could also be named `use_cache_manager` similar to RETURNN.

vieting (changes requested):
Looks good in general, just two small comments.

users/engler/fairseq/training.py
vieting: The message should also be adjusted because you're not using any local file. Printing some additional info before raising the error makes sense though.

vieting (approved):
Maybe let's wait if @christophmluscher wants to comment as well.

christophmluscher (changes requested):

users/engler/fairseq/training.py
```diff
@@ -198,3 +238,9 @@ def _get_run_cmd(self):
         if self.fairseq_python_exe is not None:
             run_cmd.insert(0, tk.uncached_path(self.fairseq_python_exe))
         return run_cmd
+
+    @classmethod
+    def hash(cls, kwargs):
```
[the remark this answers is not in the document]
christophmluscher: @vieting your opinion?

christophmluscher (approved):

PR #45, opened 2022-03-03
Job for archieving directories using zip and adjusted zip caching in Fairseq training

dierkes-j:
@vieting and @christophmluscher, this is my implementation for archiving a directory using zip as a Sisyphus job and using a zipped directory in the Fairseq training job.


users/engler/fairseq/training.py
```diff
@@ -183,25 +189,41 @@ def create_files(self):
...
+                        raise
+
+                    with open(cached_audio_fn) as local_bundle:
+                        bundle_lines = list(map(os.path.dirname, local_bundle.readlines()))
+                        assert bundle_lines.count(bundle_lines[0]) == len(bundle_lines), f"not all {name} files in same directory"
+                        manifest_lines[0] = bundle_lines[0]
+                    with open(f"{self.out_cached_audio_manifest.get_path()}/{name}", "w") as cached_audio_manifest_file:
+                        cached_audio_manifest_file.write('\n'.join(manifest_lines))
+            else:   # zipped audio data is given and we cache and unzip the zip file instead
+                if self.use_cache_manager:
```
vieting, line 213: This is redundant

users/engler/fairseq/training.py
```diff
@@ -183,25 +189,41 @@ def create_files(self):
...
+                    try:
+                        cached_audio_zip_dir = sp.check_output(["cf", self.zipped_audio_dir]).strip().decode("utf8")
+                        local_unzipped_audio_dir = f"{os.path.dirname(cached_audio_zip_dir)}/audio"
+                        sp.check_call(["unzip", "-q", "-n", cached_audio_zip_dir, "-d", local_unzipped_audio_dir])
+                    except sp.CalledProcessError:
+                        print(f"Cache manager: Error occurred for caching and unzipping audio data")
```
vieting, line 219:
```suggestion
                        print(f"Cache manager: Error occurred for caching and unzipping audio data in {self.zipped_audio_dir}")
```

users/engler/fairseq/training.py
```diff
@@ -183,25 +189,41 @@ def create_files(self):
...
+                        raise
+                    for name in ["train.tsv", "valid.tsv"]:
+                        with open(f"{manifest_path}/{name}", "r") as manifest_file:
+                            manifest_lines = manifest_file.read().splitlines()
+                        manifest_lines[0] = local_unzipped_audio_dir
+                        with open(f"{self.out_cached_audio_manifest.get_path()}/{name}", "w") as cached_audio_manifest_file:
+                            cached_audio_manifest_file.write('\n'.join(manifest_lines))
```
vieting, line 226: There is currently no check that the audio files contained in the zip file match the ones in the manifest. Maybe this is OK for now, but we should add a warning in the docstring then.

vieting (changes requested):
Minor comments for fairseq training job. I'll check the `DirectoryZipJob` once you change the file ending. It's easier to read with python syntax highlighting.

users/engler/fairseq/training.py
dierkes-j: Such a check should be simple to do. However I would only check that all files mentioned in the manifest exist in the unzipped directory and not the other way around. Do you think we should also check that all files in the unzipped directory exist in the manifest?

vieting (changes requested):
What about the warning regarding possibly mismatching manifest and zip dir? And could you please mark this comments that you fixed as resolved?

users/dierkes/tools/zip_tools.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        :param tk.Path dir_path: path to directory to be zipped
+        :param int compression_level: compression level for zipping, see zip documentation for
+            detailed description of possible compressions
+        :param list|None command_line_args: Additional command line arguments for configuring zip task
+        """
+        self.dir_path = dir_path
+        self.compression_level = compression_level
+        assert 0 <= self.compression_level <= 9 and type(self.compression_level) == int
+        self.command_line_args = command_line_args
```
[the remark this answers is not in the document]
dierkes-j: > And could you please mark this comments that you fixed as resolved?
I think I already marked all conversations that I resolved as resolved, no?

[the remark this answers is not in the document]
vieting: > I think I already marked all conversations that I resolved as resolved, no? For the comment about the warning I wrote a reply

Yes, got them now. Thanks.

users/engler/fairseq/training.py
vieting: No this one way should be enough as you create the manifest that is actually used based on the original manifest. So additional files in the zip would just be ignored.

vieting (approved):

PR #48, opened 2022-03-24
PrevCtm replace tuple by dataclass

JackTemaki:
As the tuple notation leaves room for errors or unclear interpretation I thought it is a good idea to replace this by a dataclass object.

In your configs you would have to change:

```python
        "prev_ctm": ("tri", 1.0, 22, 10, "-optlm"),
```
to the more verbose:
```python
        "prev_ctm": PrevCtm(
            prev_step_key="tri",
            pronunciation_scale=1.0,
            lm_scale=22,
            iteration=10,
            optimized_lm=True
        )
```

Hashes are untouched.

### DIFF
--- a/common/setups/rasr/util/gmm.py
+++ b/common/setups/rasr/util/gmm.py
@@ -3,6 +3,7 @@ __all__ = [
     "GmmCartArgs",
     "GmmTriphoneArgs",
     "GmmVtlnArgs",
+    "PrevCtm",
     "GmmSatArgs",
     "GmmVtlnSatArgs",
     "ForcedAlignmentArgs",
@@ -12,6 +13,7 @@ __all__ = [
 ]
 
 import copy
+from dataclasses import dataclass
 from typing import Dict, List, Optional, Tuple, Type, Union
 
 from sisyphus import tk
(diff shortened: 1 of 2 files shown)


common/setups/rasr/util/gmm.py
```diff
@@ -233,6 +235,26 @@ def __init__(
...
+class PrevCtm:
+    """
+    Defines a previous ctm output from a recognition which is used to extract the
+    speaker alignments for SAT adaptation for the dev/train sets.
+
+    :param prev_step_key: name of the step, e.g. "tri"
+    :param pronunciation_scale: one of the existing pronunciation scales
+    :param lm_scale: one of the existing lm-scales
+    :param iteration: one of the existing GMM iterations
+    :param use optimized recognition
```
michelwi, line 248:
```suggestion
    :param optimized_lm: use optimized recognition
```
missing parameter name

```suggestion
    :param optimized_lm: use automatically optimized lm.scale which might differ from param lm_scale 
```
better docstring suggestion

michelwi (approved):
> In your configs you would have to change:
[...]
to the more verbose:
[...]

Ideally all configs would be committed and you would already change it in the places where it occurs.

But as no configs of mine are touched and I agree with your sentiment, I approve.

Marvin84 (approved):

christophmluscher (approved):

PR #50, opened 2022-03-30
Black workflow

dierkes-j:
Because of a version incompatibility a newer version of black is required, see: https://github.com/psf/black/issues/2964

@vieting


vieting (approved):

michelwi (changes requested):
At i6_core we noticed the same an decided that as an immediate measure we will specify the last working version of click as a dependency to make the old version of black (21.4b2) work again.

Then, once we have assessed the changes needed for the new version of black and have installed it on the cluster, and have informed all users about it, we will raise it to 22.3.0

I think we should keep both repos synchronized about this.

cf https://github.com/rwth-i6/i6_core/pull/239

vieting:
Yes, good point. @dierkes-j can you please adapt accordingly?

michelwi (approved):
thanks.

PR #51, opened 2022-04-04
Added fairseq training task for plotting training results

dierkes-j:
We need to `os.walk` in the work directory because fairseq creates a new logging file whenever the training is restarted.
An exemplary plot looks like the following:
score_and_error
 
@vieting

### DIFF
--- a/users/engler/fairseq/training.py
+++ b/users/engler/fairseq/training.py
@@ -166,7 +166,8 @@ class FairseqHydraTrainingJob(Job):
             if k in self.keep_epochs
         }
         self.out_cached_audio_manifest = self.output_path("cached_audio_manifest", directory=True)
-        self.out_plot = self.output_path("score_and_error.png")
+        self.out_plot_se = self.output_path("score_and_error.png")
+        self.out_plot_lr = self.output_path("learning_rate.png")
 
         # Requirements:
         self.gpu_rqmt = gpu_rqmt
@@ -198,7 +199,7 @@ class FairseqHydraTrainingJob(Job):
                         manifest_lines = manifest_file.read().splitlines()
                     audio_path = manifest_lines[0]
                     bundle_lines = map(lambda line: audio_path + "/" + line.split("\t")[0], manifest_lines[1:])
-                    with open(f"{name}.bundle", 'w') as bundle_file:
+                    with open(f"{name}.bundle", "w") as bundle_file:
                         bundle_file.write("\n".join(bundle_lines))
                     try:
                         cached_audio_fn = sp.check_output(["cf", f"{name}.bundle"]).strip().decode("utf8")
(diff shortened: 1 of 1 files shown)


users/engler/fairseq/training.py
```diff
@@ -232,6 +235,69 @@ def run(self):
...
+                            try:
+                                epoch = int(split[split.index('epoch') + 1])
+                                score = float(split[split.index('train_loss') + 1])
+                                error = 1 - float(split[split.index('train_accuracy') + 1])
+                            except ValueError:
+                                continue
+                            train_score[epoch] = score
+                            train_error[epoch] = error
+                        elif not line:
+                            break
```
vieting, line 271:
```suggestion
```
See suggestion for conversion to for loop above.

vieting (changes requested):

users/engler/fairseq/training.py
```diff
@@ -232,6 +235,69 @@ def run(self):
...
+                    while True:
+                        line = f.readline()
+                        if 'begin validation on "valid" subset' in line:
+                            split = re.sub('[{,":]', '', f.readline()).split(' ')
+                            try:
+                                epoch = int(split[split.index('epoch') + 1])
+                                score = float(split[split.index('valid_loss') + 1])
+                                error = 1 - float(split[split.index('valid_accuracy') + 1])
+                            except ValueError:
+                                continue
```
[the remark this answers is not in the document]
dierkes-j: Yes, it can happen that the training crashes directly after logging the line `begin validation on "valid" subset`. Then the next line is empty and we get a `ValueError`. In this case we can just continue because the required values will be logged in the log file for the restarted training again. I think I should make this check more explicit by checking `f.readline()` being an empty string instead of catching the ValueError.

users/engler/fairseq/training.py
```diff
@@ -232,6 +235,69 @@ def run(self):
...
+        train_score, train_error = {}, {}
+        valid_score, valid_error = {}, {}
+
+        for cur in os.walk(directory):
+            dir_path = cur[0]
+            files = cur[2]
+            if 'hydra_train.log' in files:
+                with open(f'{dir_path}/hydra_train.log', 'r') as f:
+                    while True:
+                        line = f.readline()
```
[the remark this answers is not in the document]
dierkes-j: We can't just do this replacement. Once `f.readlines()` is called `f.readline()` returns only an empty string. So we would need to store the whole file in a list before the loop and then always access the next line in the list whenever we find training or validation data.

users/engler/fairseq/training.py
```diff
@@ -232,6 +236,82 @@ def run(self):
...
+        for cur in os.walk(directory):
+            dir_path = cur[0]
+            files = cur[2]
+            if "hydra_train.log" in files:
+                with open(f"{dir_path}/hydra_train.log", "r") as f:
+                    lines = f.readlines()
+                    for i in range(len(lines)-1):
+                        line = lines[i]
+                        if 'begin validation on "valid" subset' in line:
+                            split = re.sub('[{,":]', "", lines[i+1]).split(" ")
```
vieting, line 254:
```suggestion
                            split = re.sub('[{,":]', "", lines[i + 1]).split(" ")
```
I really wonder why this is not flagged by black...

users/engler/fairseq/training.py
```diff
@@ -232,6 +236,82 @@ def run(self):
...
+        valid_score, valid_error = {}, {}
+        learning_rates = {}
+
+        for cur in os.walk(directory):
+            dir_path = cur[0]
+            files = cur[2]
+            if "hydra_train.log" in files:
+                with open(f"{dir_path}/hydra_train.log", "r") as f:
+                    lines = f.readlines()
+                    for i in range(len(lines)-1):
```
vieting, line 251:
```suggestion
                    for i in range(len(lines) - 1):
```

users/engler/fairseq/training.py
```diff
@@ -232,6 +236,82 @@ def run(self):
...
+                                epoch = int(split[split.index("epoch") + 1])
+                                score = float(split[split.index("valid_loss") + 1])
+                                error = 1 - float(split[split.index("valid_accuracy") + 1])
+                            except ValueError:
+                                continue
+                            valid_score[epoch] = score
+                            valid_error[epoch] = error
+                            i += 1
+                        elif "end of epoch" in line:
+                            split = re.sub('[{,":]', "", lines[i+1]).split(" ")
```
vieting, line 265:
```suggestion
                            split = re.sub('[{,":]', "", lines[i + 1]).split(" ")
```

users/engler/fairseq/training.py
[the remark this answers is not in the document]
vieting: Yes, let's keep it but please add a brief comment on that.

vieting (changes requested):
Just some minor changes.

vieting:
Btw black is only checking the contents of `common/` which is why it didn't flag anything here.

vieting (approved):

PR #56, opened 2022-04-08
Job for automatic model conversion and dumping

dierkes-j:
For converting the official fairseq xlsr_53_56k.pt model with the newest fairseq version (commit: `31d94f556bd49bc7e61511adbda482b2c54652b5`) the following adjustments need to be made to the model first:

```
import torch
model = torch.load(chkpt_path)
model["extra_state"].pop("metrics")

model['model']['encoder.pos_conv.1.bias'] = model['model']['encoder.pos_conv.0.bias']
model['model']['encoder.pos_conv.1.weight_g'] = model['model']['encoder.pos_conv.0.weight_g']
model['model']['encoder.pos_conv.1.weight_v'] = model['model']['encoder.pos_conv.0.weight_v']
model['model'].pop('encoder.pos_conv.0.bias')
model['model'].pop('encoder.pos_conv.0.weight_g')
model['model'].pop('encoder.pos_conv.0.weight_v')

del model['cfg']['task'].eval_wer
del model['cfg']['task'].eval_wer_config
del model['cfg']['task'].eval_wer_tokenizer
del model['cfg']['task'].eval_wer_post_process
del model['cfg']['task'].autoregressive

torch.save(model, new_chkpt_path)
```

Additionally in the fairseq code the line `from .fp32_instance_norm import Fp32InstanceNorm` needs to be removed in `modules/__init__.py` to be compatible with the current version of pytorch_2_returnn. 

@vieting


users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+        self.device = device
+        assert self.device in ["cpu", "gpu"]
+        self.validate_atol = validate_atol
+        self.fairseq_root = fairseq_root
+        self.pytorch_to_returnn_root = pytorch_to_returnn_root
+
+        if not self.fairseq_root:
+            print("WARNING: no explicit fairseq root directory given")
+        if not self.pytorch_to_returnn_root:
+            print("WARNING: no explicit pytorch_to_returnn root directory given")
```
vieting, line 126: Right now I don't see a case where you would not set these, so you could also require that they are given.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+
+        self.returnn_model_out = self.output_path(self.output_name)
```
vieting, line 128:
```suggestion
        self.out_returnn_model = self.output_path(self.output_name)
```
see contributing guidelines

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+
+class FairseqWav2VecModelConvertAndDumpJob(Job):
+    """
+    Conversion Job for fairseq model checkpoints to RETURNN model dictionary and checkpoint
+    """
+
+    def __init__(
+        self,
+        wav2vec_config,
+        output_name,
```
vieting, line 97: Is it necessary to have this configurable? From my point of view, something generic like `"checkpoint"` would be fine.

users/dierkes/preprocessing/wav2vec.py
vieting: Also, I would distinguish between the checkpoint and the config file with the network dict. Please create output variables for both of them. The checkpoint should be a `Checkpoint`, see `returnn/training.py`.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+        log.Verbosity = 10
+        setup_tf_thread_pools()
+        print_available_devices()
+
+        if self.fairseq_root:
+            sys.path.append(self.fairseq_root.get_path())
+        if self.pytorch_to_returnn_root:
+            sys.path.append(self.pytorch_to_returnn_root.get_path())
+
+        def model_func(wrapped_import, inputs: torch.Tensor):
```
vieting, line 152: This could be given to the job, see general comment.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+
+            print("WARNING: simplify model by setting mask=False")
+            return w2v2_model(inputs, mask=False, features_only=True)["x"]
+
+        print("Load audio")
+        print("  ", self.audio_input.get_path())
+        audio, fr = soundfile.read(self.audio_input.get_path())
+        audio = np.reshape(audio, [4, len(audio) // 4]).astype(
+            "float32"
+        )  # dim (B, T) -> (4, *)
```
vieting, line 199: We could also make this more generic. The input (not necessarily audio btw) could be an input to the job as a numpy array. Then, we'd just do the `torch.from_numpy` here.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+                "shape": (None,),
+                "batch_dim_axis": 0,
+                "time_dim_axis": 1,
+                "feature_dim_axis": None,
+            },
+            returnn_dummy_input_shape=audio.shape,
+            export_tf_checkpoint_save_path=self.returnn_model_out.get_path(),
+            validate_allclose_kwargs={"atol": self.validate_atol}
+            if self.validate_atol
+            else None,
```
vieting, line 221: This could also be given to the job as `converter_kwargs` or similar.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+        )
+        with open(
+            self.returnn_model_out.get_path() + ".network.dict", "wt", encoding="utf-8"
```
vieting, line 224:
```suggestion
            self.out_returnn_config.get_path(), "wt", encoding="utf-8"
```
Or similar, but directly using the output path for the config, see comment above.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+        ) as f:
+            f.write(converter.get_returnn_config_serialized())
+        os.rename(
+            self.returnn_model_out.get_path() + ".network.dict",
+            self.returnn_model_out.get_path() + ".network.dict.py",
+        )
```
vieting, line 230:
```suggestion
```
Not needed if you use the output path directly above.

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -60,8 +79,155 @@ def run(self):
...
+
+        features = model_func(None, inputs=audio_pt)
+        print(f"  output shape PyTorch: {features.shape}")
```
vieting, line 233: This doesn't seem to be necessary, this is printed by the converter anyway I think.

vieting (changes requested):
I think it would make sense to have a general `ConvertPytorchToReturnnJob`. It could be located in the same dir as the `FairseqHydraTrainingJob` in `conversion.py` or so. This would be of general help, at least I would have another use case where I would like to use it. It could be more general that this one here which is tailored towards wav2vec. The `model_func` could be an input to the job so it would be usable for other models as well.

It should be possible to determine the python exe used. I'd suggest to do this similar to the `ReturnnTrainingJob`, so writing a bash script that is then called. If we have the `model_func` as input to the job, you'd probably also write a python file that is then called by the bash script.

vieting:
Maybe you could make use of the `ReturnnConfig` here to create the python script that would then be called by the bash script. Something like
```
class ConvertPytorchToReturnnJob:
  def __init__(self, model_func, inputs, conversion_kwargs):
    self._config = ReturnnConfig(
      config={
        "inputs": inputs,
        "conversion_kwargs": conversion_kwargs},
      python_epilog=[
        model_func,
        main,
        "if __name__ == '__main__':\n main()"])


def main():
  converter = verify_torch_and_convert_to_returnn(
    model_func,
    inputs=inputs,
    ...)
```

Then you would have a task `create_files` where you would call `self._config.write()`.

users/dierkes/preprocessing/wav2vec.py
dierkes-j: If the libraries are installed in the used python executable they don't need to be explicitly specified, do they?

dierkes-j: If the libraries are already installed in the python executable they don't need to be given explicitly, do they?

dierkes-j:
> > Additionally in the fairseq code the line from .fp32_instance_norm import Fp32InstanceNorm needs to be removed in modules/**init**.py to be compatible with the current version of pytorch_2_returnn.
> 
> Do you have details on why this is the case? In [rwth-i6/pytorch-to-returnn#122](https://github.com/rwth-i6/pytorch-to-returnn/pull/122) it sounded like the current main branch would already work. It would be great if we could fix that as well.

Instance norm is missing in `pytorch2returnn`. In [rwth-i6/pytorch-to-returnn#122](https://github.com/rwth-i6/pytorch-to-returnn/pull/122) I referred explicitly to the conversion compatibility of the newest fairseq version for wav2vec2. For a different conversion case this might not be the case anymore

dierkes-j:
> I think it would make sense to have a general `ConvertPytorchToReturnnJob`. It could be located in the same dir as the `FairseqHydraTrainingJob` in `conversion.py` or so. This would be of general help, at least I would have another use case where I would like to use it. It could be more general that this one here which is tailored towards wav2vec. The `model_func` could be an input to the job so it would be usable for other models as well.
> 
> It should be possible to determine the python exe used. I'd suggest to do this similar to the `ReturnnTrainingJob`, so writing a bash script that is then called. If we have the `model_func` as input to the job, you'd probably also write a python file that is then called by the bash script.

Ok, I will adjust the job.

users/dierkes/preprocessing/wav2vec.py
vieting: True, then just keep it.

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,216 @@
...
+        model_func_caller,
+        inputs=input,
+        returnn_dummy_input_shape=input.shape,
+        export_tf_checkpoint_save_path=out_returnn_checkpoint,
+        **converter_kwargs,
+    )
+    with open(out_returnn_model_dict, "wt", encoding="utf-8") as f:
+        f.write(converter.get_returnn_config_serialized())
+
+    model_func(None, inputs=input_pt)
```
vieting, line 67: What is this needed for?

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,216 @@
...
+        :param numpy.ndarray: input used for forwarding through the model
+        :param str device: cpu or gpu
+        :param dict converter_kwargs: additional parameters to converter
+        :param tk.Path conversion_python_exe: python exe being used for conversion
+        :param tk.Path|None fairseq_root: path to fairseq version to use for conversion
+        :param tk.Path|None pytorch_to_returnn_root: path to pytorch_to_returnn version to use for conversion
+        :param tk.Path|None returnn_root: returnn root for conversion
+        """
+        self.pytorch_config = pytorch_config
+        self.model_func = model_func
```
vieting, line 101: It's not really necessary to add them ass class attributes, right?

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,216 @@
...
+        pytorch_to_returnn_root=None,
+        returnn_root=None,
+        **kwargs,
+    ):
+        config = ReturnnConfig(
+            config={
+                "pytorch_config": pytorch_config.get_path(),
+                "converter_kwargs": converter_kwargs,
+                "fairseq_root": fairseq_root.get_path(),
+                "pytorch_to_returnn_root": pytorch_to_returnn_root.get_path(),
```
vieting, line 177: This could be `None`.

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,216 @@
...
+            pytorch_config=pytorch_config,
+            model_func=self.model_func,
+            converter_kwargs=converter_kwargs,
+            fairseq_root=fairseq_root,
+            pytorch_to_returnn_root=pytorch_to_returnn_root,
+            returnn_root=returnn_root,
+        )
+
+        self.out_conversion_config_file = self.output_path("conversion.config")
+        self.out_model_func_file = self.output_path("model_func.py")
```
vieting, line 127: This is not used, right?

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,216 @@
...
+        self._config = self.get_returnn_config(
+            pytorch_config=pytorch_config,
+            model_func=self.model_func,
+            converter_kwargs=converter_kwargs,
+            fairseq_root=fairseq_root,
+            pytorch_to_returnn_root=pytorch_to_returnn_root,
+            returnn_root=returnn_root,
+        )
+
+        self.out_conversion_config_file = self.output_path("conversion.config")
```
vieting, line 126:
```suggestion
        self.out_conversion_config_file = self.output_path("conversion.py")
```

vieting (changes requested):

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,216 @@
...
+
+    better_exchook.install()
+    debug_register_better_repr()
+    log.initialize(verbosity=[5])
+    log.Verbosity = 10
+    setup_tf_thread_pools()
+    print_available_devices()
+
+    def model_func_caller(wrapped_import, inputs):
+        return eval(model_func_name)(wrapped_import, inputs, pytorch_config)
```
vieting, line 44: Why is this needed? To allow for arbitrary names instead of `model_func`?

dierkes-j: yes, the input function is not limited to be named `model_func`

users/engler/fairseq/conversion.py
dierkes-j: true

users/engler/fairseq/conversion.py
vieting:
```suggestion
    model_func = globals()[model_func_name]
```
This seems easier to me, what do you think?

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,207 @@
...
+        export_tf_checkpoint_save_path=out_returnn_checkpoint,
+        **converter_kwargs,
+    )
+    with open(out_returnn_model_dict, "wt", encoding="utf-8") as f:
+        f.write(converter.get_returnn_config_serialized())
+
+
+class ConvertPytorchToReturnnJob(Job):
+    """
+    Conversion Job for fairseq model checkpoints to RETURNN model dictionary and checkpoint
```
vieting, line 70:
```suggestion
    Conversion Job for PyTorch model checkpoints to RETURNN model dictionary and checkpoint
```

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -2,16 +2,35 @@
...
+import better_exchook
 
 from sisyphus import Job, Task, tk
 
+from returnn.tf.util.basic import (
+    debug_register_better_repr,
+    setup_tf_thread_pools,
+    print_available_devices,
+)
+from returnn.log import log
```
vieting, line 18: Is this really needed? There is the same import in `main`.

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,207 @@
...
+            needs to be passed as a string
+        :param numpy.ndarray: input used for forwarding through the model
+        :param str device: cpu or gpu
+        :param dict converter_kwargs: additional parameters to converter
+        :param tk.Path conversion_python_exe: python exe being used for conversion
+        :param tk.Path|None fairseq_root: path to fairseq version to use for conversion
+        :param tk.Path|None pytorch_to_returnn_root: path to pytorch_to_returnn version to use for conversion
+        :param tk.Path|None returnn_root: returnn root for conversion
+        """
+        assert type(model_func) == str, "model function is required to be string"
```
vieting, line 98:
```suggestion
        assert isinstance(model_func, str), "model function is required to be string, see docstring"
```

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", rqmt=self.rqmt)
+
+    def create_files(self):
+        self._config.write(self.out_conversion_config_file.get_path())
+        np.save("numpy_input", self.input)
+
```
vieting, line 138:
```suggestion
```

users/engler/fairseq/conversion.py
```diff
@@ -0,0 +1,207 @@
...
+            ],
+            hash_full_python_code=True,
+        )
+        return config
+
+    @classmethod
+    def hash(cls, kwargs):
+        d = {
+            "conversion_config": cls.get_returnn_config(**kwargs),
+            "returnn_root": kwargs["returnn_root"],
```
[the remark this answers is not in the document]
dierkes-j: true

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -2,16 +2,28 @@
 import glob
 import os
 import random
+import numpy as np
+import typing
+import torch
+import sys
+import better_exchook
```
vieting, line 9: This is not needed, right? See previous comment.

vieting:
It can be merged except for the changes in `users/dierkes/preprocessing/wav2vec.py` from my point of view. Please just push them separately if it's only black formatting, otherwise create a new PR if needed.

PR #57, opened 2022-04-09
pipelines and configs, e2e and hybrid Hmm/Dnn (dirty)

tbscode:


tbscode:
This still needs a lot of cleanup and restructuring. 
Mainly:
- remove all parameters and logic from configs ( move to some `get_config_args.py` or similar )
- remove duplicate functions in configs ( move to some `config_function_helpers.py` or similar )
- cleanup network generation code ( use only *one* conformer_network.py as generator instead of dispatching different versions via network_builders/layers/* )

Once this is done the configs files in `experiments/*` can be deleted and replace by a single config.py.

Merging current version now, comment is mainly for reference.

tbscode:
> Not want to much duplicate files, only commiting baseline config for now.

Done, merging now.

PR #59, opened 2022-05-02
extended manifest creation job to multilingual manifests and adjusted…

dierkes-j:
… fairseq training job accordingly

@christophmluscher

### DIFF
--- a/users/dierkes/preprocessing/wav2vec.py
+++ b/users/dierkes/preprocessing/wav2vec.py
@@ -37,7 +37,7 @@ class FairseqAudioManifestCreationJob(Job):
         :param float upsampling_alpha: specifies how much to upsample low resource languages. Upsampling
             calculation is done according to "Unsupervised Cross-Lingual Representation Learning for Speech
             Recognition", see  arXiv:2006.13979v2
-        :param [str]|str|None manifest_audio_paths: explicitly specifies output paths in manifest for each
+        :param [str]|[tk.Path]|str|tk.Path|None manifest_audio_paths: explicitly specifies output paths in manifest for each
             audio directory respectively. Allows to use different paths in the manifest than in the audio_dir_paths
         """
         if isinstance(audio_dir_path, tk.Path):
@@ -53,7 +53,7 @@ class FairseqAudioManifestCreationJob(Job):
             ), "Only one audio directory is given, upsampling not possible"
 
         if manifest_audio_paths:
-            if isinstance(manifest_audio_paths, tk.Path):
+            if not isinstance(manifest_audio_paths, list):
                 manifest_audio_paths = [manifest_audio_paths]
             assert len(manifest_audio_paths) == len(self.audio_dir_paths)
             self.manifest_audio_paths = [
--- a/users/dierkes/preprocessing/wav2vec.py
+++ b/users/dierkes/preprocessing/wav2vec.py
@@ -2,66 +2,216 @@ import soundfile
 import glob
 import os
 import random
+import numpy as np
 
 from sisyphus import Job, Task, tk
+
 
 class FairseqAudioManifestCreationJob(Job):
     """
     Creates required manifest files for wav2vec pretraining with fairseq. For the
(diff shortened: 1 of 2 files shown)


users/dierkes/preprocessing/wav2vec.py
```diff
@@ -10,58 +11,206 @@ class FairseqAudioManifestCreationJob(Job):
...
+            self.audio_dir_paths = [audio_dir_path]
+        else:
+            assert isinstance(audio_dir_path, list)
+            self.audio_dir_paths = audio_dir_path
+        assert min([isinstance(path, tk.Path) for path in self.audio_dir_paths])
+
+        if len(self.audio_dir_paths) == 1:
+            assert (
+                upsampling_alpha is None
+            ), "Only one audio directory is given, upsampling not possible"
```
christophmluscher, line 53: I don't understand, why is upsampling not allowed if only one audio dir given?

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -10,58 +11,206 @@ class FairseqAudioManifestCreationJob(Job):
...
+        assert 0.0 <= self.valid_percent <= 1.0
         self.seed = seed
         self.path_must_contain = path_must_contain
+        self.upsampling_alpha = upsampling_alpha
+        if self.upsampling_alpha is not None:
+            assert 0.0 <= self.upsampling_alpha <= 1.0
 
         self.out_manifest_path = self.output_path("manifest/", directory=True)
-        self.rqmt = {"time": 2, "mem": 8, "cpu": 1}
+        self.rqmt = {"time": 6, "mem": 8, "cpu": 1}
```
christophmluscher, line 75: does it make sense to make `time` configurable for the user?

users/dierkes/preprocessing/wav2vec.py
```diff
@@ -10,58 +11,206 @@ class FairseqAudioManifestCreationJob(Job):
...
+            self.audio_dir_paths = audio_dir_path
+        assert min([isinstance(path, tk.Path) for path in self.audio_dir_paths])
+
+        if len(self.audio_dir_paths) == 1:
+            assert (
+                upsampling_alpha is None
+            ), "Only one audio directory is given, upsampling not possible"
+
+        if manifest_audio_paths:
+            if isinstance(manifest_audio_paths, tk.Path):
```
christophmluscher, line 56: `manifest_audio_paths` doc is `[str]|str|None`. why check if it is `Path`?

users/dierkes/preprocessing/wav2vec.py
dierkes-j: The upsampling is always done with respect to the amount of all languages. The less data given for a language the more it is upsampled compared to the other languages. If only one languages is given the upsampling factor computed is always one however for whatever alpha value is given (so no upsampling is done). I feel like it would be misleading to allow the user to specify an alpha value which does not influence the outcome of the algorithm at all. Also it would change the hash of the job without changing its output.

Do you think I should make this more clear in the assertion or in the doc-string?

users/dierkes/preprocessing/wav2vec.py
dierkes-j: I am not sure actually. I think 6 hours should cover most of the use cases so I would make it configurable for the user. I tested the job with a combined amount of ~1100 hours of speech.

users/dierkes/preprocessing/wav2vec.py
dierkes-j: Good catch! I fixed this

users/dierkes/preprocessing/wav2vec.py
christophmluscher: I would add to the doc string

christophmluscher:
@dierkes-j I think this can be merged now

users/dierkes/preprocessing/wav2vec.py
dierkes-j: I added to the doc string

dierkes-j:
> @dierkes-j I think this can be merged now

I agree! I used the changed code now for some time and it seems to work fine.

Issue #63, opened 2022-05-03
How to integrate `returnn_common.nn` into Sisyphus pipelines


JackTemaki:
When trying to create Returnn networks with `returnn_common.nn`, the following issues or questions appeared:

If `returnn_common.nn` would be used for construction within the manager:
 - A specific RETURNN and Tensorflow dependency is added to the manager environment
 - Loading Tensorflow slows down manager startup time significantly (especially when there is a slow fs somewhere)
 - The network construction logic is quite complex, so with a rising number of network constructions this will become slow (lets say 100 experiments with 20 pre-training stages each = 2000 construction calls for a manager/console startup).
 - The `returnn_common` version can be "pinned" before updating by using the git-hashed based python import call from `Returnn`, but the `Returnn` version itself can not be pinned. Nevertheless the construction might be influenced by the Returnn version, making the correct handling of network hashes as we are used to very difficult up to impossible.

As we will not solve the hashing issue without spending way to much time not doing experiments, I would propose to go with a completely different approach than we did before, which is:

-> **Create a job which takes a executable and some pre-defined parameters does the network construction as task instead of in the manager.** 

This of course would mean that the construction result is not hashed, but only the constructor location and the parameters. To be able to find potentially unwanted changes and to be able to reconstruct the networks exactly, the job should create a local copy of the used construction code at runtime (returnn and returnn_common can of course correctly be identified by their git hash). 

While this would require some manual management of the network construction to not make logic changing mistakes, I currently do not see any suitable alternative to this.

Atticus1806:
I like this approach. A mistake in the building of the config which was seen too late can then be fixed by calling `tk.remove_job_and_descendants` on the construction job

albertz:
> A specific RETURNN and Tensorflow dependency is added to the manager environment

I don't see how this is relevant.

The RETURNN version should not have any influence on `returnn_common.nn` at all (as long as new enough but otherwise it would simply crash with import errors, e.g. when `Dim` is not available).

Also the TensorFlow version should not have any influence on `returnn_common.nn`.

> Loading Tensorflow slows down manager startup time significantly (especially when there is a slow fs somewhere)

We already discussed about that here: https://github.com/rwth-i6/returnn_common/issues/27

`returnn_common.nn` itself really uses the nesting logic from TF, which we could also replace.

`returnn_common.nn` also uses some data structures from RETURNN like `Data` and `Dim`. In principle, they could also be independent from TF. But this should be done on RETURNN side.

[...]

PR #66, opened 2022-06-10
add initial nnet_constructor code

JackTemaki:
Related to #63 

This is initial code I used for testing `returnn_common` integration, but we can change this as we see fit.

### DIFF
--- a/common/setups/returnn_common/nnet_constructor.py
+++ b/common/setups/returnn_common/nnet_constructor.py
@@ -87,7 +87,7 @@ class DataInitArgs:
     :param available_for_inference: if this data is available during decoding/forward pass etc...
     :param dim_tags: list of dim tags representing an axis of the data, without batch or a hidden sparse dim
     :param sparse_dim: provide this dim to make the data sparse and define
-    :param has_batch_dim: set to False do create data without a batch dim (probably never needed)
+    :param has_batch_dim: set to False to create data without a batch dim (probably never needed)
     """
 
     name: str
@@ -135,10 +135,10 @@ class ReturnnCommonSerializer(DelayedBase):
     ):
         """
 
-        :param serializer_objects: all serializer objects which are serialized into a string  in order
-        :param returnn_common_root:
-        :param packages:
-        :param make_local_package_copy:
+        :param serializer_objects: all serializer objects which are serialized into a string in order
+        :param returnn_common_root: Path to returnn_common, if None, assumes direct import is fine
+        :param packages: Path to packages to import, if None, tries to extract them from serializer_objects
+        :param make_local_package_copy: whether to make a local copy of imported code into the Job directory 
         """
         super().__init__(None)
         self.serializer_objects = serializer_objects
@@ -301,7 +301,6 @@ class ReturnnCommonImport(SerializerObject):
         code_object: str,
     ):
         """
-
         :param code_object: path to a python object, e.g. `i6_experiments.users.username.my_rc_files.SomeNiceASRModel`
         """
         super().__init__()
--- a/common/setups/returnn_common/nnet_constructor.py
+++ b/common/setups/returnn_common/nnet_constructor.py
@@ -1,0 +1,402 @@
+"""
+Contains code that is necessary/helpful to generate Returnn Configs that make use of returnn_common
+
+Usage Example:
+
+    rc_some_unhashed_code = NonhashedCode(code=some_code)
+    rc_extern_data = ReturnnCommonExternData(extern_data=extern_data)
+    rc_model = ReturnnCommonImport(
+        "i6_experiments.users.some_user.common_modules.some_module.SomeNnModuleModel")
+    rc_construction_code = ReturnnCommonImport(
+        "i6_experiments.users.some_user.common_modules.constructor_module.some_constructor_function")
+
+    rc_network = ReturnnCommonDynamicNetwork(
+        net_func_name=rc_construction_code.get_name(),
+        net_func_map={"net_module": rc_model.get_name(),
+                      "audio_data": "audio_features",
+                      "label_data": "bpe_labels",
+                      "audio_feature_dim": "audio_features_feature",
+                      "audio_time_dim": "audio_features_time",
+                      "label_time_dim": "bpe_labels_time",
+                      "label_dim": "bpe_labels_indices"
+                     },
+        net_kwargs={... some additional constructor function args ...}
+    )
+
+    serializer = ReturnnCommonSerializer(
+        serializer_objects=[rc_recursionlimit,
+                            rc_extern_data,
+                            rc_model,
+                            rc_construction_code,
+                            rc_network],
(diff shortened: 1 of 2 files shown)


albertz:
> > Already get it to a working stable state?
> 
> The PR message clearly says that this is not the aim, but just for testing.

It says "initial code" but it was not clear what this means. Runnable and stable? Totally broken? If there are no constraints other than "any code", then I don't see the purpose of having it as a PR.

> > Any future work without PRs, until we declare it as stable? Why is this PR needed at all then? Why did you not just push it directly to master?
> 
> I do not understand. There should be no changes to `common` without PRs.

Why? You also push to your user dir without PR. This is just the same here as long as we declare it as WIP-code.

> I specifically added a README.md that says that the code is not "stable" and might still be altered, but nevertheless there should be no pushes without PRs,

Why?

> as of course I (and Benedikt and you as well I gues) will actively link setups against this code then. Maybe I could have skipped the PR for the "first" code, by why should I? With a PR people better see what is happening and get notifications if enabled.

So I don't exactly know what you want know.

You want that we see the code changes? We don't need it to be a PR for that. You can look at the commits.

You want that we discuss it here and work on the PR or custom code branch? Until what point? Until it is all stable? This is unclear to me.

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+    is_feature: bool = False
+
+
+@dataclass(frozen=True)
+class DataInitArgs:
+    """
+    A helper class to store needed input args for a nn.Data object without `returnn_common` dependency
+
+    :param name: name of the data (equivalent to the extern_data entry)
+    :param available_for_inference: if this data is available during decoding/forward pass etc...
```
Atticus1806, line 87: I dont really like the wording here, but I dont like the RETURNN Data class docstring either" `available_for_inference: e.g. the extern data "classes" is usually not available for inference`. 
The RETURNN docs say `available_for_inference defines if the data stream is also available during the “search” or “forward” task, so basically marking if a data stream is an input or only a target.` 
So maybe something like `defines if the data is also available during “search” or “forward`, maybe even change on RETURNN side too. What do you think @albertz

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+    :param dim_tags: list of dim tags representing an axis of the data, without batch or a hidden sparse dim
+    :param sparse_dim: provide this dim to make the data sparse and define
```
Atticus1806, line 89: `... and define the dim`? I think this sentence is incomplete.

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+    :param has_batch_dim: set to False do create data without a batch dim (probably never needed)
```
Atticus1806, line 90:
```suggestion
    :param has_batch_dim: set to False to create data without a batch dim (probably never needed)
```

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+    def __init__(
+        self,
+        serializer_objects: List[SerializerObject],
+        returnn_common_root: Optional[tk.Path] = None,
+        packages: Optional[List[Union[str, tk.Path]]] = None,
+        make_local_package_copy=False,
+    ):
+        """
+
+        :param serializer_objects: all serializer objects which are serialized into a string  in order
```
Atticus1806, line 138:
```suggestion
        :param serializer_objects: all serializer objects which are serialized into a string in order
```

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+        :param returnn_common_root:
+        :param packages:
+        :param make_local_package_copy:
```
Atticus1806, line 141: I think it would be good if this is documented.
```suggestion
        :param returnn_common_root: Path to returnn_common, if None, assumes direct import is fine
        :param packages: Path to packages to import, if None, tries to extract them from serializer_objects
        :param make_local_package_copy: whether to make a local copy of imported code into the Job directory 
```

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+    When passed to the ReturnnCommonSerializer it will automatically detect the local package in case of
+    `make_local_package_copy=True`, unless specific package paths are given.
+    """
+
+    def __init__(
+        self,
+        code_object: str,
+    ):
+        """
+
```
Atticus1806, line 304:
```suggestion
```

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+    A class to indicate a module or function that should be imported within the returnn config
+
+    When passed to the ReturnnCommonSerializer it will automatically detect the local package in case of
+    `make_local_package_copy=True`, unless specific package paths are given.
+    """
+
+    def __init__(
+        self,
+        code_object: str,
```
Atticus1806, line 301: Is this a string and not `tk.Path` because of the manipulations done in the init? Or does this have a different reason? If so maybe we should note somewhere that this format is not the same `tk.Path.get()` returns, so there is no confusion.

common/setups/returnn_common/nnet_constructor.py
```diff
@@ -0,0 +1,425 @@
...
+     - managing the package imports from which all imports can be found
+     - optionally make a local copy of all imported code instead if importing it directly out of the recipes
+
+    """
+
+    def __init__(
+        self,
+        serializer_objects: List[SerializerObject],
+        returnn_common_root: Optional[tk.Path] = None,
+        packages: Optional[List[Union[str, tk.Path]]] = None,
```
Atticus1806, line 133: Is there a reason that it is fine for the packages to be strings but the returnn_common_root not? This seems a bit inconsistent to me.

Atticus1806 (changes requested):
Most of it looks good, a few typos etc. and also added some suggestions.

common/setups/returnn_common/nnet_constructor.py
JackTemaki: Yes it is, because the packages are typically something that can not be verified by Sisyphus, but the the returnn common dir is definitely something that exists in the FS.

common/setups/returnn_common/nnet_constructor.py
JackTemaki: Look at the example, this is an import definition and not a physical file or directory, so it can and should not be a tk.Path

common/setups/returnn_common/nnet_constructor.py
JackTemaki: lets postpone this discussion, I will just merge so that we can start working with this code.

PR #69, opened 2022-06-23
Add function to set binaries to a crp

JackTemaki:


common/setups/rasr/rasr_system.py
```diff
@@ -185,6 +185,37 @@ def _assert_corpus_name_unique(*args):
...
+        rasr_binary_path: tk.Path,
+        rasr_arch: str = "linux-x86_64-standard",
+        python_home: Optional[tk.Path] = None,
+        python_exe: Optional[tk.Path] = None,
+    ):
+        """
+        Set explicit binaries and python for RASR.
+
+        If this is not done, RasrSystem will take the defaults from the global settings
+
```
michelwi, line 202: I think with the recent change, the RasrSystem will not default back to gs.

Also this is misleading as (afaik) the default python_* will be system python if unset.

Therefore, suggest to just leave it..
```suggestion
```

michelwi (approved):

christophmluscher (approved):

PR #70, opened 2022-06-23
add LibriSpeech cart questions

JackTemaki:


christophmluscher (approved):
@JackTemaki maybe youshould be reviewing this instead of me :D since this is my code xD

common/datasets/librispeech/cart.py
```diff
@@ -0,0 +1,939 @@
...
+                            {
+                                "type": "question",
+                                "description": "voiced",
+                                "values": "B D G V Z M N NG L R W JH TH",
+                            },
+                            {
+                                "type": "question",
+                                "description": "voiceless",
+                                "values": "P T K F S SH CH HH",
+                            },
```
michelwi, line 332: could we sort these questions further above to the other consonant questions?

I don't think it'll make much of a difference, just for my sense of order

common/datasets/librispeech/cart.py
```diff
@@ -0,0 +1,939 @@
...
+                            {
+                                "type": "question",
+                                "description": "vowel_UW",
+                                "values": "UW",
+                            },
```
michelwi, line 337: I don't think single value questions will add much benefit as they are already covered by 

```
                          {
                                "type": "for-each-value",
                                "values": self.phonemes_boundary_str,
                                "questions": [
                                    {"type": "question", "description": "context-phone"}
                                ],
                            },
```

common/datasets/librispeech/cart.py
```diff
@@ -0,0 +1,939 @@
...
+                                "values": "ER",
+                            },
+                        ],
+                    },
+                ],
+            },
+        ]
+
+        if not add_unknown:
+            self.steps[1]["questions"].pop(1)
```
michelwi, line 435: Ah I see. This answers one of my comments above.

But in case anybody modifies the order of questions in the future I am afraid that we then unintentionally remove the wrong one.
I would rather that we add the question if add_unknown or at least add an assert that the name is as we expect it.

when I check "request changes" later this is what I refer to.

common/datasets/librispeech/cart.py
```diff
@@ -0,0 +1,939 @@
...
+one variant including the ARPAbet Stress-Markers, the other not.
+"""
+__all__ = [
+    "CartQuestionsWithoutStress",
+    "CartQuestionsWithStress",
+]
+
+
+class CartQuestionsWithoutStress:
+    def __init__(self, max_leaves=9001, min_obs=1000, add_unknown: bool = True):
```
michelwi, line 12: would we not have `max_leaves=12001` as default?

common/datasets/librispeech/cart.py
```diff
@@ -0,0 +1,939 @@
...
+                            {
+                                "type": "question",
+                                "description": "vowel_ER",
+                                "values": "ER0 ER1 ER2",
+                            },
+                        ],
+                    },
+                ],
+            },
+        ]
```
michelwi, line 939: here the logic to remove the unknown question is missing. request changes also for this one.

michelwi (changes requested):
I have many comments that I am open about and two as to which I request changes (see inline).

Obviously all comments for `CartQuestionsWithoutStress` apply to the other class as well.

common/datasets/librispeech/cart.py
```diff
@@ -0,0 +1,939 @@
...
+                            {
+                                "type": "question",
+                                "description": "VOWEL-SAMPA-a",
+                                "values": "AW AY",
+                            },
+                            {
+                                "type": "question",
+                                "description": "VOWEL-SAMPA-U",
+                                "values": "UH AW OW",
+                            },
```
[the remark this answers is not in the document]
christophmluscher: the carts are from a legacy setup. I am not sure if this makes sense.. since OW is also in there, it seemed fine to me..

common/datasets/librispeech/cart.py
christophmluscher: so remove all single valued?

common/datasets/librispeech/cart.py
christophmluscher: hm to make it consistent with the baseline yes..

common/datasets/librispeech/cart.py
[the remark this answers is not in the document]
christophmluscher: thinking about this and I think depending on context and speaker AW could be more a a or u sound.. so I ll leave it as is..

michelwi (approved):

PR #72, opened 2022-06-27
add FFmpeg and RASR compile helper

JackTemaki:
Some compile helper for our pipelines.

### DIFF
--- a/common/tools/audio.py
+++ b/common/tools/audio.py
@@ -1,0 +1,42 @@
+from typing import Optional
+
+from sisyphus import tk
+
+from i6_core.tools.git import CloneGitRepositoryJob
+from i6_core.tools.compile import MakeJob
+
+
+def compile_ffmpeg_binary(
+        branch: Optional[str] = None,
+        commit: Optional[str] = None) -> tk.Path:
+    """
+    Compiles FFmpeg using the official Github master.
+
+    For requirements see:
+    https://trac.ffmpeg.org/wiki/CompilationGuide/Ubuntu
+
+    The current configure options are set in a way that the resulting FFmpeg binary lacks many features.
+    This is done because we only need limited codec capabilities.
+
+    :param branch: specify a specific branch or tag, e.g. "n4.1.4"
+    :param commit: specify a specific commit
+    :return: Path object pointing to the "ffmpeg" binary executable
+    """
+    ffmpeg_repo = CloneGitRepositoryJob(
+        "https://github.com/FFmpeg/FFmpeg",
+        branch=branch,
+        commit=commit,
+    )
+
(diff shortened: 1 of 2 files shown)


common/tools/audio.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    ffmpeg_repo = CloneGitRepositoryJob(
+        "https://github.com/FFmpeg/FFmpeg",
+        branch=branch,
+        commit=commit,
+    )
+
+    ffmpeg_make_binary = MakeJob(
+        folder=ffmpeg_repo.out_repository,
+        make_sequence=None,
```
michelwi, line 33: I think `None` defaults to `["all"]` so I suggest to explicitly write that to make it more clear what make command is executed.
```suggestion
        make_sequence=["all"],
```

michelwi (approved):

common/tools/rasr.py
```diff
@@ -0,0 +1,33 @@
...
+    ).out_repository
+    make_job = MakeJob(
+        folder=rasr_repo,
+        make_sequence=["build", "install"],
+        configure_opts=["--i6"],
+        num_processes=8,
+        link_outputs={"binaries": "arch/linux-x86_64-standard/"},
+    )
+    make_job.rqmt["mem"] = 8
+    return make_job.out_links["binaries"]
```
christophmluscher, line 33: could we return the `bin_path` and the `arch`??? since we are using these also in combination?

michelwi: we could.. but this function has `arch` as input parameter, so whoever calls it already knows about it.
Otherwise as far as defaults go, both functions (this and the set binaries) have the same default, so setting both to `None` (or not setting) also works.

christophmluscher: ? I dont see `arch` as input parameter.. but that would also be fine for me.. 
but with a different rasr branch we might get a different `arch` so maybe hard coding this is not optimal?!

michelwi: > I dont see `arch` as input parameter..

Sorry, I must have been hallucinating things.
Ok, but I also think, that the current github rasr configure script can only output the default arch.

Nevertheless we could definitively have it as a returned option and/or parameter..

christophmluscher: > Ok, but I also think, that the current github rasr configure script can only output the default arch.

I thought you can checkout a branch from RASR repo and in the branch you can then have anon default `arch` param... not 100% sure though...

I think having `arch` as a param to the function would allow us the most flexibility..

christophmluscher: thanks for the explanation :)

christophmluscher (approved):

PR #76, opened 2022-07-06
feat: Improve typings in RASR-specific i6_experiments code

NeoLegends:
Hi!

In an effort to improve documentation of the RASR specific code in i6_experiments, I plan on improving the typings and to reduce the usage of `dict`s for storing configuration options. The idea is to create classes for all the options being passed to RASR that are then composed in a type-correct way. This on one hand documents the available config knobs in RASR, and on another hand makes it easier for newcomers to know which parameters are allowed in which positions.

The existing parameter creation code (e.g. in `users/luescher/setups/librispeech/pipeline_base_args.py`) and the code in the `{Gmm, Hybrid}System` classes, however, heavily relies on either producing or receiving plain python dicts. Both code bases use `.keys()`, `.items()`, etc.

This poses a problem, given that plain data classes lack these functions. I see three options:

1. Create the classes and change all the code to use them instead of dicts
2. Create the classes, but make them behave like dicts by inheriting from the `dict` class, so that the existing code continues to work
3. Create the classes and convert them to dicts before they are being passed downstream

Option 1 seems like a lot of work. Option 2 seems like the most incremental solution given that all existing code continues to work, but might be a) verbose and b) fragile. The verbosity can possibly be reduced by using a custom decorator that works similarly to `@dataclass` (or even uses that under the hood), but this again has a learning overhead. Option 3 possibly leads to the cleanest code during typing, but then loses the type information again during the conversion back to a dict.

This PR is meant to kick off the discussion on which approach is best. I've implemented option 2 and 3 for one parameter each. I have already spoken to @christophmluscher about this, and we would like to get your feedback @JackTemaki.

cc @christophmluscher @JackTemaki for reviews

### DIFF
--- a/common/setups/rasr/util/rasr.py
+++ b/common/setups/rasr/util/rasr.py
@@ -5,9 +5,9 @@ __all__ = [
     "RasrSteps",
 ]
 
-
 from collections import OrderedDict
-from typing import Dict, Optional, Union
+from dataclasses import asdict, dataclass
+from typing import Dict, Optional, Union, Tuple, List
 
 import i6_core.meta as meta
 
@@ -48,6 +48,21 @@ class CostaArgs(dict):
         self.eval_lm = eval_lm
 
 
+@dataclass
+class AmArgs:
+    state_tying: str
+    states_per_phone: int
+    state_repetitions: int
+    across_word_model: bool
+    early_recombination: bool
+    tdp_scale: float
+    tdp_transition: Tuple[Union[float, str], Union[float, str], Union[float, str], Union[float, str]]
+    tdp_silence: Tuple[Union[float, str], Union[float, str], Union[float, str], Union[float, str]]
+    tying_type: str
+    nonword_phones: Union[str, List[str]]
+    tdp_nonword: Tuple[Union[float, str], Union[float, str], Union[float, str], Union[float, str]]
+
+
 class RasrInitArgs:
     """
     Class holds general information for the complete pipeline.
@@ -62,7 +77,7 @@ class RasrInitArgs:
     def __init__(
         self,
         costa_args: Union[dict, CostaArgs],
-        am_args: dict,
+        am_args: Union[dict, AmArgs],
         feature_extraction_args: dict,
         scorer: Optional[str] = None,
         scorer_args: Optional[Dict] = None,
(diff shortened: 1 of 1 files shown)


JackTemaki:
I like the variant the most where we use dataclasses and then use `asdict()`. This offers quite the freedom, as dataclasses can also be initialized from dicts (I already use them like that in an unrelated project), so we could in the end also gradually change the whole internal code or just parts of it to work with dataclasses instead of dicts, if needed or wanted.

JackTemaki:
Do not forget to also comment in the dataclasses for which exact Job, function or so they are used.
For your AmArgs this would be:
`i6_core.am.config.acoustic_model_config()`

JackTemaki:
Also, you should consider introducing typing variables in case you have quite long types.

E.g.:
```python
TdpTuple = Tuple[Union[float, str], Union[float, str], Union[float, str], Union[float, str]]
```

christophmluscher:
I thought `namedtuples` has the `_` in `_asdict` to avoid naming conflicts?! also they are used in the documentation like a public method... so it was done on purpose but not a directly bad one??

I would also go with variant 3 and only extend to variant 1 for stuff really still WIP. If people really want to tinker, then they have to look into the code anyway and understand what efects what. For people just using the pipeline variant 3 should suffice

PS black formatting is missing ^^

michelwi:
Starting from python3.8 there is [typing.TypedDict](https://docs.python.org/3.8/library/typing.html#typing.TypedDict) which sounds almost exactly what would fit the existing code: it is a dict with a defined set of keys and typed values.

JackTemaki:
> Starting from python3.8 there is [typing.TypedDict](https://docs.python.org/3.8/library/typing.html#typing.TypedDict) which sounds almost exactly what would fit the existing code: it is a dict with a defined set of keys and typed values.

But this is then not "frozen", which is a feature which might be quite helpful...

christophmluscher:
in the code we have Dict[str, Dict] quite often. So even if the outer Dict is changed to something immutable the inner Dict is still mutable. enforcing unmutability is then quite a bit of more work or not !?

christophmluscher:
l would go for Named Dict for GMM Pipeline. least amount of changes to code. hybrid I would be more open for other suggestions

NeoLegends:
Thank you for your input, I'm going to eventually re-do this PR properly with Named Dicts.

PR #79, opened 2022-07-26
fix VTLN and store allophones

JackTemaki:
VTLN crashes when having no eval corpora, so I moved code that was not needed in the loop.

Also registering allophones which is convenient for using the archiver to view alignments.

### DIFF
--- a/common/setups/rasr/gmm_system.py
+++ b/common/setups/rasr/gmm_system.py
@@ -559,16 +559,16 @@ class GmmSystem(RasrSystem):
         :param kwargs:
         :return:
         """
+        self.vtln_features(
+            name=name,
+            corpus=train_corpus_key,
+            raw_feature_flow=self.feature_flows[train_corpus_key][raw_feature_flow_key],
+            warping_map=self.vtln_files[train_corpus_key][
+                vtln_files_key + "_warping_map"
+            ],
+            **kwargs,
+        )
         for c in eval_corpora_keys:
-            self.vtln_features(
-                name=name,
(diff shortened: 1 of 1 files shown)


common/setups/rasr/gmm_system.py
```diff
@@ -559,16 +559,16 @@ def extract_vtln_features(
...
         for c in eval_corpora_keys:
-            self.vtln_features(
-                name=name,
-                corpus=train_corpus_key,
-                raw_feature_flow=self.feature_flows[c][raw_feature_flow_key],
-                warping_map=self.vtln_files[train_corpus_key][
-                    vtln_files_key + "_warping_map"
-                ],
-                **kwargs,
-            )
```
michelwi, line 571: how did this ever make sense? it will extract the training data as vtln features again and again (ok, maybe `self.feature_flows[c][raw_feature_flow_key]` is constant for all c, so the same job is created) and overwrites the `self.feature_flows[train_corpus_key][name+"+vtln"]` every time.

michelwi (commented):
I think the change is good.

is `self.feature_flows[train_corpus_key][raw_feature_flow_key]` always the same as `self.feature_flows[c][raw_feature_flow_key]`?

common/setups/rasr/gmm_system.py
JackTemaki: I did not make sense, thus I changed it. The hashes did not change though.

michelwi (approved):

christophmluscher (approved):

PR #80, opened 2022-07-28
Nick allophone quick fix

JackTemaki:
There was a small mistake in yesterdays #79 

Not sure what should be the correct thing to do here, but I would like to merge this quick fix in so that I can continue working.

It has no implications on any hash.

### DIFF
--- a/common/setups/rasr/gmm_system.py
+++ b/common/setups/rasr/gmm_system.py
@@ -1394,8 +1394,8 @@ class GmmSystem(RasrSystem):
 
         for trn_c in self.train_corpora:
             # TODO: this will always overwrite base if we have multiple corpora
-            self.store_allophones(trn_c)
-            tk.register_output(f"{trn_c}.allophones", self.allophone_files["base"])
+            self.store_allophones(source_corpus=trn_c, target_corpus=trn_c)
+            tk.register_output(f"{trn_c}.allophones", self.allophone_files[trn_c])
 
         for eval_c in self.dev_corpora + self.test_corpora:
             stm_args = (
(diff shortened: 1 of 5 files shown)


JackTemaki:
(The first commit is just user stuff, it seems Github does not detect that this is the new main branch now)

common/setups/rasr/gmm_system.py
```diff
@@ -1393,8 +1393,9 @@ def run(self, steps: Union[List[str], RasrSteps]):
                 self.jobs[all_c]["costa"].update_rqmt("run", {"mem": 8, "time": 24})
 
         for trn_c in self.train_corpora:
+            # TODO: this will always overwrite base if we have multiple corpora
             self.store_allophones(trn_c)
```
michelwi, line 1397: I think according to https://github.com/rwth-i6/i6_core/blob/f82690c6f4cf585154100221c916b82ba324fb29/meta/system.py#L228 the proper fix would be 
```suggestion
            self.store_allophones(source_corpus=trn_c, target_corpus=trn_c)
```

JackTemaki: I was not 100% sure if this has any other implications, as then the allophones are not registered for the eval corpora if not written into "base".

michelwi: Eval corpora usually don't need an allophone file, as we don't have alignments for them. This having said.. there might be implications for SAT recognition.

But anyway we need to consider then what happens when different train corpora have different allophone files..

Another suggestion would be to assert that all allophones are the same and then omit the loop(s) altogether and register only one output.?

christophmluscher (approved):
I think this should be fine.. Nick promised me tests so we will see if it crashes ^^

PR #81, opened 2022-07-29
base and hybrid decoder class and utils [wip+draft]

christophmluscher:
still very much work in progress. but I think the general layout and structure can be discussed and used

`BaseDecoder` class
stand alone decoder class. gets the setup from `*System` class (passing full `CommonRasrParameters`) **or** by passing `RasrConfigs`. then perform decoding and gather results some how.

`HybridDecoder` class
Idea: interface to the corresponding `HybridSystem` class and set specific decoding settings like FeatureScorer for hybdird decoding.


common/setups/rasr/util/decode.py
```diff
@@ -0,0 +1,43 @@
...
+class BaseRecognitionParameters(TypedDict):
+    am_scales: List[float]
+    lm_scales: List[float]
+    pronunciation_scales: List[float]
+    transition_tdp: List[TransitionTdp]
+    silence_tdp: List[SilenceTdp]
+
+
+@dataclass()
+class PriorArgs:
```
JackTemaki, line 26: why is this a dataclass and the rest a typed dict?

common/setups/rasr/hybrid_decoder.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        super().__init__(rasr_binary_path=rasr_binary_path, rasr_arch=rasr_arch)
+
+        self.eval_corpora = []
+
+        self.eval_input_data = None
+
+    def init_system(
+        self,
+        hybrid_init_args: RasrInitArgs,
```
JackTemaki, line 29: See #82, I would call the variable also always `rasr_init_args`.

common/setups/rasr/util/decode.py
christophmluscher: For `PriorArgs` I wanted to have the option to add methods. For example to add prior computation "povey" style into the class.
@Marvin84

Marvin84: Similarly to the tdps we can also have a flexible prior computation class with following options: 

1.  Average on the train data using the model (povey)
2.  Estimate based on a Viterbi best path, i. e. alignment
3.  Estimate from transcription with specified heuristics based on average phoneme duration

common/setups/rasr/util/decode.py
```diff
@@ -0,0 +1,43 @@
...
+from ..config.am_config import TransitionTdp, SilenceTdp
+
+
+class BaseRecognitionParameters(TypedDict):
+    am_scales: List[float]
+    lm_scales: List[float]
+    pronunciation_scales: List[float]
+    transition_tdp: List[TransitionTdp]
+    silence_tdp: List[SilenceTdp]
+
```
Marvin84, line 23: am scale is not important, we always use equal to one. What matters more is tdp scale. I would substitute `am_scales` with `tdp_scales`

Marvin84: `prior_scales` are also missing

common/setups/rasr/util/decode.py
christophmluscher: Yes, in `PriorArgs` or somewhere else.. didnt think about the design yet.

common/setups/rasr/base_decoder.py
```diff
@@ -0,0 +1,146 @@
...
+            scorer_job.out_report_dir,
+        )
+
+    def _optimize_scales(
+        self, name: str, corpus_key: str, optimize_args: OptimizeJobArgs
+    ) -> (float, float):
+        opt_job = recog.OptimizeAMandLMScaleJob(**optimize_args)
+        opt_job.add_alias(f"{self.alias_output_prefix}optimize_{corpus_key}/{name}")
+
+        return opt_job.out_best_am_score, opt_job.out_best_lm_score
```
michelwi, line 113: mh.. maybe return the scales instead of the scores?^^

common/setups/rasr/base_decoder.py
```diff
@@ -0,0 +1,146 @@
...
+                corpus_key=corpus_key,
+                search_job_args={},
+                lat_2_ctm_args={},
+                scorer_args={},
+            )
+            if optimize_am_lm_scales:
+                best_am_scale, best_lm_scale = self._optimize_scales(
+                    scorer_cls=self.scorer_job_class,
+                    scorer_kwargs={},
+                    opt_only_lm_scale=True if best_am_scale == 1.0 else False,
```
michelwi, line 134: I don't understand this line.. you don't know that the best am scale is 1.0 until you did the optimization?

also all the args do not match your definition from above..

common/setups/rasr/config/am_config.py
```diff
@@ -0,0 +1,77 @@
...
+    skip: Union[float, str] = "infinity"
+    exit: Union[float, str] = 0.0
+
+
+@dataclass()
+class SilenceTdp(_Tdp):
+    loop: Union[float, str] = 0.0
+    forward: Union[float, str] = 3.0
+    skip: Union[float, str] = "infinity"
+    exit: Union[float, str] = 20.0
```
michelwi, line 40: does it make sense to define this as separate dataclasses? I would say these are just specific instances of the dataclass `_Tdp` that are by chance used as default value here (but Tina would use different values)

additionally `TransitionTdp` -> `SpeechTdp`

common/setups/rasr/config/am_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@dataclass()
+class AmRasrConfig:
+    state_tying: str = "monophone"
```
michelwi, line 45: mh.. If you really want to be fancy, could this be an instance of an `enum` object that holds all available state tying methods as `str`?

common/setups/rasr/config/am_config.py
```diff
@@ -0,0 +1,77 @@
...
+    states_per_phone: int = 3
+    state_repetitions: int = 1
+    across_word_model: bool = True
+    early_recombination: bool = False
+    tdp_scale: float = 1.0
+    tdp_transition: TransitionTdp = TransitionTdp()
+    tdp_silence: SilenceTdp = SilenceTdp()
```
michelwi, line 52:
```suggestion
    tdp_transition: _Tdp = TransitionTdp()
    tdp_silence: _Tdp = SilenceTdp()
```
if any TDP is allowed here, use base class. Then also rename base class to be public

common/setups/rasr/config/am_config.py
```diff
@@ -0,0 +1,77 @@
...
+    tying_type: str = "global"
+    nonword_phones: str = ""
+    tdp_nonword: TransitionTdp = TransitionTdp(
+        0.0, 3.0, "infinity", 6.0
+    )  # only used when tying_type = global-and-nonword
```
michelwi, line 57: > only used when tying_type = global-and-nonword

so it should be optional?

common/setups/rasr/config/am_config.py
```diff
@@ -0,0 +1,77 @@
...
+    state_tying_file: Optional[tk.Path] = None
+
+    def get(self):
```
michelwi, line 60:
```suggestion
    def get_config(self):
```
more descriptive while keeping generality towards all other `*RasrConfig` classes

common/setups/rasr/util/decode.py
```diff
@@ -0,0 +1,52 @@
...
+from ..config.am_config import TransitionTdp, SilenceTdp
+
+
+class BaseRecognitionParameters(TypedDict):
+    am_scales: List[float]
+    lm_scales: List[float]
+    pronunciation_scales: List[float]
+    tdp_scales: List[float]
+    transition_tdp: List[TransitionTdp]
+    silence_tdp: List[SilenceTdp]
```
michelwi, line 26:
```suggestion
    transition_tdps: List[TransitionTdp]
    silence_tdps: List[SilenceTdp]
```
plural to be more consistent.

Again, if you assume there could be different entries, the typing should be the base TDP class and not the derived class with hardcoded values.

common/setups/rasr/config/am_config.py
michelwi: I am wondering how data classes and imutabillity behaves. e.g.
```
a = AmRasrConfig()
a.tdp_transition.loop = 10
b = AmRasrConfig()
print(b.tdp_transition.loop)  # prints 10 
```
ok, the way it is implemented here, all AmRasrConfig share the same `TransitionTdp` instance. 
remedy: use https://docs.python.org/3/library/dataclasses.html#dataclasses.field
```suggestion
    tdp_transition: _Tdp = field(default_factory=TransitionTdp)
    tdp_silence: _Tdp = field(default_factory=SilenceTdp)
```
add `from dataclasses import field` to imports

common/setups/rasr/base_decoder.py
```diff
@@ -0,0 +1,146 @@
...
+        rasr.crp_add_default_output(
+            self.crp["base"], compress, append, unbuffered, compress_after_run
+        )
+        self.crp["base"].set_executables(rasr_binary_path, rasr_arch)
+
+        self.eval_corpora = []
+
+        self.alias_output_prefix = alias_output_prefix
+
+    def init_decoder(
```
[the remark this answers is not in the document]
christophmluscher: `__init__` settings for the class

two different kind of ways to set the params: `init_decoder` and `copy_crp_from_system`

common/setups/rasr/base_decoder.py
```diff
@@ -0,0 +1,146 @@
...
+        self,
+        *,
+        acoustic_model_config: rasr.RasrConfig,
+        lexicon_config: rasr.RasrConfig,
+        language_model_config: rasr.RasrConfig,
+        lm_lookahead_config: rasr.RasrConfig,
```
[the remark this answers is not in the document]
christophmluscher: yes I think so. suggestions?

common/setups/rasr/util/flow/tensorflow.py
```diff
@@ -0,0 +1,86 @@
...
+
+from typing import List, Optional, Union
+
+from sisyphus import tk
+
+import i6_core.rasr as rasr
+import i6_core.returnn as returnn
+
+
+def get_tf_flow(
```
JackTemaki, line 11: Do we want to have this in `i6_core` instead? This seems to be a very fundamental function.

Marvin84: You mean the tensorflow flow node configuration? IMO, this should live under `i6_experiments`

JackTemaki: Why? All other flows are in `i6_core`, so I thought this should also be there.

common/setups/rasr/util/flow/tensorflow.py
```diff
@@ -0,0 +1,86 @@
...
+    tf_fwd = tf_flow.add_node("tensorflow-forward", "tf-fwd", {"id": "$(id)"})
+    tf_flow.link(f"network:{tf_fwd_input_name}", tf_fwd + ":input")
+    tf_flow.link(tf_fwd + ":log-posteriors", "network:features")
+
+    tf_flow.config = rasr.RasrConfig()
+
+    tf_flow.config[tf_fwd].input_map.info_0.param_name = "input"
+    tf_flow.config[
+        tf_fwd
+    ].input_map.info_0.tensor_name = "extern_data/placeholders/data/data"
```
albertz, line 35: Is it good to have such hardcoded assumption on the input data name?

JackTemaki: No, see https://github.com/rwth-i6/i6_core/pull/307/

PR #83, opened 2022-07-29
stress markers incorrectly labeled

christophmluscher:


JackTemaki (approved):
I was really scared at first... good that this is only the export. Still a problem though for everyone accessing direct files.

Marvin84 (approved):

christophmluscher:
Yeah that got me worried as well

michelwi (approved):
>  Still a problem though for everyone accessing direct files.

people will notice in recognition (assuming the lexicon is correctly labeled)

PR #84, opened 2022-07-29
LibriSpeech 960h GMM Pipeline

christophmluscher:


users/luescher/setups/librispeech/pipeline_base_args.py
```diff
@@ -322,25 +336,25 @@ def get_triphone_args(
...
         "lm_lookahead": True,
         "lookahead_options": None,
         "create_lattice": True,
         "eval_single_best": True,
         "eval_best_in_lattice": True,
         "search_parameters": {
-            "beam_pruning": 12.0,
-            "beam-pruning-limit": 100000,
+            "beam_pruning": 15.0,
+            "beam-pruning-limit": 50000,
```
JackTemaki, line 348: I will do a full review later, but I can already say that 50k here is not a good idea. With pruning 15, 100k is already a very strict limit. At least this is what I interpret from:
```
zless /work/common/asr/librispeech/data/sisyphus_export_setup/alias/baselines/librispeech/ls100/gmm/common_baseline/recognition/train-clean-100-vtln+sat/recog_train-clean-100-vtln+sat-dev-other-ps1.00-lm25.00-iter10-optlm/output/search.log.*.gz  | grep "states after pruning" -A 3 | grep max | grep -o "[0-9]*" | sort -h
```

users/luescher/setups/librispeech/pipeline_base_args.py
```diff
@@ -322,25 +336,25 @@ def get_triphone_args(
...
             "word-end-pruning": 0.5,
-            "word-end-pruning-limit": 15000,
+            "word-end-pruning-limit": 10000,
```
JackTemaki, line 350: 10k here is a good idea though, for a dev-other recognition I had about 20-30 sequences above 10k and the rest below.

common/datasets/librispeech/export.py
```diff
@@ -126,7 +137,11 @@ def _export_legacy_bpe(output_prefix):
     )
 
 
-def export_all(output_prefix):
+def export_all(
+    output_prefix: str,
+    returnn_python_exe: Optional[tk.Path] = None,
```
JackTemaki, line 142: This does not make sense to be configurable, the `export_all` (and other export functions) are explicitly what the `asruser` repo should export. If you say we should have a "second" version of exporting .ogg zip files without the settings entry we can do that, but then this would be explicit here and not a configurable parameter.

common/setups/rasr/util/gmm.py
```diff
@@ -241,11 +244,11 @@ class PrevCtm:
     Defines a previous ctm output from a recognition which is used to extract the
     speaker alignments for SAT adaptation for the dev/train sets.
 
-    :param prev_step_key: name of the step, e.g. "tri"
-    :param pronunciation_scale: one of the existing pronunciation scales
-    :param lm_scale: one of the existing lm-scales
-    :param iteration: one of the existing GMM iterations
-    :param optimized_lm: use automatically optimized lm.scale which might differ from param lm_scale
+    prev_step_key: name of the step, e.g. "tri"
```
JackTemaki, line 247: This was correct before, `dataclass` allows to set `:param` in the docstring.

users/luescher/setups/librispeech/pipeline_base_args.py
```diff
@@ -653,6 +666,8 @@ def get_data_inputs(
     train_corpus="train-other-960",
     add_unknown_phoneme_and_mapping: bool = False,
     use_eval_data_subset: bool = False,
+    g2p_python: Optional[Union[str, tk.Path]] = None,
```
JackTemaki, line 669: We should not allow `str` for new parameters that are supposed to be `Path`.

users/luescher/setups/librispeech/pipeline_base_args.py
```diff
@@ -675,10 +690,18 @@ def get_data_inputs(
         "normalize_pronunciation": False,
     }
 
+    g2p_train_apply_args = {}
```
JackTemaki, line 693: I do not understand why passing the g2p args is needed, because either:
 -  someone is running the pipeline at i6, in this case all g2p jobs already exist and can be imported
 - someone outside is running the pipeline, but then it is no issue to have g2p installed in the Sisyphus env, just as the `requirements.txt` specifies.

users/luescher/setups/librispeech/pipeline_base_args.py
```diff
@@ -41,9 +44,22 @@ def get_init_args(
     mfcc_extra_args: Optional[Dict] = None,
     gt_normalization: bool = True,
     gt_options_extra_args: Optional[Dict] = None,
+    tying_type: str = "global",
+    nonword_phones: str = "",
     tdp_transition: Tuple[
         Union[float, str], Union[float, str], Union[float, str], Union[float, str]
-    ] = (3.0, 0.0, 30.0, 0.0),
+    ] = (3.0, 0.0, "infinity", 0.0),
```
JackTemaki, line 51: Having everything configurable as parameter is very different to my approach, that the baseline only gives a fixed setting and you should copy and alter the whole file.

maybe we should discuss this to have it consistent across the different baselines.

JackTemaki (changes requested):

common/datasets/librispeech/export.py
christophmluscher: so we should always symlink from the common work dir for `export` functions?!

common/setups/rasr/util/gmm.py
christophmluscher: `:param prev_step_key:`

this gives me an `Unresolved reference` error in PyCharm

removing `:` is enough to fix this, removing `"param` is not necessary. Or is this incorrect from PyCharm?

christophmluscher: seems to be a problem of PyCharm: https://youtrack.jetbrains.com/issue/PY-28549

only a quick google search so might also be something else

users/luescher/setups/librispeech/pipeline_base_args.py
christophmluscher: > users/luescher

I think we can fix as soon as we move to `common` dir :)

users/luescher/setups/librispeech/pipeline_base_args.py
christophmluscher: at the moment you cannot pass arguemtns to the g2p at all.. which probably is also not necessary.. so maybe this is a bit over-kill ^^
also remove `g2p_&_args` from `get_g2p_augmented_bliss_lexicon_dict`?

users/luescher/setups/librispeech/pipeline_base_args.py
```diff
@@ -204,34 +219,29 @@ def get_monophone_args(
...
         "create_lattice": True,
         "eval_single_best": True,
         "eval_best_in_lattice": True,
         "search_parameters": {
-            "beam-pruning": 18.0,
-            "beam-pruning-limit": 100000,
-            "word-end-pruning": 0.75,
-            "word-end-pruning-limit": 15000,
+            "beam-pruning": 15.0,
+            "beam-pruning-limit": 50000,
```
JackTemaki, line 230: here it is still 50k not 100k

users/luescher/setups/librispeech/pipeline_base_args.py
JackTemaki: I would suggest yes, we should avoid that people use different G2P.

common/datasets/librispeech/export.py
JackTemaki: Yes exactly, the `export` function define what can be symlinked.

christophmluscher:
pump to py3.8 missing -> hash fail.

either cherry-pick or merge from here: #92

JackTemaki (approved):
Approve for now, some fixes might come later.

common/setups/rasr/hybrid_system.py
```diff
@@ -482,11 +487,25 @@ def nn_recognition(
...
+
                 flow = self.add_tf_flow_to_base_flow(feature_flow, tf_flow)
                 flow.config.tf_fwd.loader.saved_model_file = checkpoints[epoch]
 
+                self.feature_scorers[recognition_corpus_key][
+                    f"pre-nn-{name}-{prior:02.2f}"
+                ] = scorer
+                self.feature_flows[recognition_corpus_key][
+                    f"{feature_flow_key}-tf-{epoch:03d}"
+                ] = flow
```
vieting, line 501: @christophmluscher is this used lateron?

Note that `self.feature_flows[recognition_corpus_key]` can also be a `rasr.FlowNetwork` already and not a dict, see a couple of lines above. I get an error due to this
```
TypeError: 'FlowNetwork' object does not support item assignment
```

PR #86, opened 2022-08-02
GMM Forced Align Fix and proposed changes to output_step

Atticus1806:
This PR includes a small change and fix to the forced align step which was broken/did not work properly before. 
After that we noticed that the output step tries to always get the alignment from the previous step, but if the last step was forced align and the alignment was on a different corpus this would crash.

@christophmluscher and I thought of a solution for this that works, but there might be a cleaner way to solve this (we could not find due to time constraints yet), so feel free to propose (bigger) changes (cf. e.g. TODO in rasr.py)
Our idea was that this way the list of steps which (in the future) have similar implications could just be added to the `get_non_gmm_steps` list.


common/setups/rasr/gmm_system.py
```diff
@@ -1447,9 +1447,10 @@ def run(self, steps: Union[List[str], RasrSteps]):
 
             # ---------- Forced Alignment ----------
             if step_name.startswith("forced_align"):
-                for trn_c in self.train_corpora:
+                corpus_keys = step_args.pop("corpus_keys", [])
```
michelwi, line 1450: should the default not stay `self.train_corpora`?
```suggestion
                corpus_keys = step_args.pop("corpus_keys", self.train_corpora)
```

common/setups/rasr/util/rasr.py
```diff
@@ -198,5 +198,17 @@ def get_step_iter(self):
...
+        return ["forced_align"]
+
+    def get_gmm_steps_names_as_list(self):
+        step_names = self.get_step_names_as_list()
+        # TODO nicer list comprehension
+        for non_gmm_step in self.get_non_gmm_steps_as_list():
+            step_names = [
+                name for name in step_names if not name.startswith(non_gmm_step)
+            ]
+        return list(self._step_names_args.keys())
```
michelwi, line 211: this is the same as in the function `get_step_names_as_list`. did you mean
```suggestion
        return step_names
```

common/setups/rasr/util/rasr.py
```diff
@@ -198,5 +198,17 @@ def get_step_iter(self):
...
+    def get_non_gmm_steps_as_list(self):
+        return ["forced_align"]
+
+    def get_gmm_steps_names_as_list(self):
+        step_names = self.get_step_names_as_list()
+        # TODO nicer list comprehension
+        for non_gmm_step in self.get_non_gmm_steps_as_list():
+            step_names = [
+                name for name in step_names if not name.startswith(non_gmm_step)
+            ]
```
michelwi, line 210: My suggestion for "TODO nicer list comprehension"
```suggestion
        step_names = filter(lambda x: not any(x.startswith(step) for step in self.get_non_gmm_steps_as_list()), self.get_step_names_as_list())
```

common/setups/rasr/util/rasr.py
```diff
@@ -198,5 +198,17 @@ def get_step_iter(self):
     def get_step_names_as_list(self):
         return list(self._step_names_args.keys())
 
+    def get_non_gmm_steps_as_list(self):
```
michelwi, line 201: why do we need this distinction into `gmm_steps` and `non_gmm_steps`? 

does it make sense to hardcode the list of non_gmm_steps to `["forced_align"]` are there no other non_gmm_steps?

common/setups/rasr/gmm_system.py
```diff
@@ -1087,19 +1087,19 @@ def get_gmm_output(
                 corpus_key
             ].get(scorer_key, [None])[-1]
 
-        if corpus_type == "dev" and self.alignments[corpus_key].get(
-            f"alignment_{steps.get_step_names_as_list()[step_idx - 1]}", False
-        ):
+        if (corpus_type == "dev" or corpus_type == "test") and self.alignments[
```
michelwi, line 1090: usually we don't have alignments for the test sets, because we "assume" that the transcription is unknown..

Atticus1806: I agree that we usually dont have that, but I still think it should be supported for cases like TTS or when you would want to do some statistics etc. where you need an alignment.

common/setups/rasr/util/rasr.py
Atticus1806: So our idea (maybe the naming is off) is that non_gmm steps are steps that do not (per default) include some gmm training where we have guaranteed alignments and mixture and thus these outputs cannot be used (since they dont exist)

Allowed steps right now are: "extract, mono, cart, tri, vtln, sat, vtln+sat, forced_align" 
I am not sure how to handle extract, because instantly calling output after extract should not be usual behavior. Maybe add this here aswell?

common/setups/rasr/gmm_system.py
Atticus1806: nvm this already is a list, somehow thought this was just a corpus, then this would have worked ofc.

Atticus1806:
so there was one last problem: Just taking `step_idx-1` of course might cause a problem because e.g. 4 steps but only 3 gmm steps so `step_idx - 1` would actually try to get the output of the output step which does not exist yet, so the number of `non_gmm_steps` had to be taken into account.
Now this should be ready and working though, tested it on my setup and the graph builds correctly.

common/setups/rasr/gmm_system.py
michelwi: ok, we can keep this

common/setups/rasr/util/rasr.py
michelwi: ok, good idea.. But I'll let you guys at the chair decide on the design here

common/setups/rasr/gmm_system.py
```diff
@@ -1087,19 +1087,22 @@ def get_gmm_output(
...
+            corpus_key
+        ].get(f"alignment_{steps.get_step_names_as_list()[step_idx - 1]}", False):
             gmm_output.alignments = self.alignments[corpus_key][
                 f"alignment_{steps.get_step_names_as_list()[step_idx - 1]}"
             ][-1]
 
         if corpus_type == "train":
+            gmm_step_list = steps.get_gmm_steps_names_as_list()
+            step_list = steps.get_step_names_as_list()
+            dist_last_step = 1 + (len(step_list) - len(gmm_step_list))
```
michelwi, line 1100: are you sure this logic works in any case, even if there are non_gmm_steps interleaved with gmm_steps?

e.g. `steps.get_step_names_as_list() == ["extract", "mono", "force_align", "tri", "force_align"]`

Atticus1806: You are totally right, this would not have worked. I changed the logic a bit and introduced `get_prev_gmm_step` which (at least with the test you proposed) worked. 
In general this makes `get_gmm_steps_names_as_list` kind of redundant and unused, but personally i would not mind to keep this function for potential future usage (since its already written) but this is by far no strong opinion so if you think this clutters the code I would just remove it.

Atticus1806:
So is there anything else, that needs to be changed? Or can this be merged now? @JackTemaki @michelwi @christophmluscher @Marvin84

michelwi:
I am fine with the changes, but I am not a user of this code so I'll abstain.

JackTemaki (approved):
Now that there is a hash-test for the GMM baseline I think we can approve this for now as this does not affect any existing systems. I will then test if there are any other issues, which I do not think there are.

JackTemaki:
@christophmluscher Ping :)

christophmluscher (approved):

PR #87, opened 2022-08-02
Add SCTK compilation to GMM pipeline

JackTemaki:


common/tools/sctk.py
```diff
@@ -0,0 +1,30 @@
...
+    sctk_repo = CloneGitRepositoryJob(
+        url=sctk_git_repository, branch=branch, commit=commit
+    ).out_repository
+    sctk_make = MakeJob(
+        folder=sctk_repo,
+        make_sequence=["config", "all", "check", "install", "doc"],
+        link_outputs={"bin": "bin/"},
+    )
+    # This is needed for the compilation to work in the i6 environment, otherwise still untested
+    sctk_make._sis_environment.set("CPPFLAGS", "-std=c++11")
```
michelwi, line 29: mh.. I didn't need this, but then I was compiling on a different system possibly using a different compiler version..

michelwi (approved):

common/tools/sctk.py
christophmluscher: I dont think I needed this either.. but we might to update this anyway for OS switch

christophmluscher (approved):

Marvin84 (approved):

PR #91, opened 2022-08-26
cut and stitch job: modify parallelization

vieting:
The `CutAndStitchSpeechSegmentsFromCorpusJob` currently does parallelization via `multiprocessing.Pool`. This is limited by the number of workers available on a single node. This PR modifies the job to parallelize via splitting the list of recordings and processing a subset of the recordings in each job split. This is also more in line with how we parallelize in other jobs.

Adding @dierkes-j for visibility since you wrote the job in the first place, but I cannot add you as a reviewer. Please feel free to comment.

### DIFF
--- a/users/dierkes/preprocessing/segment_cutting.py
+++ b/users/dierkes/preprocessing/segment_cutting.py
@@ -54,7 +54,11 @@ class CutAndStitchSpeechSegmentsFromCorpusJob(Job):
         self.length_dist_rqmt = None  # noqa, for backwards compatibility
 
     def tasks(self):
-        yield Task("cut_and_stitch_segments", rqmt=self.cut_and_stitch_segments_rqmt, args=range(1, self.n_workers + 1))
+        yield Task(
+            "cut_and_stitch_segments",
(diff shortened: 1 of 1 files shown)


dierkes-j:
We should add `ogg` to `assert file_extension in ["wav", "flac", "mp3"]`  and adjust this in the doc string of the job.

vieting:
> We should add `ogg` to `assert file_extension in ["wav", "flac", "mp3"]` and adjust this in the doc string of the job.

That is a different topic. This PR is about the parallelization. Adding ogg can be a helpful extension, but it would be better to open a separate PR for that imho.

dierkes-j:
> > We should add `ogg` to `assert file_extension in ["wav", "flac", "mp3"]` and adjust this in the doc string of the job.
> 
> That is a different topic. This PR is about the parallelization. Adding ogg can be a helpful extension, but it would be better to open a separate PR for that imho.

Yes, correct. I will do a PR for this adjustment.

dierkes-j:
I checked the PR and it looks good to me.

PR #92, opened 2022-08-30
feat: Add typings based on Typed Dicts for the hybrid pipeline

NeoLegends:
This PR improves the documentation for the hybrid pipeline by introducing types for a subset of the parameters. To avoid heavy code changes I used Typed Dicts. The declared classes continue working like plain old regular dicts and inherit all their methods. This is important because the code in the `HybridSystem` makes use of 1) dict-specific methods like `.get()` and 2) uses `**` to forward kwargs, which doesn't work with non-dicts.

cc @christophmluscher @JackTemaki

### DIFF
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -248,7 +248,7 @@ class NnRecogArgs(TypedDict):
     use_gpu: Optional[bool]
 
 
-KeyedRecogArgs = Dict[str, Union[Dict[str, Any], NnRecogArgs]]
+KeyedRecogArgsType = Dict[str, Union[Dict[str, Any], NnRecogArgs]]
 
 
 class EpochPartitioning(TypedDict):
@@ -282,7 +282,7 @@ class HybridArgs:
         returnn_training_configs: Dict[str, returnn.ReturnnConfig],
         returnn_recognition_configs: Dict[str, returnn.ReturnnConfig],
         training_args: Union[Dict[str, Any], NnTrainingArgs],
-        recognition_args: KeyedRecogArgs,
+        recognition_args: KeyedRecogArgsType,
         test_recognition_args: Optional[KeyedRecogArgs] = None,
     ):
         """
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -2,7 +2,6 @@ __all__ = ["ReturnnRasrDataInput", "OggZipHdfDataInput", "HybridArgs", "NnRecogA
 
 from typing import Any, Dict, List, Optional, Type, Union, TypedDict
 
-from i6_core.rasr import RasrConfig
 from sisyphus import tk
 
 import i6_core.am as am
@@ -260,8 +259,8 @@ class NnTrainingArgs(TypedDict):
     cpu_rqmt: Optional[int]
     device: Optional[str]
     disregarded_classes: Optional[Any]
-    extra_rasr_config: Optional[RasrConfig]
-    extra_rasr_post_config: Optional[RasrConfig]
+    extra_rasr_config: Optional[rasr.RasrConfig]
+    extra_rasr_post_config: Optional[rasr.RasrConfig]
     horovod_num_processes: Optional[int]
     keep_epochs: Optional[bool]
     log_verbosity: Optional[int]
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -248,6 +248,9 @@ class NnRecogArgs(TypedDict):
     use_gpu: Optional[bool]
 
 
+KeyedRecogArgs = Dict[str, Union[Dict[str, Any], NnRecogArgs]]
+
+
 class EpochPartitioning(TypedDict):
     dev: int
     train: int
@@ -278,9 +281,9 @@ class HybridArgs:
         self,
         returnn_training_configs: Dict[str, returnn.ReturnnConfig],
         returnn_recognition_configs: Dict[str, returnn.ReturnnConfig],
-        training_args: NnTrainingArgs,
-        recognition_args: Dict[str, NnRecogArgs],
-        test_recognition_args: Optional[Dict[str, NnRecogArgs]] = None,
+        training_args: Union[Dict[str, Any], NnTrainingArgs],
+        recognition_args: KeyedRecogArgs,
+        test_recognition_args: Optional[KeyedRecogArgs] = None,
     ):
         """
         ##################################################
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -278,7 +278,7 @@ class HybridArgs:
         self,
         returnn_training_configs: Dict[str, returnn.ReturnnConfig],
         returnn_recognition_configs: Dict[str, returnn.ReturnnConfig],
-        training_args: Dict[str, NnTrainingArgs],
+        training_args: NnTrainingArgs,
         recognition_args: Dict[str, NnRecogArgs],
         test_recognition_args: Optional[Dict[str, NnRecogArgs]] = None,
     ):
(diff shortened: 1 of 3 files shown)


common/setups/rasr/util/nn.py
```diff
@@ -1,19 +1,16 @@
 __all__ = ["ReturnnRasrDataInput", "OggZipHdfDataInput", "HybridArgs", "NnRecogArgs"]
 
-from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Type, Union
+from typing import Any, Dict, List, Optional, Type, Union, TypedDict
 
+from i6_core.rasr import RasrConfig
```
christophmluscher, line 5: This is in the wrong place.

christophmluscher: Please move to other i6_core imports

common/setups/rasr/util/nn.py
```diff
@@ -198,52 +195,113 @@ def get_data_dict(self):
...
+
 class HybridArgs:
     def __init__(
         self,
         returnn_training_configs: Dict[str, returnn.ReturnnConfig],
         returnn_recognition_configs: Dict[str, returnn.ReturnnConfig],
-        training_args: Dict[str, Any],
-        recognition_args: Dict[str, Dict[str, Dict]],
-        test_recognition_args: Optional[Dict[str, Dict[str, Dict]]] = None,
+        training_args: Dict[str, NnTrainingArgs],
```
christophmluscher, line 282: This still allows a dict, correct?

christophmluscher:
The hash test fails.. therefore changes are required!

christophmluscher (changes requested):
Hash tests need to pass

common/setups/rasr/util/nn.py
NeoLegends: Type inference might complain (which is what we want, I think), but at runtime, typed dicts are equivalent to regular dicts.

NeoLegends:
> Hash tests need to pass

Looking into the hash test I found that it uses python 3.7, while typed dicts have only been introduced starting 3.8. Is it okay to upgrade the hash test to 3.8?

common/setups/rasr/util/nn.py
JackTemaki: This typing is wrong, because it says you want a dict of multiple keyed NnTrainingArgs. I assume you want: `Union[Dict[str, Any], NnTrainingArgs]` so either a str->any dict or the correct typed dict.

JackTemaki:
> > Hash tests need to pass
> 
> Looking into the hash test I found that it uses python 3.7, while typed dicts have only been introduced starting 3.8, and that this is why it fails. Is it okay to upgrade the hash test to 3.8?

From my side it is okay, 3.8 is anyway our default Python version.

common/setups/rasr/util/nn.py
NeoLegends: Oh, good catch. Thanks. I wonder if we want to drop the `Dict[str, Any]` part altogether and have type inference complain or if we want to be more laissez-faire, and optionally support the types there, but also not complain if one doesn't use the typed dicts.

NeoLegends:
> From my side it is okay, 3.8 is anyway our default Python version.

I'll update the test to see if it would pass tests in that case. 👍

Atticus1806:
> From my side it is okay, 3.8 is anyway our default Python version.

Is that the case? returnn_common supports >=3.7 and i6_core tests also uses 3.7.
Change is fine by me, but we should make sure this stays consistent I think.

michelwi:
> i6_core tests also uses 3.7. 
Change is fine by me, but we should make sure this stays consistent I think.

AppTek uses i6_core heavily and still uses python3.6. Until they have upgraded all setups, at least i6_core should still support this version.
For i6_experimens nothing speaks against updating the python requirement from my side.

NeoLegends:
Hm, requesting a re-review from @christophmluscher automatically removed @Marvin84 and @michelwi. Your reviews would still be appreciated!

common/setups/rasr/util/nn.py
```diff
@@ -198,52 +194,116 @@ def get_data_dict(self):
...
+class LatticeToCtmArgs(TypedDict):
+    best_path_algo: Optional[str]
+    encoding: Optional[str]
+    extra_config: Optional[Any]
+    extra_post_config: Optional[Any]
+    fill_empty_segments: Optional[bool]
+
+
+class NnRecogArgs(TypedDict):
+    acoustic_mixture_path: Optional[tk.Path]
```
christophmluscher, line 228: I noticed an inconsistency...
`acoustic_mixture_path` and `acoustic_mixtures` for the same kind of object... I think I introduced that inconsistency...
@michelwi @vieting @JackTemaki @Marvin84 preferences?

christophmluscher: Address this in a new PR or here?

christophmluscher: @JackTemaki your opinion?

common/setups/rasr/util/nn.py
```diff
@@ -198,52 +194,116 @@ def get_data_dict(self):
...
 class HybridArgs:
     def __init__(
         self,
         returnn_training_configs: Dict[str, returnn.ReturnnConfig],
         returnn_recognition_configs: Dict[str, returnn.ReturnnConfig],
-        training_args: Dict[str, Any],
-        recognition_args: Dict[str, Dict[str, Dict]],
-        test_recognition_args: Optional[Dict[str, Dict[str, Dict]]] = None,
+        training_args: Union[Dict[str, Any], NnTrainingArgs],
+        recognition_args: KeyedRecogArgs,
```
christophmluscher, line 285: just wondering.. allow training with out recog?? maybe not in this PR.. but in general @JackTemaki @Marvin84 @michelwi

michelwi: yea.. should be acceptable. Maybe I just want to start a bunch of trainings and care about recog later.

common/setups/rasr/util/nn.py
```diff
@@ -198,52 +194,116 @@ def get_data_dict(self):
         }
 
 
+# Attribute names are invalid identifiers, therefore use old syntax
+SearchParameters = TypedDict(
+    "SearchParameters",
+    {
+        "beam-pruning": float,
```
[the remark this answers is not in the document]
michelwi: carefull, RASR does no such conversion. the crp object acts like a (nested) dict, but adds all keys as members. Therefore as the member cannot have a minus sign, all "-" are converted to "_".
If you use dict access, however, the minus should be used. I.e. `crp["beam-pruning"]` and `crp.beam_pruning` are equivalent.

so only if this string is later used with `setattr` it should be changed.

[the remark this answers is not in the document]
christophmluscher: yes, you are right. I just checked the AdvTreeJob, it needs "-" for the dict access

```
config.flf_lattice_tool.network.recognizer.recognizer.beam_pruning = (
            search_parameters["beam-pruning"]
        )
```

common/setups/rasr/util/nn.py
christophmluscher: I would still address this in a separate PR, since this might need a change in the pipeline

JackTemaki, line 285:
```suggestion
        recognition_args: KeyedRecogArgsType,
```

common/setups/rasr/util/nn.py
JackTemaki: New issue.

common/setups/rasr/util/nn.py
```diff
@@ -198,52 +194,116 @@ def get_data_dict(self):
...
     def __init__(
         self,
         returnn_training_configs: Dict[str, returnn.ReturnnConfig],
         returnn_recognition_configs: Dict[str, returnn.ReturnnConfig],
-        training_args: Dict[str, Any],
-        recognition_args: Dict[str, Dict[str, Dict]],
-        test_recognition_args: Optional[Dict[str, Dict[str, Dict]]] = None,
+        training_args: Union[Dict[str, Any], NnTrainingArgs],
+        recognition_args: KeyedRecogArgsType,
+        test_recognition_args: Optional[KeyedRecogArgs] = None,
```
JackTemaki, line 286:
```suggestion
        test_recognition_args: Optional[KeyedRecogArgsType] = None,
```

NeoLegends:
🎉

PR #94, opened 2022-09-08
extend OggZipHdfDataInput

vieting:
This PR extends the `OggZipHdfDataInput` in some ways:
- `oggzip_files` is now a list of files as its plural formulation already suggested. It is also passed to RETURNN as a list instead of a tuple which is consistent with the docstring there. Note that this changes the hashes, but I think no one is using it so far, at least according to @christophmluscher.
- Make `targets` optional as I think in many cases it would not be used.
- Allow setting more options for the `OggZipDataset` using `ogg_args`. We might also add `hdf_args` if needed at some point.
- Add `acoustic_mixtures`. This is required by the recognition in the `HybridSystem`. I'm not 100% sure whether it's the best solution to add this to the `OggZipHdfDataInput`, but at least it would be consistent with the way it is done for the `ReturnnRasrDataInput`.

### DIFF
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -148,13 +148,15 @@ class ReturnnRasrDataInput:
 class OggZipHdfDataInput:
     def __init__(
         self,
-        oggzip_files: tk.Path,
+        oggzip_files: List[tk.Path],
         alignments: tk.Path,
         context_window: Dict,
         audio: Dict,
-        targets: str,
+        targets: Optional[str] = None,
         partition_epoch: int = 1,
         seq_ordering: str = "laplace:.1000",
+        ogg_args: Optional[Dict[str, Any]] = None,
+        acoustic_mixtures: Optional[Union[tk.Path, str]] = None,
     ):
         """
         :param oggzip_files:
@@ -172,6 +174,8 @@ class OggZipHdfDataInput:
         self.partition_epoch = partition_epoch
         self.seq_ordering = seq_ordering
         self.targets = targets
+        self.ogg_args = ogg_args
+        self.acoustic_mixtures = acoustic_mixtures
 
     def get_data_dict(self):
         return {
@@ -188,10 +192,13 @@ class OggZipHdfDataInput:
                     "class": "OggZipDataset",
                     "audio": self.audio,
                     "partition_epoch": self.partition_epoch,
-                    "path": tuple(self.oggzip_files.get_path()),
+                    "path": [
+                        oggzip_file.get_path() for oggzip_file in self.oggzip_files
+                    ],
                     "seq_ordering": self.seq_ordering,
                     "targets": self.targets,
                     "use_cache_manager": True,
+                    **(self.ogg_args or {}),
                 },
             },
             "seq_order_control_dataset": "ogg",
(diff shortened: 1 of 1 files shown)


common/setups/rasr/util/nn.py
```diff
@@ -148,13 +148,15 @@ def get_crp(self, **kwargs):
 class OggZipHdfDataInput:
     def __init__(
         self,
-        oggzip_files: tk.Path,
+        oggzip_files: List[tk.Path],
         alignments: tk.Path,
```
christophmluscher, line 152: I was just wondering if this should also be Optional?

common/setups/rasr/util/nn.py
```diff
@@ -148,13 +148,15 @@ def get_crp(self, **kwargs):
...
         context_window: Dict,
         audio: Dict,
-        targets: str,
+        targets: Optional[str] = None,
         partition_epoch: int = 1,
         seq_ordering: str = "laplace:.1000",
+        ogg_args: Optional[Dict[str, Any]] = None,
+        acoustic_mixtures: Optional[Union[tk.Path, str]] = None,
```
christophmluscher, line 159: I think we should disallow `str` and only use `tk.Path`.

common/setups/rasr/util/nn.py
```diff
@@ -188,10 +192,13 @@ def get_data_dict(self):
                     "class": "OggZipDataset",
                     "audio": self.audio,
                     "partition_epoch": self.partition_epoch,
-                    "path": tuple(self.oggzip_files.get_path()),
+                    "path": [
+                        oggzip_file.get_path() for oggzip_file in self.oggzip_files
```
christophmluscher, line 196: I see the potential that this would hard-code hashes?!
Can we avoid that?

michelwi: uhm.. not using `get_path()` here but leave in the `tk.Path` object s.t. it will be converted by the write config method?

michelwi: only one way to find out xP

christophmluscher: In the `ReturnnConfig` https://github.com/rwth-i6/i6_core/blob/e2cce78e2be5a9ad3f04cbb48d743b533aaf628b/returnn/config.py#L254

```
pp = pprint.PrettyPrinter(indent=2, width=150, **self.pprint_kwargs)
for k, v in sorted(config.items()) if self.sort_config else config.items():
    if pprint.isreadable(v):
        config_lines.append("%s = %s" % (k, pp.pformat(v)))
    else:
        unreadable_data[k] = v
```

`v` should be the OggZip dict object. The dict should get parsed as a string via `pp.pformat`. The default depth is None. Which should convert all nested objects as well. And the Sisyphus Path Object `__str__` method returns the cached path as a `str`. So should work... I think :D

christophmluscher: @vieting can you confirm? xP

common/setups/rasr/util/nn.py
vieting: `alignments` you mean, right? If we make this optional, the question is why it should still be the oggzipHDFdatainput class as it's only using ogg zip. We should also not put the `HDFDataset` into the data dict in this case becase we don't have any argument for `files`, and then the `data_map` would also need to be reduced. I think it would be easier to have a `OggZipDataInput` for that case.

common/setups/rasr/util/nn.py
vieting: Yes, that works 👍

common/setups/rasr/util/nn.py
christophmluscher: Mabye then something like: `OggZipHdfDataInput(OggZipDataInput)` ^^

christophmluscher (approved):

PR #97, opened 2022-09-20
LibriSpeech Data Fixes

christophmluscher:
* training dataset does not need a LM. since the lexicon does not match, I think we should the disable this
* a wrong key is passed to corpus creation


JackTemaki (approved):

NeoLegends (approved):
lgtm

michelwi:
> training dataset does not need a LM. since the lexicon does not match, I think we should the disable this

For sequence training we do need a LM for train. But then the lexicon does have to match... Until then we can disable it.

PR #101, opened 2022-11-02
Hybrid updates for TF-flow creation and returnn_common support

JackTemaki:
- use tf-flow functions from `i6_core`
 - option to compile the graph epoch specific (e.g. for dynamic networks with returnn_common)


common/setups/rasr/hybrid_system.py
```diff
@@ -561,6 +480,29 @@ def nn_recog(
...
+        :param epoch:
+        :return: the TF graph
+        """
+        graph_compile_job = returnn.CompileTFGraphJob(
+            returnn_config,
+            epoch=epoch,
+            returnn_root=self.returnn_root,
+            returnn_python_exe=self.returnn_python_exe,
+        )
+        graph_compile_job.add_alias(f"nn_recog/graph/{name}.meta")
```
michelwi, line 503: the alias is for a folder.. so maybe not let the name end on ".meta" because this might suggest this is the compiled meta file.

michelwi (commented):
I approve of the changes, but I'll let the guys actually using it press "approve" :)

common/setups/rasr/hybrid_system.py
JackTemaki: Ah this was already in the old code, but yes this is no excuse, will fix.

JackTemaki:
> Not sure if this is still WIP or pending changes, some docstring for the functions you changed/added would be nice.

The whole Hybrid setup is WIP, but yes I can (and should) add the docstrings already

JackTemaki:
@vieting @Marvin84 Can you by any chance review this PR? This is unfortunately blocking my Switchboard PR, as otherwise I will create an even larger mess in my current experiment setup.

common/setups/rasr/hybrid_system.py
```diff
@@ -437,6 +372,8 @@ def nn_recognition(
         rtf: int,
         mem: int,
         epochs: Optional[List[int]] = None,
+        use_epoch_for_compile=False,
+        forward_output_layer="output",
```
vieting, line 376: `forward_output_layer` is not automatically inferred from the `returnn_config` anymore. Are there cases where it does not match? Besides, the default is changed from `"log_output"` to `"output"`, is that desired?

vieting (approved):
Looks good to me, just a small question regarding the `forward_output_layer`.

common/setups/rasr/hybrid_system.py
JackTemaki: Log output was something I changed at some point, the original was `output`, same as with the dependency on the `returnn_config`. Now it is back like it was before.

christophmluscher (approved):

PR #102, opened 2022-11-02
Docs and step marking support

Atticus1806:
Adds documentation and the support for https://github.com/rwth-i6/i6_core/pull/334.

### DIFF
--- a/common/setups/rasr/gmm_system.py
+++ b/common/setups/rasr/gmm_system.py
@@ -1478,13 +1478,13 @@ class GmmSystem(RasrSystem):
             self.create_stm_from_corpus(eval_c, **stm_args)
             self._set_scorer_for_corpus(eval_c)
 
-            # ppl_job = lm.ComputePerplexityJob(
-            #    self.crp[eval_c],
-            #    corpus_recipes.CorpusToTxtJob(
-            #        self.crp[eval_c].corpus_config.file
-            #    ).out_txt,
-            # )
-            # tk.register_output(f"lms/{eval_c}.ppl", ppl_job.perplexity)
+            ppl_job = lm.ComputePerplexityJob(
+                self.crp[eval_c],
+                corpus_recipes.CorpusToTxtJob(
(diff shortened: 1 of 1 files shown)


Atticus1806:
For `feature_energy_flow_key` and `feature_flow` I am not sure what to put as doc, suggest things if you like, will add them.

JackTemaki (approved):

Atticus1806:
ping @michelwi :)

common/setups/rasr/gmm_system.py
```diff
@@ -245,12 +245,14 @@ def monophone_training(
             )
 
         action_sequence = meta.align_and_accumulate_sequence(
-            align_iter, 1, mark_accumulate=False, mark_align=False
+            align_iter,
+            1,
+            mark_accumulate=kwargs.pop("mark_accumulate", False),
+            mark_align=kwargs.pop("mark_align", False),
```
michelwi, line 251: I prefer to have this explicitly as input arguments to the job and not hide it in the kwargs.

common/setups/rasr/gmm_system.py
```diff
@@ -245,12 +245,14 @@ def monophone_training(
...
         )
         action_sequence += meta.split_and_accumulate_sequence(
             splits, accs_per_split
         ) + ["align!"]
```
michelwi, line 255: why is nothing marked here?

It seems to me that your choice of marking is somewhat specific to your setup and could profit from generalizability.

michelwi (commented):
I have given some critique inline, but as I am not using i6_experiments I would abstain from approving/declining and leave this to the users.

common/setups/rasr/gmm_system.py
Atticus1806: Inside the function the accums are marked, the splits are not. This from the i6 baseline from what I know and since that was not part of what I need at the moment, I did not really touch it. But I agree at some point it would be good to also add it here.

PR #109, opened 2022-12-06
add ls960 GMM baseline

JackTemaki:
This is the GMM setup for ls960, mostly copied from Chris with some small simplifying changes.

Please check if there is anything that seems not correct to you in terms of parameters. In terms of formatting and documentation we can (and probably will) make adjustments later on. So the most important thing here is that I can run this pipeline "officially" and make it available to everyone.

@Marvin84 I included you for the monophone step.

### DIFF
--- a/common/baselines/librispeech/ls960/gmm/baseline_args.py
+++ b/common/baselines/librispeech/ls960/gmm/baseline_args.py
@@ -51,7 +51,10 @@ def get_init_args() -> util.RasrInitArgs:
                 "warping_function": "mel",
                 # to be compatible with our old magic number, we have to use 20 features
                 "filter_width": filter_width_from_channels(
-                    channels=20, warping_function="mel", f_max=8000
+                    channels=20,
+                    warping_function="mel",
+                    f_max=8000,
+                    f_min=0,
                 ),
                 "normalize": True,
                 "normalization_options": None,
--- a/common/baselines/librispeech/ls960/gmm/baseline_args.py
+++ b/common/baselines/librispeech/ls960/gmm/baseline_args.py
@@ -37,7 +37,7 @@ def get_init_args() -> util.RasrInitArgs:
             0.0,
             3.0,
             "infinity",
-            6.0,
+            21.0,
         ),  # only used when tying_type = global-and-nonword
     }
 
@@ -150,6 +150,7 @@ def get_cart_args(
     max_leaves: int = 12001,
     min_obs: int = 1000,
     hmm_states: int = 3,
+    n_phones: int = 3,
     feature_flow: str = "mfcc+deriv+norm",
     add_unknown: bool = False,
 ) -> util.GmmCartArgs:
@@ -159,6 +160,7 @@ def get_cart_args(
     :param max_leaves: total number of final CART states
     :param min_obs: minimum observation count needed for a label
     :param hmm_states: number of states per label for the HMM-topology
+    :param n_phones: phoneme context per label, use 3 for triphone and 2 for diphone context
     :param feature_flow:
     :param add_unknown: set to true if an unknown phoneme exists
     """
@@ -171,6 +173,7 @@ def get_cart_args(
         max_leaves=max_leaves,
         min_obs=min_obs,
         add_unknown=add_unknown,
+        n_phones=n_phones,
     )
 
     cart_questions = cart.PythonCartQuestions(
--- a/common/baselines/librispeech/ls960/gmm/baseline_args.py
+++ b/common/baselines/librispeech/ls960/gmm/baseline_args.py
@@ -86,9 +86,7 @@ def get_init_args():
     )
 
 
-def get_monophone_args(
-        feature_flow: str = "mfcc+deriv+norm"
-) -> util.GmmMonophoneArgs:
+def get_monophone_args(feature_flow: str = "mfcc+deriv+norm") -> util.GmmMonophoneArgs:
 
     linear_alignment_args = {
         "minimum_segment_length": 0,
@@ -200,9 +198,7 @@ def get_cart_args(
     )
 
 
-def get_triphone_args(
-    feature_flow: str = "mfcc+context+lda"
-) -> util.GmmTriphoneArgs:
+def get_triphone_args(feature_flow: str = "mfcc+context+lda") -> util.GmmTriphoneArgs:
     """
     :param feature_flow:
     """
@@ -257,9 +253,7 @@ def get_triphone_args(
     )
 
 
-def get_vtln_args(
-    feature_flow: str = "mfcc+context+lda"
-) -> util.GmmVtlnArgs:
+def get_vtln_args(feature_flow: str = "mfcc+context+lda") -> util.GmmVtlnArgs:
     """
     :param feature_flow:
     """
@@ -328,13 +322,11 @@ def get_vtln_args(
     )
 
 
-def get_sat_args(
-        feature_flow: str = "mfcc+context+lda"
-) -> util.GmmSatArgs:
+def get_sat_args(feature_flow: str = "mfcc+context+lda") -> util.GmmSatArgs:
     """
     :param feature_flow:
     """
-    base_flow = feature_flow.split('+')[0]
+    base_flow = feature_flow.split("+")[0]
     sat_training_args = {
         "name": "sat",
         "mixtures": "estimate_mixtures_sdm.tri",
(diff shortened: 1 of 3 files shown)


common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,457 @@
...
+        "tdp_transition": (3.0, 0.0, "infinity", 0.0),  # loop, forward, skip, exit
+        "tdp_silence": (0.0, 3.0, "infinity", 20.0),
+        "tying_type": "global",
+        "nonword_phones": "",
+        "tdp_nonword": (
+            0.0,
+            3.0,
+            "infinity",
+            6.0,
+        ),  # only used when tying_type = global-and-nonword
```
michelwi, line 41: As these influence the hash but do not have an effect, better remove or comment them out. 
(Or change the tying type to "global-and-nonword" and provide some "nonword_phones")

If it was used, I would for `tdp_nonword[3]` (i.e. exit tdp) replace 6.0 with 21.0 to make it higher than silence.exit so that if in doubt the TDP select silence over nonwords

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,457 @@
...
+
+from i6_experiments.common.setups.rasr import util
+from i6_experiments.common.datasets.librispeech.cart import (
+    CartQuestionsWithStress,
+    CartQuestionsWithoutStress,
+)
+from i6_experiments.common.baselines.librispeech.default_tools import SCTK_BINARY_PATH
+
+
+def get_init_args():
```
michelwi, line 18:
```suggestion
def get_init_args() -> util.RasrInitArgs:
```
typing

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,467 @@
...
+        "tdp_scale": 1.0,
+        "tdp_transition": (3.0, 0.0, "infinity", 0.0),  # loop, forward, skip, exit
+        "tdp_silence": (0.0, 3.0, "infinity", 20.0),
+        "tying_type": "global",
+        "nonword_phones": "",
+        "tdp_nonword": (
+            0.0,
+            3.0,
+            "infinity",
+            6.0,
```
Marvin84, line 40: IMO, the exit penalty of non-word should not be less than exit penalty of silence. This might not matter so much for LBS though. Just as a general comment.

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,467 @@
...
+    :param hmm_states: number of states per label for the HMM-topology
+    :param feature_flow:
+    :param add_unknown: set to true if an unknown phoneme exists
+    """
+
+    CartQuestions = (
+        CartQuestionsWithStress if use_stress_marker else CartQuestionsWithoutStress
+    )
+
+    cart_questions_class = CartQuestions(
```
Marvin84, line 170: There was a [pull request](https://github.com/rwth-i6/i6_experiments/pull/90) where the parameter `n_phones` was introduced to give the felxibility of having also diphone CART, IMO, this should be explicitely set here, so that the user see the possibility.

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,457 @@
...
+        steps=cart_questions_class.steps,
+        max_leaves=max_leaves,
+        hmm_states=hmm_states,
+    )
+
+    cart_lda_args = {
+        "name": "cart_mono",
+        "alignment": "train_mono",
+        "initial_flow_key": feature_flow,
+        "context_flow_key": feature_flow.split("+")[0],
```
michelwi, line 187: this only works by convention and not in general. But I'll allow it.

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,457 @@
...
+    )
+
+
+def get_vtln_args(feature_flow: str = "mfcc+context+lda") -> util.GmmVtlnArgs:
+    """
+    :param feature_flow:
+    """
+    vtln_training_args = {
+        "feature_flow": {
+            "name": f"uncached_ {feature_flow}",
```
michelwi, line 262: This space is probably wrong?
```suggestion
            "name": f"uncached_{feature_flow}",
```

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,457 @@
...
+        "align_iter": 75,
+        "splits": 10,
+        "accs_per_split": 2,
+        "dump_alignment_score_report": True,
+    }
+
+    monophone_recognition_args = {
+        # GmmSystem.recognition() args:
+        "iters": [10],
+        "lm_scales": [10],
```
michelwi, line 117: Is this scale close to optimum? I think I remember that GMMs go with LM scales closer to the 20s..

Edit: all other scales are 25 so maybe
```suggestion
        "lm_scales": [25],
```

common/baselines/librispeech/ls960/gmm/baseline_args.py
JackTemaki: Okay here it is not used, but good to know for switchboard.

common/baselines/librispeech/ls960/gmm/baseline_args.py
JackTemaki: Got it, thanks!

Marvin84 (approved):

JackTemaki:
I will start to run the pipeline but keep the PR open for now.

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,460 @@
...
+
+    feature_extraction_args = {
+        "mfcc": {
+            "num_deriv": 2,
+            "num_features": None,  # confusing name: number of max features, above number -> clipped
+            "mfcc_options": {
+                "warping_function": "mel",
+                # to be compatible with our old magic number, we have to use 20 features
+                "filter_width": filter_width_from_channels(
+                    channels=20, warping_function="mel", f_max=8000
```
vieting, line 54: `warping_function="mel"` and `f_max=8000` are both default. I'm fine with explicitly having them here, but then we could also add `f_min=0` to be complete, right?

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,460 @@
...
+    costa_args = {"eval_recordings": True, "eval_lm": True}
+
+    feature_extraction_args = {
+        "mfcc": {
+            "num_deriv": 2,
+            "num_features": None,  # confusing name: number of max features, above number -> clipped
+            "mfcc_options": {
+                "warping_function": "mel",
+                # to be compatible with our old magic number, we have to use 20 features
+                "filter_width": filter_width_from_channels(
```
vieting, line 53: This here results in a filter width of 270.478 while the default is 268.258. Is this difference intended? I'm not sure where exactly the 268.258 comes from.

common/baselines/librispeech/ls960/gmm/baseline_args.py
```diff
@@ -0,0 +1,460 @@
...
+            3.0,
+            "infinity",
+            21.0,
+        ),  # only used when tying_type = global-and-nonword
+    }
+
+    costa_args = {"eval_recordings": True, "eval_lm": True}
+
+    feature_extraction_args = {
+        "mfcc": {
```
vieting, line 47: Is there a specific reason for using MFCC instead of Gammatones like in the Graz paper?

vieting (approved):
Looks good to me, just some small comments.

common/baselines/librispeech/ls960/gmm/baseline_args.py
vieting: I see, thanks for pointing that out!

common/baselines/librispeech/ls960/gmm/baseline_args.py
JackTemaki: I appreciate your generosity :)

common/baselines/librispeech/ls960/gmm/baseline_args.py
JackTemaki: Done

michelwi:
As this is "official" now, we should add this setup to the hash test.

PR #112, opened 2023-01-06
full switchboard dataset helper definitions

JackTemaki:
Here are now the re-worked Switchboard pipeline helpers.

It is not all yet sufficiently documented, but this should be a good starting point.

I am not sure about the "concurrent" settings though, as here the duration difference between the sets is quite large...


common/datasets/switchboard/eval.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    :return: hub5e00 corpus object
+    """
+    hub5e00 = get_hub5e00()
+    hub5e00_corpus_object = CorpusObject()
+    hub5e00_corpus_object.corpus_file = hub5e00.bliss_corpus
+    hub5e00_corpus_object.audio_format = "wav"
+    hub5e00_corpus_object.audio_dir = None
+    hub5e00_corpus_object.duration = durations["hub5e00"]
+    return hub5e00_corpus_object
```
Atticus1806, line 66: In my opinion it would be better to return a copy of the object here. Otherwise its easy for the user to make an error and overwrite something in the corpus_object and manipulate more experiments than desired.

common/datasets/switchboard/eval.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    :return: hub5e01 corpus object
+    """
+    hub5e01 = get_hub5e01()
+    hub5e01_corpus_object = CorpusObject()
+    hub5e01_corpus_object.corpus_file = hub5e01.bliss_corpus
+    hub5e01_corpus_object.audio_format = "wav"
+    hub5e01_corpus_object.audio_dir = None
+    hub5e01_corpus_object.duration = durations["hub5e01"]
+    return hub5e01_corpus_object
```
Atticus1806, line 97: same as for 00

common/datasets/switchboard/eval.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    :return: rt03s corpus object
+    """
+    rt03s = get_rt03s()
+    rt03s_corpus_object = CorpusObject()
+    rt03s_corpus_object.corpus_file = rt03s.bliss_corpus
+    rt03s_corpus_object.audio_format = "wav"
+    rt03s_corpus_object.audio_dir = None
+    rt03s_corpus_object.duration = durations["rt03s"]
+    return rt03s_corpus_object
```
Atticus1806, line 132: same as for 00

common/datasets/switchboard/lexicon.py
```diff
@@ -0,0 +1,147 @@
...
+    lex.add_lemma(
+        lexicon.Lemma(orth=["[UNKNOWN]"], synt=["<unk>"], eval=[[]], special="unknown")
+    )
+
+    return lex
+
+
+def get_text_lexicon(subdir_prefix: str = SUBDIR_PREFIX):
+    """
+    :param subdir_prefix:
```
Atticus1806, line 88:
```suggestion
    :param subdir_prefix: alias prefix name
```

common/datasets/switchboard/export.py
```diff
@@ -0,0 +1,83 @@
...
+from i6_experiments.common.datasets.switchboard.train import (
+    get_train_bliss_corpus_ldc,
+    get_train_bliss_corpus_i6_legacy,
+)
+
+
+def _export_eval(subdir_prefix: str = SUBDIR_PREFIX):
+    """
+    Registers bliss, stm and glm for the eval parts
+    :param subdir_prefix:
```
Atticus1806, line 20:
```suggestion
    :param subdir_prefix: alias prefix name
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    return corpus.out_corpus
+
+
+def get_train_bliss_corpus_ldc(subdir_prefix: str = SUBDIR_PREFIX) -> tk.Path:
+    """
+    Switchboard-1 training corpus based on the original LDC file
+    Uses i6-custom text processing, and the transcriptions are fully
+    identical to the "i6-legacy" version.
+
+    :param subdir_prefix:
```
Atticus1806, line 71:
```suggestion
    :param subdir_prefix: alias prefix name
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    corpus.add_alias(os.path.join(subdir_prefix, "create_train_corpus_job"))
+
+    return corpus.out_corpus
+
+
+def get_train_bliss_corpus_i6_legacy(subdir_prefix: str = SUBDIR_PREFIX) -> tk.Path:
+    """
+    i6-internal Switchboard-1 training corpus
+
+    :param subdir_prefix:
```
Atticus1806, line 101:
```suggestion
    :param subdir_prefix: alias prefix name
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    subdir_prefix = os.path.join(subdir_prefix, "Switchboard-i6-legacy")
+    train_bliss_corpus = get_train_bliss_corpus(
+        SWITCHBOARD1_LEGACY_PATH, subdir_prefix=subdir_prefix
+    )
+    return train_bliss_corpus
+
+
+def get_train_corpus_object_ldc(subdir_prefix: str = SUBDIR_PREFIX):
+    """
+    :param subdir_prefix:
```
Atticus1806, line 113:
```suggestion
    :param subdir_prefix: alias prefix name
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def get_train_bliss_corpus_ldc(subdir_prefix: str = SUBDIR_PREFIX) -> tk.Path:
+    """
+    Switchboard-1 training corpus based on the original LDC file
+    Uses i6-custom text processing, and the transcriptions are fully
+    identical to the "i6-legacy" version.
+
+    :param subdir_prefix:
+    :return:
```
Atticus1806, line 72:
```suggestion
    :return: Switchboard-1 bliss corpus
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+
+    return corpus.out_corpus
+
+
+def get_train_bliss_corpus_i6_legacy(subdir_prefix: str = SUBDIR_PREFIX) -> tk.Path:
+    """
+    i6-internal Switchboard-1 training corpus
+
+    :param subdir_prefix:
+    :return:
```
Atticus1806, line 102:
```suggestion
    :return: Swichboard-1 bliss corpus
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    train_bliss_corpus = get_train_bliss_corpus(
+        SWITCHBOARD1_LEGACY_PATH, subdir_prefix=subdir_prefix
+    )
+    return train_bliss_corpus
+
+
+def get_train_corpus_object_ldc(subdir_prefix: str = SUBDIR_PREFIX):
+    """
+    :param subdir_prefix:
+    :return:
```
Atticus1806, line 114:
```suggestion
    :return: LDC corpus object
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    corpus_object.duration = 311.61
+
+    return corpus_object
+
+
+def get_train_corpus_object_i6_legacy(
+    subdir_prefix: str = SUBDIR_PREFIX,
+) -> CorpusObject:
+    """
+    :param subdir_prefix:
```
Atticus1806, line 129:
```suggestion
    :param subdir_prefix: alias prefix name
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    :return:
```
Atticus1806, line 130:
```suggestion
    :return: legacy corpus object
```

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+def get_train_corpus_object_ldc(subdir_prefix: str = SUBDIR_PREFIX):
+    """
+    :param subdir_prefix:
+    :return:
+    """
+    corpus_object = CorpusObject()
+    corpus_object.corpus_file = get_train_bliss_corpus_ldc(subdir_prefix=subdir_prefix)
+    corpus_object.audio_format = "wav"
+    corpus_object.audio_dir = None
+    corpus_object.duration = 311.61
```
Atticus1806, line 120: Is there a reason this is not in the constants?

common/datasets/switchboard/train.py
```diff
@@ -0,0 +1,188 @@
...
+    :param subdir_prefix:
+    :return:
+    """
+    corpus_object = CorpusObject()
+    corpus_object.corpus_file = get_train_bliss_corpus_i6_legacy(
+        subdir_prefix=subdir_prefix
+    )
+    corpus_object.audio_format = "wav"
+    corpus_object.audio_dir = None
+    corpus_object.duration = 311.78
```
Atticus1806, line 138: Is there a reason this is not in the constants?

Atticus1806 (commented):
Just a lot of small things

common/datasets/switchboard/eval.py
JackTemaki: solved via frozen dataclass

common/datasets/switchboard/lexicon.py
JackTemaki: I think this is self-explainatory, and in some cases this is also output and not only alias.

common/datasets/switchboard/train.py
JackTemaki: here it is a deprecated custom corpus so we keep it out of the constants.

common/datasets/switchboard/constants.py
```diff
@@ -0,0 +1,17 @@
+SUBDIR_PREFIX = "Switchboard"
+
+durations = {"train": 311.62, "hub5e00": 3.61, "hub5e01": 6.17, "rt03s": 6.27}
+
+concurrent = {
+    "train": 60,
+    "hub5e00": 10,
```
christophmluscher, line 7: maybe 5? so that the size per sub corpus for dev and test are the same?

common/datasets/switchboard/corpus_train.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+
+    return corpus.out_corpus
+
+
+def get_train_bliss_corpus_ldc(subdir_prefix: str = SUBDIR_PREFIX) -> tk.Path:
+    """
+    Switchboard-1 training corpus based on the original LDC file
+    Uses i6-custom text processing, and the transcriptions are fully
+    identical to the "i6-legacy" version.
```
christophmluscher, line 69: I am not sure but is this the right documentation. I understood `_ldc` is the new corpora and `_i6_legacy` is the "inherited" corpora.

JackTemaki: Yes, but this is not conflicting with this text, is it? Maybe you can say how to rewrite this if it is not clear... maybe I write "transcription format" instead of "transctriptions"...

common/datasets/switchboard/lexicon.py
Atticus1806: Yes it mostly is. Was just wondering because for some function this is the docstring and then there were docs where this was just empty, so i copied it over.

common/datasets/switchboard/corpus_train.py
christophmluscher: I would add something what is actually different to the i6-legacy version. Otherwise it seems that this is the same as i6-legacy

common/datasets/switchboard/constants.py
JackTemaki: Hmm, if we want to be very accurate we could use 7 and 12...
I am not sure about this...

common/datasets/switchboard/corpus_train.py
vieting: It's also not intuitively clear to me which version is what exactly from looking at the docstrings.

common/datasets/switchboard/corpus_train.py
```diff
@@ -0,0 +1,179 @@
...
+    :param subdir_prefix: alias name prefix
+    :param skip_empty_ldc_file: exclude sw2167B which is empty in the official LDC download
+    :return: Path to switchboard training corpus
+    """
+    swb_trans_and_dict = DownloadSwitchboardTranscriptionAndDictJob()
+    swb_trans_and_dict.add_alias(
+        os.path.join(subdir_prefix, "download_trans_and_dict_job")
+    )
+
+    speakers_list = get_speakers_list_legacy(subdir_prefix=subdir_prefix)
```
vieting, line 48: Is it correct that the legacy speaker list is always used here? When calling `get_train_bliss_corpus()` (as opposed to `..._i6_legacy()`), users might not expect to get legacy behavior.

common/datasets/switchboard/lexicon.py
```diff
@@ -0,0 +1,147 @@
...
+    raw_lexicon_dir = DownloadSwitchboardTranscriptionAndDictJob()
+
+    # apply preprocessing on words to be consistent with the training corpus
+    mapped_raw_lexicon_file_job = CreateSwitchboardLexiconTextFileJob(
+        raw_lexicon_dir.out_raw_dict
+    )
+    mapped_raw_lexicon_file_job.add_alias(
+        os.path.join(subdir_prefix, "create_lexicon_text_file_job")
+    )
+    from i6_core.text.processing import PipelineJob
```
vieting, line 99: Why not importing at the top of the file?

vieting (commented):
Just some small comments.

common/datasets/switchboard/corpus_train.py
JackTemaki: I will remove this whole function.

mmz33 (approved):

vieting (approved):
Looks good to me, but I'd remove the empty file `users/rossenbach/experiments/unnormalized_tts/rc_networks/features.py` from this PR before merging (@JackTemaki).

Marvin84 (approved):

Atticus1806 (approved):

JackTemaki:
> Looks good to me, but I'd remove the empty file `users/rossenbach/experiments/unnormalized_tts/rc_networks/features.py` from this PR before merging (@JackTemaki).

Thanks for seeing this!

PR #116, opened 2023-02-02
add SMS-WSJ RETURNN datasets

vieting:
I added some helper classes which allow wrapping the [SMS-WSJ dataset](https://github.com/fgnt/sms_wsj/) in RETURNN. They basically wrap RETURNN's `MapDatasetBase` and `MapDatasetWrapper`. They are part of `i6_experiments` to allow integrating them into a pipeline. An example usage in the sisyphus config would be
```
def segment_to_rasr(seg_name):
    """
    Maps SMS-WSJ segnemt name (e.g. "0_4k6c0303_4k4c0319") to list of RASR segment names
    (e.g. ["cv_dev93/4k6c0303/0000", "cv_dev93/4k4c0319/0000"]), which can be found in HDF.
    """
    segs = str(seg_name).split("_")[1:]
    return [f"cv_dev93/{seg}/0000" for seg in segs]


returnn_config = ReturnnConfig(
  dict(
    dev={
      "class": CodeWrapper("SmsWsjMixtureEarlyAlignmentDataset"),
      "dataset_name": "cv_dev93",
      "json_path": "/path/to/sms_wsj.json",
      "rasr_num_outputs": 9001,
      "segment_to_rasr": segment_to_rasr
      "rasr_classes_hdf": "/path/to/cv_dev93_alignments.hdf",
      "pad_label": 9000,
      "seq_ordering": "default",
    },
    ...
  ),
  python_prolog=[
    SmsWsjBase,
    SmsWsjBaseWithRasrClasses,
    SmsWsjWrapper,
    SmsWsjMixtureEarlyDataset,
    SmsWsjMixtureEarlyAlignmentDataset,
  ],
  ...
)
```
I will add a baseline pipeline for a joint separation and recognition system lateron.

### DIFF
--- a/common/datasets/sms_wsj/returnn_datasets.py
+++ b/common/datasets/sms_wsj/returnn_datasets.py
@@ -283,9 +283,7 @@ class SmsWsjBaseWithRasrClasses(SmsWsjBase):
         """
         super(SmsWsjBaseWithRasrClasses, self).__init__(**kwargs)
 
-        self._rasr_classes_hdf = HDFDataset(
-            [rasr_classes_hdf], use_cache_manager=True
-        )
+        self._rasr_classes_hdf = HDFDataset([rasr_classes_hdf], use_cache_manager=True)
         self._segment_to_rasr = segment_to_rasr
         self._pad_label = pad_label
         self._hdf_data_key = hdf_data_key
@@ -293,20 +291,17 @@ class SmsWsjBaseWithRasrClasses(SmsWsjBase):
     def __getitem__(self, seq_idx: int) -> Dict[str, np.array]:
         d = self._get_seq_by_idx(seq_idx)
         rasr_seq_tags = self._segment_to_rasr(d["seq_tag"])
-        assert len(rasr_seq_tags) == d["target_signals"].shape[1], (
-            f"got {len(rasr_seq_tags)} segment names, but there are {d['target_signals'].shape[1]} target signals")
+        assert (
+            len(rasr_seq_tags) == d["target_signals"].shape[1]
+        ), f"got {len(rasr_seq_tags)} segment names, but there are {d['target_signals'].shape[1]} target signals"
         rasr_targets = [
             self._rasr_classes_hdf.get_data_by_seq_tag(rasr_seq_tag, self._hdf_data_key)
             for rasr_seq_tag in rasr_seq_tags
         ]
-        padded_len = max(
-            rasr_target.shape[0] for rasr_target in rasr_targets
-        )
+        padded_len = max(rasr_target.shape[0] for rasr_target in rasr_targets)
         for speaker_idx in range(len(rasr_targets)):
             pad_start = int(round(d["offset"][speaker_idx] / d["seq_len"] * padded_len))
-            pad_end = (
-                padded_len - rasr_targets[speaker_idx].shape[0] - pad_start
-            )
+            pad_end = padded_len - rasr_targets[speaker_idx].shape[0] - pad_start
             if pad_end < 0:
                 pad_start += pad_end
                 assert pad_start >= 0
--- a/common/datasets/sms_wsj/returnn_datasets.py
+++ b/common/datasets/sms_wsj/returnn_datasets.py
@@ -525,3 +525,67 @@ class SmsWsjMixtureEarlyAlignmentDataset(SmsWsjMixtureEarlyDataset):
                 rasr_num_outputs,
                 1,
             ]  # target alignments are sparse with the given dim
+
+
+class SmsWsjMixtureEarlyBpeDataset(SmsWsjMixtureEarlyDataset):
+    """
+    Dataset with audio mixture, target early signals and target BPE labels.
+    """
+
+    def __init__(
+        self,
+        dataset_name,
+        json_path,
+        bpe,
+        text_proc=None,
+        num_outputs=None,
+        zip_cache=None,
+        **kwargs,
+    ):
+        """
+        :param str dataset_name: "train_si284", "cv_dev93" or "test_eval92"
+        :param str json_path: path to SMS-WSJ json file
+        :param Dict[str] bpe: opts for :class:`BytePairEncoding`
+        :param Optional[Callable] text_proc: function to preprocess the transcriptions before applying BPE
+        :param Optional[Dict[str, List[int]]] num_outputs: num_outputs for RETURNN dataset
+        :param Optional[str] zip_cache: zip archive with SMS-WSJ data which can be cached, unzipped and used as data dir
+        """
+        sms_wsj_base = SmsWsjBase(
+            dataset_name=dataset_name,
+            json_path=json_path,
+            pre_batch_transform=self._pre_batch_transform,
+            scenario_map_args={"add_speech_reverberation_early": True},
+            zip_cache=zip_cache,
+        )
+        super(SmsWsjMixtureEarlyBpeDataset, self).__init__(
+            dataset_name,
+            json_path,
+            num_outputs=num_outputs,
+            zip_cache=zip_cache,
+            sms_wsj_base=sms_wsj_base,
+            **kwargs,
+        )
+        from returnn.datasets.util.vocabulary import BytePairEncoding
+        self.bpe = BytePairEncoding(**bpe)
+        self.text_proc = text_proc or (lambda x: x)
+        if num_outputs is not None:
+            self.num_outputs = num_outputs
+        else:
+            assert (
+                bpe is not None
+            ), "either num_outputs or bpe has to be given"
+            self.num_outputs["target_bpe"] = [
+                self.bpe.num_labels,
+                1,
+            ]  # target BPE labels are sparse with the given dim
+
+    def _pre_batch_transform(self, inputs: Dict[str, Any]) -> Dict[str, np.array]:
+        """
+        Used to process raw SMS-WSJ data
+        :param inputs: input as coming from SMS-WSJ
+        """
+        return_dict = SmsWsjMixtureEarlyDataset._pre_batch_transform(inputs)
+        for speaker, orth in enumerate(inputs["kaldi_transcription"]):
+            return_dict[f"target_bpe_{speaker}"] = np.array(self.bpe.get_seq(self.text_proc(orth)), dtype="int32")
+            return_dict[f"target_bpe_{speaker}_len"] = np.array(return_dict[f"target_bpe_{speaker}"].size)
+        return return_dict
--- a/common/datasets/sms_wsj/returnn_datasets.py
+++ b/common/datasets/sms_wsj/returnn_datasets.py
@@ -265,6 +265,7 @@ class SmsWsjBaseWithRasrClasses(SmsWsjBase):
         rasr_corpus=None,
         rasr_segment_prefix="",
         rasr_segment_postfix="",
+        pad_label=None,
         **kwargs,
     ):
         """
@@ -272,13 +273,16 @@ class SmsWsjBaseWithRasrClasses(SmsWsjBase):
         :param Optional[str] rasr_corpus: RASR corpus file for reading segment start and end times for padding
         :param str rasr_segment_prefix: prefix to map SMS-WSJ segment name to RASR segment name
         :param str rasr_segment_postfix: postfix to map SMS-WSJ segment name to RASR segment name
+        :param Optional[int] pad_label: target label assigned to padded areas
         :param kwargs:
         """
         super(SmsWsjBaseWithRasrClasses, self).__init__(**kwargs)
 
         # self.data_types = {"target_signals": {"dim": 2, "shape": (None, 2)}}
         self._rasr_classes_hdf = None
+        self.pad_label = pad_label
         if rasr_classes_hdf is not None:
+            assert pad_label is not None, "Label for padding is needed"
             self._rasr_classes_hdf = HDFDataset(
                 [rasr_classes_hdf], use_cache_manager=True
             )
@@ -331,9 +335,9 @@ class SmsWsjBaseWithRasrClasses(SmsWsjBase):
                     pad_end = 0
                 rasr_targets[speaker_idx] = np.concatenate(
                     [
-                        9000 * np.ones(pad_start),
+                        self.pad_label * np.ones(pad_start),
                         rasr_targets[speaker_idx],
-                        9000 * np.ones(pad_end),
+                        self.pad_label * np.ones(pad_end),
                     ]
                 )
             d["target_rasr"] = np.stack(rasr_targets).T
@@ -475,6 +479,7 @@ class SmsWsjMixtureEarlyAlignmentDataset(SmsWsjMixtureEarlyDataset):
         rasr_segment_postfix="",
         rasr_classes_hdf=None,
         rasr_corpus=None,
+        pad_label=None,
         zip_cache=None,
         **kwargs,
     ):
@@ -487,6 +492,7 @@ class SmsWsjMixtureEarlyAlignmentDataset(SmsWsjMixtureEarlyDataset):
         :param str rasr_segment_postfix: postfix to map SMS-WSJ segment name to RASR segment name
         :param str rasr_classes_hdf: hdf file with dumped RASR class labels
         :param str rasr_corpus: RASR corpus file for reading segment start and end times for padding
+        :param Optional[int] pad_label: target label assigned to padded areas
         :param Optional[str] zip_cache: zip archive with SMS-WSJ data which can be cached, unzipped and used as data dir
         """
         sms_wsj_base = SmsWsjBaseWithRasrClasses(
@@ -498,6 +504,7 @@ class SmsWsjMixtureEarlyAlignmentDataset(SmsWsjMixtureEarlyDataset):
             rasr_corpus=rasr_corpus,
             rasr_segment_prefix=rasr_segment_prefix,
             rasr_segment_postfix=rasr_segment_postfix,
+            pad_label=pad_label,
             zip_cache=zip_cache,
         )
         super(SmsWsjMixtureEarlyAlignmentDataset, self).__init__(
--- a/common/datasets/sms_wsj/returnn_datasets.py
+++ b/common/datasets/sms_wsj/returnn_datasets.py
@@ -1,0 +1,411 @@
+"""
+Dataset wrapper to allow using the SMS-WSJ dataset in RETURNN.
+"""
+
+import functools
+import json
+import numpy as np
+import os.path
+import subprocess as sp
+from typing import Dict, Tuple, Any, Optional
+# noinspection PyUnresolvedReferences
+from returnn.datasets.basic import DatasetSeq
+# noinspection PyUnresolvedReferences
+from returnn.datasets.hdf import HDFDataset
+# noinspection PyUnresolvedReferences
+from returnn.datasets.map import MapDatasetBase, MapDatasetWrapper
+# noinspection PyUnresolvedReferences
+from returnn.log import log
+# noinspection PyUnresolvedReferences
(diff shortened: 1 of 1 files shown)


common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,606 @@
...
+        :param Optional[Callable] text_proc: function to preprocess the transcriptions before applying BPE
+        :param Optional[Dict[str, List[int]]] num_outputs: num_outputs for RETURNN dataset
+        :param Optional[str] zip_cache: zip archive with SMS-WSJ data which can be cached, unzipped and used as data dir
+        """
+        sms_wsj_base = SmsWsjBase(
+            dataset_name=dataset_name,
+            json_path=json_path,
+            pre_batch_transform=self._pre_batch_transform,
+            scenario_map_args={"add_speech_reverberation_early": True},
+            zip_cache=zip_cache,
```
larissakl, line 570: There is a small error here. `SmsWsjBase` requires a `data_types` argument.
I think the definition should be similar to `SmsWsjMixtureEarlyAlignmentDataset`, so something like
```
data_types = {
            "target_signals": {"dim": 2, "shape": (None, 2)},
            "target_bpe": {
                "sparse": True,
                "dim": num_labels,
                "shape": (None, 2),
            },
        }
```

In this case, `num_labels` would be a new parameter.

vieting: Thanks for spotting this!

vieting:
I added an extension to allow using WSJ alignments directly instead of the tailored ones for an SMS-WSJ bliss corpus. A corresponding `segment_to_rasr` function needs to be given for each case now.

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,595 @@
...
+            sms_wsj_base=sms_wsj_base,
+            **kwargs,
+        )
+
+        self.text_proc = text_proc or (lambda x: x)
+        if num_outputs is not None:
+            self.num_outputs = num_outputs
+        else:
+            assert bpe is not None, "either num_outputs or bpe has to be given"
```
SimBe195, line 576: `bpe` can't be None, otherwise `BytePairEncoding(**bpe)` would already fail.

SimBe195 (changes requested):

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,595 @@
...
+
+        scenario_map_args = {
+            "add_speech_image": False,
+            "add_speech_reverberation_early": False,
+            "add_speech_reverberation_tail": False,
+            "add_noise_image": False,
+            **(scenario_map_args or {}),
+        }
+        ds = ds.map(functools.partial(scenario_map_fn, **scenario_map_args))
+        ds = ds.map(functools.partial(pre_batch_transform))
```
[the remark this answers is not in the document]
vieting: Right, we don't.
```suggestion
        ds = ds.map(pre_batch_transform)
```

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,595 @@
...
+
+        self._ds = ds
+        self._ds_iterator = iter(self._ds)
+
+        self._use_buffer = buffer
+        if self._use_buffer:
+            self._ds = self._ds.prefetch(4, 8).copy(freeze=True)
+        self._buffer = {}  # type Dict[int,[Dict[str,np.array]]]
+        self._buffer_size = 40
```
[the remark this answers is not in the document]
vieting: Good point. I think they come from an example that Christoph gave me maybe together with some try-and-error. I couldn't provide insights on this right now. Should we make this configurable? Do you think we should test this more?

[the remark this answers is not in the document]
SimBe195: Then let's make them configurable with default values. I don't think it's important to test this right now but maybe we want to at some point.

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,617 @@
...
+        """
+        print(f"Cache and unzip SMS-WSJ data from {zip_cache}", file=returnn_log.v4)
+
+        # cache file
+        try:
+            zip_cache_cached = sp.check_output(["cf", zip_cache]).strip().decode("utf8")
+            assert (
+                zip_cache_cached != zip_cache
+            ), "cached and original file have the same path"
+            local_base_dir = os.path.dirname(zip_cache_cached)
```
larissakl, line 214: I think the definition of `local_base_dir` has to be changed. When trying to run a train-job, the zip-file I used was already cached in Simon's tmp-directory and consequently it also wanted to unzip it there and not in my own tmp-directory.
Maybe `local_base_dir` could be a global variable or something, to make sure your proper directory is used.

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,617 @@
...
+        data = self._dataset[corpus_seq_idx]
+        assert "seq_tag" in data
+        return DatasetSeq(seq_idx, features=data, seq_tag=data["seq_tag"])
+
+    def init_seq_order(self, epoch: Optional[int] = None, **kwargs) -> bool:
+        """
+        Override this in order to update the buffer. get_seq_length is often called before _collect_single_seq,
+        therefore the buffer does not contain the initial indices when continuing the training from an epoch > 0.
+        """
+        out = super().init_seq_order(epoch=epoch, **kwargs)
```
larissakl, line 404:
```suggestion
        self.seq_ordering = self._seq_ordering
        out = super().init_seq_order(epoch=epoch, **kwargs)
```

Not sure if this is relevant in general, but when using `ReturnnSearchJobV2`, `self.seq_ordering` might be changed by `def search()` in `returnn/tf/engine.py`, which leads to problems with the `update_buffer` function, so it has to be reset

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,617 @@
...
+    """
+    Base class for datasets that can be used in RETURNN config.
+    """
+
+    def __init__(self, sms_wsj_base, **kwargs):
+        """
+        :param Optional[SmsWsjBase] sms_wsj_base: SMS-WSJ base class to allow inherited classes to modify this
+        """
+        if "seq_ordering" not in kwargs:
+            print("Warning: no shuffling is enabled by default", file=returnn_log.v2)
```
larissakl, line 363:
```suggestion
        if "seq_ordering" not in kwargs:
            print("Warning: no shuffling is enabled by default", file=returnn_log.v2)
        else:
            self._seq_ordering = kwargs["seq_ordering"]
```

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,617 @@
...
+                        "WARNING: reached final index of dataset, but iterator has more sequences. "
+                        "Maybe the training was restarted from an epoch > 1?",
+                        file=returnn_log.v3,
+                    )
+                print(
+                    f"Current buffer indices: {self._buffer.keys()}",
+                    file=returnn_log.v5,
+                )
+                self._ds_iterator = iter(self._ds)
+                for idx_ in range(min(self._buffer.max_size // 2, len(self))):
```
larissakl, line 192: I like the idea of having a class for the buffer, but I have faced a problem with the `update_buffer` function, more precisely, the automatic popping of the oldest inserted element. 
When the end of the dataset is reached, the buffer is "refilled" with `self._buffer.max_size // 2` sequences from the beginning of the dataset and the oldest `self._buffer.max_size // 2` sequences are popped. However, when working with batches, this may cause that sequences which are needed for the batch are removed from the buffer too soon.

This could maybe be avoided by filling one sequence from the beginning of the dataset at a time. Maybe something like:
 `for idx_ in range((seq_idx + self._buffer.max_size // 2) - len(self) + 1)`

common/datasets/sms_wsj/returnn_datasets.py
vieting: The idea was that it is desired to reuse that unzipped folder. So if it already exists, why creating another copy of it?
`unzip -n` should already do that, but it does not work if the user doesn't have write permissions, that's why I added the `chmod -R o+w <local_base_dir>`. There might be nodes in the cluster where the folder already exists without proper permissions. However, I'm not sure if there might be more issues than this.

common/datasets/sms_wsj/returnn_datasets.py
vieting: The issue with that might be that you only get here if `idx == len(self) - 1 and 0 not in self._buffer`. So probably when `seq_idx + self._buffer.max_size // 2 == len(self) - 1`. So you would have `range(0)`, right? A simpler thing might be to just increase the buffer size in your case.

common/datasets/sms_wsj/returnn_datasets.py
vieting: Could you please elaborate what the issue is in `update_buffer`?

common/datasets/sms_wsj/returnn_datasets.py
larissakl: I forgot to say that I had to remove `chmod -R o+w <local_base_dir>` because I had no permission to change permissions...

common/datasets/sms_wsj/returnn_datasets.py
larissakl: Ok, I'm sorry, what I said was not 100% right. I recreated the problem. The issue is not `update_buffer`. 
In my case, `self.seq_ordering` was changed to `"sorted_reverse"`. The `init_seq_order()` function in `MapDatasetWrapper` calls `self.get_seq_order_for_epoch`. If `self.seq_ordering` is `"sorted_reverse"` `get_seq_len()` is called and it tries to get the sequence length for sequences which are not in the buffer. Therefore, `self.seq_ordering` has to be `"default"`, so not all sequence lengths are needed. Maybe there is a better and more elegant solution to this problem, but this was the easiest one I could think of

common/datasets/sms_wsj/returnn_datasets.py
larissakl: Ok, I'm sorry, what I said was not 100% right. I recreated the problem. The issue is not `update_buffer`.
In my case, `self.seq_ordering` was changed to `"sorted_reverse"`. The `init_seq_order()` function in `MapDatasetWrapper` calls `self.get_seq_order_for_epoch`. If `self.seq_ordering` is `"sorted_reverse"`, `get_seq_len()` is called and it tries to get the sequence length for sequences which are not in the buffer. Therefore, `self.seq_ordering` has to be `"default"`, so not all sequence lengths are needed. Maybe there is a better and more elegant solution to this problem, but this was the easiest one I could think of

common/datasets/sms_wsj/returnn_datasets.py
larissakl: I don't think increasing the buffer size will help because the oldest `self._buffer.max_size // 2` sequences are popped, no matter if the buffer size is 40 or 400, you will always loose the first half of sequences in the buffer with that loop.

So tell me if I'm wrong, but due to `min(seq_idx + self._buffer.max_size // 2, len(self))`, we get there when `seq_idx + self._buffer.max_size // 2 -1 == len(self) -1` and not `seq_idx + self._buffer.max_size // 2 == len(self) - 1`. That's why I added `+1`, so we have at least `range(1)`. However you are right that the condition `0 not in self._buffer` causes a problem

vieting:
The unzipping logic was quite annoying and prone to errors. Therefore, I added a custom reader class to allow reading the data directly from the zip in the `SmsWsj` dataset. This clearly simplifies the pre-processing and I hope it does not affect the dataset performance but I don't have numbers on that.

common/datasets/sms_wsj/returnn_datasets.py
vieting: I removed that logic and read the data directly from the zip file now which should make this discussion obsolete.

common/datasets/sms_wsj/returnn_datasets.py
vieting: See other thread https://github.com/rwth-i6/i6_experiments/pull/116#discussion_r1103830555

common/datasets/sms_wsj/returnn_datasets.py
vieting: Sure, the oldest `self._buffer.max_size // 2` sequences are always popped, but if the buffer size is large enough, the remaining old sequences are sufficient to cover everything that you need for your remaining batches, no?

larissakl: I think it's easier to give an example:
Let's say `len(self)` is 125 and my buffer size is 40. My batch begins at sequence 100 and will contain 10 sequences. Now, at some point, `update_buffer` is called with `seq_idx=105`. 
Then, in the first part, the buffer is filled with `85,...,105,...,124` (because `124 == seq_idx + self._buffer.max_size // 2 -1`, so we do `for idx in range(105, 125)`). Now, the condition `if idx == len(self) - 1 and 0 not in self._buffer` is fulfilled (because `124 == len(self) - 1`). `for idx_ in range(min(self._buffer.max_size // 2, len(self)))` fills the buffer with indices 0 to 19. As a result, the first 20 elements of the buffer are popped, so the buffer is now filled with elements `105,...,124,0,...,19`. However, for the batch we need sequences 100 to 104, which are not in the buffer anymore.
That means, everything before sequence 105 is removed, there are no old sequences remaining. With an increased batch size, the problem would just appear earlier.
Again, tell me if I am wrong, maybe I am doing a mistake here.

SimBe195: Yes I agree. In general as soon as update_buffer is called on `seq_idx = len(self) - buffer_size // 2` the start of the dataset is added and the buffer is now filled with sequences `[len(self) - buffer_size // 2 ... len(self) - 1] + [0 ... buffer_size // 2 - 1]` so this only keeps the current minibatch intact if `seq_idx` happens to be the first sequence in the minibatch, regardless of buffer_size. In the example of @larissakl it only runs successfully if the batch begins at sequence 105.

I ran into this error as well and changed the instances of `// 2` to `// 3` locally which seems to give enough room for old sequences but maybe the safer solution is to only add `seq_idx + buffer_size // 2 - len(self)` sequences from the dataset start. For example when `update_buffer(118)` is called, the first `13` sequences would be added such that the buffer contains `98,...,118,...,124,0,...,12` afterwards.

BTW I think we should add some documentation that the buffer size should be set large enough so that the number of sequences in a mini-batch never exceeds `buffer_size // 2`; otherwise the training will crash.

larissakl: > I ran into this error as well and changed the instances of `// 2` to `// 3` locally which seems to give enough room for old sequences but maybe the safer solution is to only add `seq_idx + buffer_size // 2 - len(self)` sequences from the dataset start. For example when `update_buffer(118)` is called, the first `13` sequences would be added such that the buffer contains `98,...,118,...,124,0,...,12` afterwards.

That's what I meant by `for idx_ in range((seq_idx + self._buffer.max_size // 2) - len(self) + 1)`. I tried the following and it seems to work:

```
        for idx in range(seq_idx, min(seq_idx + self._buffer.max_size // 2, len(self))):
            if idx not in self._buffer:
                self._buffer[idx] = next(self._ds_iterator)
            if idx == len(self) - 1:
                if 0 not in self._buffer:
                    print(f"Reached end of dataset, reset iterator", file=returnn_log.v4)
                    try:
                        next(self._ds_iterator)
                    except StopIteration:
                        pass
                    else:
                        print(
                            "WARNING: reached final index of dataset, but iterator has more sequences. "
                            "Maybe the training was restarted from an epoch > 1?",
                            file=returnn_log.v3,
                        )
                    print(
                        f"Current buffer indices: {self._buffer.keys()}",
                        file=returnn_log.v5,
                    )
                    self._ds_iterator2 = iter(self._ds)
                for idx_ in range((seq_idx + self._buffer.max_size // 2) - len(self) + 1):
                    if idx_ not in self._buffer:
                        self._buffer[idx_] = next(self._ds_iterator2)
                    if seq_idx == len(self) - 1:
                        self._ds_iterator = self._ds_iterator2
```

It realizes the idea of only adding `seq_idx + buffer_size // 2 - len(self)` sequences from the dataset start. The iterator is only reset if 0 is not in the buffer. If this happens, we have to work with two iterators in parallel, one at the end of the dataset (until the last sequence is reached) and one at the start.

vieting: I pushed a fix which also simplifies this a bit. When we reach the last index (`len(self) - 1`) for the first time, we reset the iterator and add the index 0). Then, we always add one index more at the beginning of the dataset by using `idx % len(self)` instead of just `idx` capped at `len(self)`.

vieting: Please go ahead and test that, for my local test case the issue is resolved.

vieting: > BTW I think we should add some documentation that the buffer size should be set large enough so that the number of sequences in a mini-batch never exceeds buffer_size // 2; otherwise the training will crash.

I added that in the docstring, you can have a look whether you find that sufficient.

common/datasets/sms_wsj/returnn_datasets.py
```diff
@@ -0,0 +1,557 @@
...
+        assert "seq_tag" in data
+        return DatasetSeq(seq_idx, features=data, seq_tag=data["seq_tag"])
+
+    def init_seq_order(self, epoch: Optional[int] = None, **kwargs) -> bool:
+        """
+        Override this in order to update the buffer. get_seq_length is often called before _collect_single_seq,
+        therefore the buffer does not contain the initial indices when continuing the training from an epoch > 0.
+        """
+        out = super().init_seq_order(epoch=epoch, **kwargs)
+        buffer_index = ((epoch or 1) - 1) * self.num_seqs % len(self._dataset)
```
SimBe195, line 366: I think this logic may be buggy. I'm having trouble re-starting a training that crashed at a later sub-epoch because the `seq_idx` requested in `get_seq_length` is not present in the buffer. I think this calculation here assumes that `self.num_seqs` is the same for all sub-epochs which may be incorrect? If the total sequence count is not divisible by `partition_epoch` then `self.num_seqs` may differ between subepochs which can accumulate here when `epoch` is high.

vieting: Maybe you can try
```suggestion
        buffer_index = ((epoch or 1) - 1) % (self.partition_epoch or 0) * self.num_seqs
```
Here, the error does not accumulate much but at maximum `self.partition_epoch` times.

vieting:
Just as a note: We might fix the start index for higher epochs in `init_seq_order()` and study the effect of the buffer size. I suggest to do that after the deadline.

vieting:
As agreed offline, let's get active on this PR again. The main issue right now is the bug for restarting training at epochs > 1. There are also other open threads and the question about the effect of the buffer size.

@larissakl: @SimBe195 mentioned that you might have some idea to fix the restarting issue. Is that right?

Regarding the buffer size, I think we could also study that after merging.

larissakl:
In general, this issue should go along with the `partiton_epoch` parameter, because when restarting from an epoch divisible by `partition_epoch` it works fine.
In returnn's Dataset class (returnn/datasets/basic) the `get_seq_order_for_epoch()` function seems to take `partition_epoch` into accout. Therefore, I thought `self.partition_epoch` might not be set properly. However, unfortunately, I tested and this did not solve it. Maybe we should nevertheless check if all relevant parameters are set and if the right functions, especially in the super classes, are called

vieting:
Closing due to inactivity.

PR #119, opened 2023-02-23
update sequitur version

JackTemaki:
Originally I planned to keep Sequitur fix, but under Ubuntu 22 and Python 3.10 the currently pinned version no longer installs.


michelwi (approved):
sure.. why not.

christophmluscher (approved):

PR #122, opened 2023-04-06
add __init__

vieting:
I'd like to add the `__init__.py` file so I can import the `PyTorchOnnxHybridSystem`. Since it's under your user folder, please confirm that you're OK wth it :)


JackTemaki:
For me it works without the init, but sure!

PR #111, opened 2023-01-03
Tedlium 2 GMM Baseline

christophmluscher:
Params are still open for suggestion/improvement..

PR #127, opened 2023-04-13
TedLium-V2 Dataset

christophmluscher:
PR for only the data of TedLium-V2.
Split from #111

### DIFF
--- a/common/datasets/tedlium2/download.py
+++ b/common/datasets/tedlium2/download.py
@@ -1,0 +1,56 @@
+import os
+from dataclasses import dataclass
+from functools import lru_cache
+from typing import Any, Dict
+
+from sisyphus import tk
+
+from i6_core.datasets.tedlium2 import (
+    DownloadTEDLIUM2CorpusJob,
+    CreateTEDLIUM2BlissCorpusJob,
+)
+
+
+@dataclass()
+class TedLium2Data:
+    """Class for storing the TedLium2 data"""
+
+    data_dir: Dict[str, tk.Path]
+    lm_dir: tk.Path
+    vocab: tk.Path
+    bliss_nist: Dict[str, tk.Path]
+    stm: Dict[str, tk.Path]
+
+    def as_dict(self) -> Dict[str, Any]:
+        return {
+            "data_dir": self.data_dir,
+            "lm_dir": self.lm_dir,
+            "vocab": self.vocab,
+            "bliss_nist": self.bliss_nist,
+            "stm": self.stm,
+        }
+
+
+@lru_cache()
+def download_data_dict(output_prefix: str = "datasets") -> Dict[str, Any]:
+    download_tedlium2_job = DownloadTEDLIUM2CorpusJob()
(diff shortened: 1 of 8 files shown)


common/datasets/tedlium2/lexicon.py
```diff
@@ -0,0 +1,114 @@
...
+        compressed=True,
+    )
+    convert_lexicon_job.add_alias(os.path.join(output_prefix, "convert_text_to_bliss_lexicon_job"))
+
+    return convert_lexicon_job.out_bliss_lexicon
+
+
+@lru_cache()
+def get_bliss_lexicon(
+    output_prefix="datasets",
```
michelwi, line 73:
```suggestion
    output_prefix: str = "datasets",
```

common/datasets/tedlium2/lexicon.py
```diff
@@ -0,0 +1,114 @@
...
+        compressed=True,
+    )
+    merge_lexicon_job.add_alias(os.path.join(output_prefix, "merge_lexicon_job"))
+
+    return merge_lexicon_job.out_bliss_lexicon
+
+
+@lru_cache()
+def get_g2p_augmented_bliss_lexicon(
+    output_prefix="datasets",
```
michelwi, line 97:
```suggestion
    output_prefix: str = "datasets",
```

Atticus1806 (commented):
PR looks good to me, one thing to discuss, otherwise will be approved.

JackTemaki (changes requested):
Docstrings for every function would be nice, but the important thing is the dataclass->dict conversion which should be removed.

common/datasets/tedlium2/download.py
```diff
@@ -0,0 +1,50 @@
...
+
+from sisyphus import tk
+
+from i6_core.datasets.tedlium2 import (
+    DownloadTEDLIUM2CorpusJob,
+    CreateTEDLIUM2BlissCorpusJob,
+)
+
+
+@dataclass()
```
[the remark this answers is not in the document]
christophmluscher: using the class directly now

DanEnergetics (approved):
All looks good to me.

Atticus1806:
So @JackTemaki all changes adressed? From my side PR is approved, but since you asked for changes I will let you make the last call.

JackTemaki (approved):
I have to trust you on the lexicon/cart, but in formality all seems fine now.

JackTemaki:
@christophmluscher @Atticus1806 do you create the folder under `/work/common` or do you want to first see how the pipeline is running until we push the dataset "officially"?

Atticus1806:
For now I would run it locally because I think it is not fully functional yet. Datasets should be fine, but recognition results do not seem correct yet.

JackTemaki:
Okay then lets wait and check if we need a revision also of the data.

christophmluscher:
@Atticus1806 can we merge this for now? the pipeline is running and we can still do further PRs to change things

Atticus1806:
From my end yes.

PR #129, opened 2023-04-13
importable librispeech 960 gmm baseline

vieting:
For some setup, I'd like to import the Librispeech 960h GMM baseline pipeline to get the alignments from it. I just care about the alignments, so I don't want to run recognition (not all jobs could be imported from `/work/common/asr/` and I don't want to run them since I don't need them). These small changes allow to facilitate this.


JackTemaki:
Well, I did not plan to spend multiple hours/days on fixing weird old GMM bugs if we abandon that pipeline soon. So this PR is good to go :)

JackTemaki (approved):

PR #130, opened 2023-04-15
Extend ReturnnRasrDataInput to produce valid Dataset config dict

DanEnergetics:
Similar to the `OggZipHdfDataInput` the `ReturnnRasrDataInput` gets a working `get_data_dict` implementation. This is useful for replacing the `ReturnnRasrTrainingJob` with the standard `ReturnnTrainingJob` for more functionality, e.g. producing different feature flow files for train and dev (see also [this discussion](https://github.com/rwth-i6/i6_core/issues/165))

### DIFF
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -8,7 +8,7 @@ __all__ = [
 ]
 
 import copy
-from dataclasses import dataclass
+from dataclasses import dataclass, asdict
 from typing import Any, Dict, List, Optional, Tuple, Type, TypedDict, Union
 
 from sisyphus import tk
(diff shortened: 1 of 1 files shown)


Marvin84:
My understanding was that we would for now include all different subroutines needed for everyone, leaving the freedom of choosing, and once everything is available we will decide how to merge?

Marvin84:
The. main point for having this is to be able to have separate feature
flows for train and dev. Do other approaches offer this possibility?

DanEnergetics:
I'm also okay with more (offline) discussions and would even welcome it. Regarding your comments, @JackTemaki, I'm unsure about these things:
- Where to copy `ReturnnRasrTrainingJob.create_config` function? Another module? As a static method of the `ReturnnRasrDataInput` class?
- Which parameters to drop? I know which one I'm not using but I cannot be sure that they aren't relevant to someone. Right now exactly those parameters are hashed that the `ReturnnRasrTrainingJob` uses for its hash.

michelwi:
>  I leave this also to [...] @michelwi to decide.

I will most likely not use or work with this code, so you may do however you want.

JackTemaki:
> * I know which one I'm not using but I cannot be sure that they aren't relevant to someone

Exactly this is the problem, this should be cleaned up so that we actually know what the parameters are doing. But I understand this is a time critical issue, and you just want working code. Then please add a comment to the "get" functions that this is outdated code triggering unknown parameters of RASR via the ReturnnRasrTrainingJob and should be used with care. 

I would still require to unify the input features for the dataset separately.

JackTemaki (approved):
Oh nice, you even changed it to frozen dataclass :) I did not want to complain about everything so I skipped on that one.

Maybe some extra docstrings, otherwise good to go

DanEnergetics:
Yes no problem :D it was mostly necessary because the `ReturnnRasrTrainingJob.create_config` method does not have any default values.
The docstrings will be added by tomorrow.

christophmluscher (changes requested):

christophmluscher (approved):

PR #133, opened 2023-04-17
fix: Fix jobs dict not existing

NeoLegends:
```
}, feature_extraction=None, stm_args=None, stm_paths={'dev-other': <Path /u/mgunz/setups/2023-04--thesis-baselines/...rt/CorpusToStmJob.cWMBwQxemXAT/output/corpus.stm>})
    184
    185             for job_type in ["search", "lat2ctm", "score", "optlm"]:
--> 186                 self.jobs[corpus_key][job_type] = {}
        self.jobs = {}
        corpus_key = 'dev-other'
        job_type = 'search'
    187
    188     @staticmethod

KeyError: 'dev-other'
```

### DIFF
--- a/common/setups/rasr/base_decoder.py
+++ b/common/setups/rasr/base_decoder.py
@@ -182,8 +182,6 @@ class BaseDecoder:
                 )
                 self.feature_flows[corpus_key] = features.basic_cache_flow(feature_path)
 
-            self.jobs[corpus_key] = {job_type: {} for job_type in ["search", "lat2ctm", "score", "optlm"]}
-
     @staticmethod
     def _get_scales_string(
         am_scale: Union[float, tk.Variable],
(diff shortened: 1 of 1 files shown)


christophmluscher (changes requested):
I don't like the place of that code that much...

christophmluscher:
I ll take a look at home

JackTemaki (changes requested):
I agree with @christophmluscher that the position it is now is clearly wrong. `_get_scales_string` should not initialize things. @NeoLegends can determine why this is not intialized? is `init_eval_datasets` never called in your case? I would assume the actual error here is somewhere else.

NeoLegends:
The issue is that the `corpus_key` there isn't actually the corpus key, but the "derived" corpus key (as the code calls it) which additionally includes various scales, like the prior scales and so on. But I agree this is a bandaid fix, not a proper solution. Let's do this differently.

PR #136, opened 2023-04-27
small fix for segment chunk size in ReturnnRasrDataInput

Marvin84:
Un-hardcode the value

### DIFF
--- a/common/setups/rasr/util/nn.py
+++ b/common/setups/rasr/util/nn.py
@@ -77,7 +77,7 @@ class ReturnnRasrDataInput:
         features: Optional[Union[RasrCacheTypes, Dict[str, RasrCacheTypes]]] = None,
         acoustic_mixtures: Optional[Union[tk.Path, str]] = None,
         feature_scorers: Optional[Dict[str, Type[rasr.FeatureScorer]]] = None,
-        shuffle_data: bool = True,
+        shuffling_paramters: Optional[Dict] = None,
         stm: Optional[tk.Path] = None,
         glm: Optional[tk.Path] = None,
         returnn_rasr_training_args: Optional[ReturnnRasrTrainingArgs] = None,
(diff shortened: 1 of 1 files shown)


Marvin84:
In the current solution, the parameter is set only when `update_crp_with` is called. I am not sure if this is right. I think adding this very specific parameter during the class initialization is not optimal. I would then maybe pass it by the `kwargs`?
Not sure. Please suggest!

Atticus1806:
I dont know about the insides of this job and if `update_crp_with` is always called before `get_crp`, but this would break if its not the case.
Maybe `self.segment_order_sort_by_time_length_chunk_size` can be initialized with `None` or (if already available during `__init__`) with the correct value and then check for `None` in `get_crp` and raise an error.

Marvin84:
> I but this would break

Yes, exactly. But I also think that this is too much of a low level parameter and should not be set explicitly during `__init__`. So what do you think about passing it via `kwargs`?

Atticus1806:
I am not sure, I think `kwargs` might hide it. If its required as function input maybe just add it to the function as parameter?

christophmluscher:
`update_crp_with` will not always be called before `get_crp`

maybe removing `shuffle_data` and introducing `shuffling_parameters`?

```
shuffling_parameters = {
    "shuffle": True,
    "time_length_chunk_size": 384,
}
```

christophmluscher:
I think the general question how likely will the parameter list be extended... kwargs should be fine if the documentation makes the option clear..

Marvin84:
> maybe removing `shuffle_data` and introducing `shuffling_parameters`?

Not sure if this is a good idea, one can shuffle without sorting by length. I actually do not understand why we are putting these two together, maybe some legacy setup?

JackTemaki:
> > maybe removing `shuffle_data` and introducing `shuffling_parameters`?
> 
> Not sure if this is a good idea, one can shuffle without sorting by length. I actually do not understand why we are putting these two together, maybe some legacy setup?

None of this code makes sense, so yes, this was done here like this purely for legacy reasons. I have no opinion here, because this whole block can be already considered deprecated. I think I am fine with just merging this without discussion (from my side), because anyway at some point we will either completely abandon the extern rasr dataset or at least having having a consistent helper for that and not use it with the ReturnnRasrTrainingJob.

JackTemaki (approved):

christophmluscher (approved):

PR #150, opened 2023-07-05
Add CreateFairseqLabeledDataJob

AndreasPlt:
@vieting Draft pull request for the jobs regarding task creation for fairseq finetuning. 

To-Dos:
- fix FairseqCorpusCreationJob to not unnecessarily copy audio files
- fix FairseqTaskCreationJob, currently there is bug regarding path concatenation


users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+    def __init__(self, corpus_path):
+        """
+        :param [tk.Path]|tk.Path corpus_path: list of paths or single path to corpus xml files.
+        """
+        if isinstance(corpus_path, tk.Path):
+            self.corpus_paths = [corpus_path]
+        else:
+            assert isinstance(corpus_path, list)
+            self.corpus_paths = corpus_path
+        assert min([isinstance(path, tk.Path) for path in self.corpus_paths])
```
vieting:
```suggestion
        assert all([isinstance(path, tk.Path) for path in self.corpus_paths])
```

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    def __init__(self, corpus_path):
+        """
+        :param [tk.Path]|tk.Path corpus_path: list of paths or single path to corpus xml files.
+        """
+        if isinstance(corpus_path, tk.Path):
+            self.corpus_paths = [corpus_path]
+        else:
+            assert isinstance(corpus_path, list)
+            self.corpus_paths = corpus_path
```
vieting:
```suggestion
        if not isinstance(corpus_path, list):
            corpus_path = [corpus_path]
        self.corpus_paths = corpus_path
```
I guess it's debatable if we should rename `corpus_path` to `corpus_paths` from the beginning then.

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+        assert min([isinstance(path, tk.Path) for path in self.corpus_paths])
+
+        self.dest_paths = [self.output_path(f"corpus_{i}", directory=True) for i in range(len(self.corpus_paths))]
```
vieting:
```suggestion
        self.out_dest_paths = [self.output_path(f"corpus_{i}", directory=True) for i in range(len(self.corpus_paths))]
```
Please check the contributing guidelines for our recipes: https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTING.md

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+        :param float valid_percent: percentage of files to be in validation set
+        :param int seed: random seed for splitting into train and valid set
+        :param str|None path_must_contain: if set, path must contain this substring
+            for a file to be included in the manifest
+        """
+        if isinstance(audio_dir_path, tk.Path):
+            self.audio_dir_paths = [audio_dir_path]
+        else:
+            assert isinstance(audio_dir_path, list)
+            self.audio_dir_paths = audio_dir_path
```
vieting: See above

AndreasPlt:
@vieting update: I just rewrote the task creation job, combined the previous two into one, fixed the bug and removed the copy of audio files. Now, the needed task files are created directly, without first creating a corpus directory.

The file creation works. However, I did not test if the fairseq finetune works with these created files.

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+
+        self.file_extension = file_extension
+        self.valid_percent = valid_percent
+        assert 0.0 <= self.valid_percent <= 1.0
+        self.seed = seed
+        self.path_must_contain = path_must_contain
+
+        self.rqmt = {"time": 6, "mem": 8, "cpu": 1}
+
+        self.out_task_path = self.output_path("task", directory=True)
```
vieting: Do you need that folder? Couldn't the files below just all be in the normal `output/` folder?

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        if isinstance(audio_dir_path, tk.Path):
+            self.audio_dir_paths = [audio_dir_path]
+        else:
+            assert isinstance(audio_dir_path, list)
+            self.audio_dir_paths = audio_dir_path
+        assert min([isinstance(path, tk.Path) for path in self.audio_dir_paths])
+
+        
+        self.manifest_audio_paths = None
```
vieting: This seems to be unused.

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        #self.create_manifest()
+        self.create_tsv()
+        self.create_labels(split="train")
+        self.create_labels(split="valid")
+        self.download_dict_ltr()
+
+    def create_tsv(self):
```
vieting: What is the difference of this and the output of the manifest creation job that we already have?

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+                    " ".join(list(transcriptions[dir][part].replace(" ", "|"))) + " |",
+                    file=ltr_out,
+                )
+
+
+    def download_dict_ltr(self):
+        """
+        Downloads the dict.ltr.txt file for fairseq wav2vec finetuning.
+        """
+        subprocess.check_call(["wget", "-O", self.out_dict_ltr_path.get(), "https://dl.fbaipublicfiles.com/fairseq/wav2vec/dict.ltr.txt"])
```
vieting: Since it's just a rather small file, I'd prefer to store it as a string here and print it into a file rather than relying on this link which might be changed or removed in the future.

users/vieting/jobs/fairseq_labels.py
```diff
@@ -0,0 +1,252 @@
...
+            train_data = []
+            search_path = os.path.join(dir_path, "**/*." + self.file_extension)
+
+            for path in glob.iglob(search_path, recursive=True):
+                frames = soundfile.info(path).frames
+                path = os.path.realpath(path)
+
+                #if self.path_must_contain and self.path_must_contain not in path:
+                if False:
+                    continue
```
vieting: That's WIP I guess.

vieting (changes requested):
Just some general comments so far. Sorry, I started before you pushed the changes. I'll have a quick look.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+import random
+import subprocess
+import os
+
+from i6_core.lib import corpus
+
+from sisyphus import Job, Task, tk
+
+
+class CreateTaskDataJob(Job):
```
vieting, line 13:
```suggestion
class CreateFairseqLabeledDataJob(Job):
```
I think just task data is too generic. If you have a better suggestion than mine, feel free to propose it.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+    - train.tsv
+    - train.ltr
+    - train.wrd
+    - valid.tsv
+    - valid.ltr
+    - valid.wrd
+    - dict.ltr.txt
+
+    For the script see https://github.com/pytorch/fairseq/blob/main/examples/wav2vec/wav2vec_manifest.py for .tsv creation,
+    https://github.com/facebookresearch/fairseq/blob/91c364b7ceef8032099363cb10ba19a85b050c1c/examples/wav2vec/libri_labels.py aswell as
```
vieting, line 25:
```suggestion
    https://github.com/facebookresearch/fairseq/blob/91c364b7ceef8032099363cb10ba19a85b050c1c/examples/wav2vec/libri_labels.py as well as
```

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+        :param float valid_percent: percentage of files to be in validation set
+        :param int seed: random seed for splitting into train and valid set
+        :param str|None path_must_contain: if set, path must contain this substring
+            for a file to be included in the task
+        """
+        if isinstance(corpus_paths, tk.Path):
+            self.corpus_paths = [corpus_paths]
+        else:
+            assert isinstance(corpus_paths, list)
+            self.corpus_paths = corpus_paths
```
vieting, line 49: See previous review.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+        assert min([isinstance(path, tk.Path) for path in self.corpus_paths])
```
vieting, line 50: See previous review.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+
+        self.file_extension = file_extension
+        self.valid_percent = valid_percent
+        assert 0.0 <= self.valid_percent <= 1.0
+        self.seed = seed
+        self.path_must_contain = path_must_contain
+
+        self.rqmt = {"time": 6, "mem": 8, "cpu": 1}
+
+        self.out_task_path = self.output_path("task", directory=True)
```
vieting, line 60: This seems to be unused.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+        self.out_train_ltr_path = self.output_path("task/train.ltr")
+        self.out_train_wrd_path = self.output_path("task/train.wrd")
+        self.out_valid_ltr_path = self.output_path("task/valid.ltr")
+        self.out_valid_wrd_path = self.output_path("task/valid.wrd")
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        #self.create_manifest()
```
vieting, line 76:
```suggestion
```

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+    the issue https://github.com/facebookresearch/fairseq/issues/2493 for .wrd and .ltr creation.
+    """
+
+    def __init__(
+        self,
+        corpus_paths,
+        file_extension="wav",
+        valid_percent=0.01,
+        seed=42,
+        path_must_contain=None
```
vieting, line 35: Please annotate types here rather than in the docstring below.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        Creates both tsv files (train.tsv, valid.tsv) and labels (train.ltr, train.wrd, valid.ltr, valid.wrd) from the given corpora.
+        """
+        rand = random.Random(self.seed)
+        
+        common_dir = self.get_common_dir()
+
+        valid_tsv = (
+            open(self.out_valid_tsv_path, "w")
+            if self.valid_percent > 0
```
vieting, line 108: I don't think you need to condition it here. If `valid_percent` is 0, no files would be added. I think the difference is that in this case you would get an empty file, but I think that's intended behavior anyway because otherwise you define an output that then does not exist.

users/vieting/jobs/fairseq_finetuny.py
```diff
@@ -0,0 +1,185 @@
...
+        train_ltr.close()
+        train_wrd.close()
+
+
+
+    def download_dict_ltr(self):
+        """
+        Downloads the dict.ltr.txt file for fairseq wav2vec finetuning.
+        """
+        subprocess.check_call(["wget", "-O", self.out_dict_ltr_path.get(), "https://dl.fbaipublicfiles.com/fairseq/wav2vec/dict.ltr.txt"])
```
vieting, line 185: See previous review.

vieting (changes requested):
Some more comments. It looks much more concise already.

users/vieting/jobs/fairseq_labels.py
AndreasPlt: How could I then "`tk.register()`" the whole task folder at once? My idea behind this `task/ `folder was to combine all files into one folder so that one can just do `tk.register()` this `task/` folder instead of all files separately. Would this be also possible when removing the `task/` folder and saving all the created files in the `output/` folder directly

users/vieting/jobs/fairseq_finetuny.py
AndreasPlt: Regarding `self.out_task_path`: See comment above for similar line
Regarding `self.rqmt`: Should I completely remove this? I just copied the `rqmt` from the manifest creation job, assuming that this is necessary for every job class

AndreasPlt:
@vieting Just fixed most of your requests, thank you for the detailed review! For the remaining requests, I commented them with question I had with them.

users/vieting/jobs/fairseq_labels.py
vieting: I see. I don't think you'd need to register it, but off the top of my head I don't know how you would refer to the folder in a clean way (like here `self.out_task_path`). So let's just keep it like that.

users/vieting/jobs/fairseq_finetuny.py
vieting: Yes, see comment above.

`self.rqmt` should stay. It's also used in `self.tasks()`. I just think it could be reduced, right? You mentioned the job takes 5-10 minutes, so `"time": 1` should be more than enough. Regarding memory, I'm not sure how much you'd need there, but probably also much less. You can give it a try, but I don't think it's too important to "tune" this.

users/vieting/jobs/fairseq_finetune.py
```diff
@@ -0,0 +1,194 @@
...
+V 3936
+K 3456
+' 1023
+X 636
+J 598
+Q 437
+Z 213
+"""
+        with open(self.out_dict_ltr_path.get(), 'w') as f:
+            f.write(dict_ltr_content)
```
vieting, line 194: Add empty line at file end.

users/vieting/jobs/fairseq_finetune.py
```diff
@@ -0,0 +1,194 @@
+import soundfile
```
vieting, line 1: I'd like to keep the filename a bit more generic so that we could possibly add other fairseq (but not necessarily fine-tinung) related jobs here. So `fairseq.py` seems reasonable to me.

vieting (changes requested):
Looks good to me for the most parts, just commented some nits. Please give an update and mark the PR as ready once you tested everything.

users/vieting/jobs/fairseq_finetuny.py
AndreasPlt: I just tested to reduce the memory, however then the manager somehow stopped logging after some time, leading to some weird error messages, I would just leave that then as it is.
Regarding the `time` requirement, the 5-10mins was only when creating labeled data from a single corpus. However, when training with LS960 for example, we have to create labeled data from 3 corpus files, leading to a much longer execution time. I would probably just leave this as it is as well.

AndreasPlt:
I tested the output files for usage with fairseq, it seems to work as it the training does not crash in the first hour (even with multiple  corpora). However, I am not 100% sure that this works correctly as I didn't have a look under the hood of the fairseq training. Might need to look at it, especially if there will occur unexpected bugs during finetuning, this job might be a first suspect.

users/vieting/jobs/fairseq.py
```diff
@@ -191,4 +176,24 @@ def create_dict_ltr(self):
...
-            f.write(dict_ltr_content)
\ No newline at end of file
+            f.write(dict_ltr_content)
+
+    
+    def get_common_dir(self):
+        """
+        Returns the common directory of all audios given in the corpora.
+        """
+        # TODO test this
```
vieting, line 186: Is it tested now? Then please remove the comment.

vieting (changes requested):
Is the `CreateFairseqLabeledDataJob` ready and tested now? Then please move the config for fine-tuning in a separate PR and we can merge this PR here.

users/vieting/jobs/fairseq.py
AndreasPlt: done!

AndreasPlt:
The job is now tested and the PR ready for merge!

vieting (approved):

PR #153, opened 2023-07-17
Add Ogg and BPE preparation functions for Ted2

mmz33:


common/datasets/tedlium2/corpus.py
```diff
@@ -93,3 +93,40 @@ def get_stm_dict(output_prefix: str = "datasets") -> Dict[str, tk.Path]:
     :return:
     """
     return download_data_dict(output_prefix=output_prefix).stm
+
+
+def get_ogg_zip_dict(
+    output_prefix: str = "datasets",
+    returnn_python_exe: Optional[tk.Path] = None,
+    returnn_root: Optional[tk.Path] = None,
+    bliss_to_ogg_job_rqmt: Optional[Dict] = None,
```
JackTemaki, line 102:
```suggestion
    bliss_to_ogg_job_rqmt: Optional[Dict[str, Any]] = None,
```

common/datasets/tedlium2/corpus.py
```diff
@@ -93,3 +93,40 @@ def get_stm_dict(output_prefix: str = "datasets") -> Dict[str, tk.Path]:
     :return:
     """
     return download_data_dict(output_prefix=output_prefix).stm
+
+
+def get_ogg_zip_dict(
+    output_prefix: str = "datasets",
```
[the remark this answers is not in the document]
Atticus1806: But for `get_bpe_settings` there is output. So I am not sure if I like the `subdir_prefix` name.

Atticus1806 (approved):

common/datasets/tedlium2/corpus.py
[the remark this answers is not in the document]
vieting: For Librispeech, it's also called `output_prefix`.

common/datasets/tedlium2/corpus.py
```diff
@@ -93,3 +93,39 @@ def get_stm_dict(output_prefix: str = "datasets") -> Dict[str, tk.Path]:
...
+    :param tk.Path returnn_root: python to returnn root
+    :param dict bliss_to_ogg_job_rqmt: rqmt for bliss to ogg job
+    :return: dictionary with ogg zip paths for each corpus (train, dev, test)
+    """
+    from i6_core.returnn.oggzip import BlissToOggZipJob
+
+    ogg_zip_dict = {}
+    bliss_corpus_dict = get_bliss_corpus_dict(audio_format="wav", output_prefix=subdir_prefix)
+    for name, bliss_corpus in bliss_corpus_dict.items():
+        ogg_zip_job = BlissToOggZipJob(
```
vieting, line 120: Did you already run this? I often prefer to add segments to parallelize this. How long does it take without?

common/datasets/tedlium2/corpus.py
```diff
@@ -93,3 +93,39 @@ def get_stm_dict(output_prefix: str = "datasets") -> Dict[str, tk.Path]:
...
+            bliss_corpus,
+            no_conversion=False,
```
vieting, line 122:
```suggestion
            no_conversion=False,  # cannot be used for corpus with multiple segments per recording
```

common/datasets/tedlium2/corpus.py
```diff
@@ -93,3 +93,39 @@ def get_stm_dict(output_prefix: str = "datasets") -> Dict[str, tk.Path]:
...
+            returnn_python_exe=returnn_python_exe,
+            returnn_root=returnn_root,
+        )
```
vieting, line 125: I'd like to add rasr caches for synchronization sometimes. Can we maybe add something like this:
```suggestion
            **(extra_args[name] or {}),
        )
```

common/datasets/tedlium2/corpus.py
mmz33: yes i already run this. it took 4 hours i think.

common/datasets/tedlium2/corpus.py
mmz33: yes we can add this. note that an assert statement is also added to the bliss-to-ogg script now so it will crash anw.

common/datasets/tedlium2/corpus.py
mmz33: ok

common/datasets/tedlium2/corpus.py
vieting: Yes, I saw that, that's good :+1: I think the comment still helps to make it clear why this is the case.

common/datasets/tedlium2/corpus.py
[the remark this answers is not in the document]
mmz33: So what should we decide here? it really does not matter for me.

common/datasets/tedlium2/corpus.py
mmz33: can you give me an example how would you use the rasr features here? you just want to have `rasr_cache` as parameter or?

JackTemaki (approved):

vieting (approved):

PR #154, opened 2023-07-18
swb hybrid with no/time-only specaugment

Max-Ryujin:
`time_specaug.py` is based on `specaug_jingjing.py` and adds a parameter for deselecting specaug.
`enable_specaug` is default True and has to be set to False.
`specaug_time_only` can be set to True, to disable frequency specaug.


users/vieting/experiments/switchboard/hybrid/feat/baseline_args.py
```diff
@@ -26,6 +26,10 @@
     specaug_layer_sorted,
     get_funcs_sorted,
 )
+from .time_specaug import (
```
vieting, line 29:
```suggestion
from .specaug_time import ( 
```
I think this file name would be more consistent with the other SpecAugment-related file names.

users/vieting/experiments/switchboard/hybrid/feat/baseline_args.py
```diff
@@ -221,11 +227,19 @@ def get_returnn_config(
...
-            prolog = get_funcs_jingjing()
+                if specaug_time_only:
+                    network["source"] = specaug_layer_only_time(in_layer=["features"])
+                    prolog = get_funcs_only_time()
+                else:
+                    network["source"] = specaug_layer_jingjing(in_layer=["features"])
+                    prolog = get_funcs_jingjing()
 
         network = fix_network_for_sparse_output(network)
+    else:
```
vieting, line 240:
```suggestion
    else:
        assert not specaug_mask_sorting and not specaug_after_first_layer and not specaug_time_only, "specaug options are specified, but enable_specaug=False"
```
This assertion can help to avoid misunderstandings like by accident switching SpecAugment off even one of those options is desired. Please format it correctly using black or similar, because most likely the line is too long right now.

users/vieting/experiments/switchboard/hybrid/feat/baseline_args.py
```diff
@@ -221,11 +227,19 @@ def get_returnn_config(
                 network["features"]["subnetwork"]["specaug"] = specaug_layer_jingjing(in_layer=["conv_h_act"])
                 network["features"]["subnetwork"]["conv_h_split"]["from"] = "specaug"
                 network["source"] = {"class": "copy", "from": "features"}
+                prolog = get_funcs_jingjing()
             else:
-                network["source"] = specaug_layer_jingjing(in_layer=["features"])
-            prolog = get_funcs_jingjing()
+                if specaug_time_only:
```
vieting, line 232: It should be possible to have time-only SpecAugment after the first layer as well.

users/vieting/experiments/switchboard/hybrid/feat/time_specaug.py
```diff
@@ -0,0 +1,52 @@
+from .specaug_jingjing import (
```
vieting, line 1: See comment for file name above.

vieting (changes requested):
Looks mostly good to me, just some small comments.

vieting (approved):

PR #155, opened 2023-07-19
Add OnnxPrecomputedHybridSystem

vieting:
This PR adds a system class to run recognition with the onnx forward node in the feature flow and `PrecomputedHybridFeatureScorer`. The function `make_precomputed_hybrid_onnx_feature_flow` is very similar to `make_precomputed_hybrid_tf_feature_flow` in i6_core and I use `add_tf_flow_to_base_flow` here. I'd suggest to do a separate PR in i6_core to generalize and allow using the functions there for both tf and onnx.

### DIFF
--- a/common/setups/rasr/onnx_precomputed_hybrid_system.py
+++ b/common/setups/rasr/onnx_precomputed_hybrid_system.py
@@ -23,7 +23,6 @@ class OnnxPrecomputedHybridSystem(HybridSystem):
         name: str,
         returnn_config: returnn.ReturnnConfig,
         checkpoints: Dict[int, Union[returnn.Checkpoint, returnn.PtCheckpoint]],
-        train_job: Union[returnn.ReturnnTrainingJob, returnn.ReturnnRasrTrainingJob],
         prior_scales: List[float],
         pronunciation_scales: List[float],
         lm_scales: List[float],
@@ -36,10 +35,9 @@ class OnnxPrecomputedHybridSystem(HybridSystem):
         rtf: int,
         mem: int,
         epochs: Optional[List[int]] = None,
-        quantize_dynamic: bool = False,
         needs_features_size: bool = True,
         acoustic_mixture_path: Optional[tk.Path] = None,
-        nn_prior: bool = True,
+        nn_prior: bool = False,
         **kwargs,
     ):
         """
--- a/common/setups/rasr/onnx_precomputed_hybrid_system.py
+++ b/common/setups/rasr/onnx_precomputed_hybrid_system.py
@@ -127,14 +127,13 @@ class OnnxPrecomputedHybridSystem(HybridSystem):
                 onnx_job.add_alias(f"export_onnx/{name}/epoch_{epoch}")
                 onnx_model = onnx_job.out_onnx_model
 
-                io_map = {
-                    "features": "data",
-                    "output": "classes"
-                }
+                io_map = {"features": "data", "output": "classes"}
                 if needs_features_size:
                     io_map["features-size"] = "data_len"
                 onnx_flow = make_precomputed_hybrid_onnx_feature_flow(
-                    onnx_model=onnx_model, io_map=io_map, cpu=kwargs.get("cpu", 1),
+                    onnx_model=onnx_model,
+                    io_map=io_map,
+                    cpu=kwargs.get("cpu", 1),
                 )
                 flow = add_tf_flow_to_base_flow(feature_flow, onnx_flow, tf_fwd_input_name="onnx-fwd-input")
 
--- a/common/setups/rasr/onnx_precomputed_hybrid_system.py
+++ b/common/setups/rasr/onnx_precomputed_hybrid_system.py
@@ -5,7 +5,7 @@ from sisyphus import tk
 
 import i6_core.rasr as rasr
 import i6_core.returnn as returnn
-from i6_core.returnn.compile import OnnxExportJob
+from i6_core.returnn.compile import TorchOnnxExportJob
 from i6_core.returnn.flow import make_precomputed_hybrid_onnx_feature_flow, add_fwd_flow_to_base_flow
 from i6_experiments.common.setups.rasr.hybrid_system import HybridSystem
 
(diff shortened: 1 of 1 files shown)


common/setups/rasr/onnx_precomputed_hybrid_system.py
```diff
@@ -0,0 +1,167 @@
...
+                if needs_features_size:
+                    io_map["features-size"] = "data_len"
+                onnx_flow = make_precomputed_hybrid_onnx_feature_flow(
+                    onnx_model=onnx_model,
+                    io_map=io_map,
+                    cpu=kwargs.get("cpu", 1),
+                )
+                flow = add_tf_flow_to_base_flow(feature_flow, onnx_flow, tf_fwd_input_name="onnx-fwd-input")
+
+                if nn_prior:
```
Atticus1806, line 140: I think this check rather would construct a prior file which is given to the precomputed scorer. 
So I think we can do smth. like:
```
                    scorer = rasr.PrecomputedHybridFeatureScorer(
                        prior_mixtures=acoustic_mixture_path,
                        priori_scale=prior,
                        prior_file=ComputeNNPriorJob(...).out_prior_file if nn_prior else None
                    )

Comment (vieting) [reply common/setups/rasr/onnx_precomputed_hybrid_system.py:140]: Yes. But I didn't use that so far and you were using a custom `ReturnnForwardComputePriorJob`, so I thought we might first add that or agree on a common way to compute the priors and then we can add it here.

**Comment by vieting** [comment]
I added a PR on generalizing the flow functions: https://github.com/rwth-i6/i6_core/pull/431

The adaptations to that for this PR still need to be added.

**Comment by JackTemaki** [comment]
Ideally we might want to implement this as Decoder, instead of a modification to the HybridSystem. But on the other hand, I have no pipeline for that... @christophmluscher you have maybe an example?

**Comment by christophmluscher** [comment]
I have a `BaseDecoder` and a `HybridDecoder` but I only have a decoding pipeline so far. I would move forward with the `System` class and later we can refactor this, since this does not change hashes.

**Comment by vieting** [comment]
Yes, I also needed to add that first :sweat_smile:

**Review by christophmluscher — APPROVED** [review]

**Comment by christophmluscher** [comment]
For me, this looks good. I would suggest moving to "normal" PR. and merging afer the tests work.

**Comment by vieting** [comment]
Thanks, @christophmluscher!

https://github.com/rwth-i6/i6_core/pull/431 is merged. I'll move the PR to ready for review.

@JackTemaki do you already have feedback after testing this?

File: common/setups/rasr/onnx_precomputed_hybrid_system.py [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+        search_parameters: Dict,
+        lattice_to_ctm_kwargs: Dict,
+        parallelize_conversion: bool,
+        rtf: int,
+        mem: int,
+        epochs: Optional[List[int]] = None,
+        quantize_dynamic: bool = False,
+        needs_features_size: bool = True,
+        acoustic_mixture_path: Optional[tk.Path] = None,
+        nn_prior: bool = True,
```
JackTemaki, line 42: The default should not result in an `NotImplementedError`, so:


```suggestion
        nn_prior: bool = False,
```

common/setups/rasr/onnx_precomputed_hybrid_system.py
```diff
@@ -0,0 +1,109 @@
...
+        optimize_am_lm_scale: bool,
+        recognition_corpus_key: str,
+        feature_flow_key: str,
+        search_parameters: Dict,
+        lattice_to_ctm_kwargs: Dict,
+        parallelize_conversion: bool,
+        rtf: int,
+        mem: int,
+        epochs: Optional[List[int]] = None,
+        quantize_dynamic: bool = False,
```
JackTemaki, line 39:
```suggestion
```

This no longer exists as explicit parameter.

JackTemaki (changes requested):
Actually, I have some small comments after testing.

vieting:
Thanks for the suggestions!

common/setups/rasr/onnx_precomputed_hybrid_system.py
christophmluscher: But the default should be to compute the prior. Can we use the `ReturnnComputePriorJobV2` here?
I'd rather have the PR merged so that the default throws an error.

christophmluscher: So that the user must make a decision to use the acoustic mixtures for the prior

vieting: I don't mind. We should add support for NN prior either way. Maybe @Atticus1806 could merge his local code at some point? I'm inclinded to merge this PR before, though. And I'm fine with either default until then. @JackTemaki what do you think?

JackTemaki: What about no default for now?

common/setups/rasr/onnx_precomputed_hybrid_system.py
```diff
@@ -0,0 +1,107 @@
...
+        feature_flow_key: str,
+        search_parameters: Dict,
+        lattice_to_ctm_kwargs: Dict,
+        parallelize_conversion: bool,
+        rtf: int,
+        mem: int,
+        epochs: Optional[List[int]] = None,
+        needs_features_size: bool = True,
+        acoustic_mixture_path: Optional[tk.Path] = None,
+        nn_prior: bool = False,
```
vieting, line 40:
```suggestion
```

common/setups/rasr/onnx_precomputed_hybrid_system.py
```diff
@@ -0,0 +1,107 @@
...
+        pronunciation_scales: List[float],
+        lm_scales: List[float],
+        optimize_am_lm_scale: bool,
+        recognition_corpus_key: str,
+        feature_flow_key: str,
+        search_parameters: Dict,
+        lattice_to_ctm_kwargs: Dict,
+        parallelize_conversion: bool,
+        rtf: int,
+        mem: int,
```
vieting, line 36:
```suggestion
        mem: int,
        nn_prior: bool,
```

vieting:
@JackTemaki Your changes are addressed. Could you please check and possibly approve?

common/setups/rasr/onnx_precomputed_hybrid_system.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    System class for hybrid systems that train PyTorch models and export them to onnx for recognition. The NN
+    precomputed hybrid feature scorer is used.
+    """
+
+    def nn_recognition(
+        self,
+        name: str,
+        returnn_config: returnn.ReturnnConfig,
+        checkpoints: Dict[int, Union[returnn.Checkpoint, returnn.PtCheckpoint]],
```
JackTemaki, line 25: only PtCheckpoint I guess?

JackTemaki (approved):
Only the type fix remaining, otherwise good!

PR #156, opened 2023-07-20
ModelConfiguration serialization function

SimBe195:
This is my current logic for serializing `ModelConfiguration` objects which are used for configuration of models in [i6_models](https://github.com/rwth-i6/i6_models). There are two main components in this PR:

1. A `Call` serializer object that simply serializes the call of a callable with given args/kwargs.
2. A function `build_config_constructor_serializers` that returns such a `Call` object in order to perform a constructor call that re-constructs a given `ModelConfiguration` object. Along the way it also identifies and returns all necessary imports.

As such, no manual `Import` specifications are necessary for the model config and there is no conversion from/to dictionaries happening.

The implementation at the moment supports the following range of value-types inside the `ModelConfiguration`:

- Other `ModelConfiguration` objects (nesting)
- `ModuleFactoryV1` objects
- `torch.nn.Module` objects
- Functions (e.g. `torch.nn.functional.silu` or `sum`)
- Sisyphus `DelayedBase` objects
- Python primitives

The PR also includes a change in the `Import` serializer object which adds a `unhashed_package_root` parameter and is mostly copied from @JackTemaki.

When building the ReturnnConfig object inside sisyphus the logic might look roughly something like this:
```
serializer_objects = [...]

model_import = Import("i6_experiments.users.berger.pytorch.models.conformer_ctc.ConformerCTCModel")
serializer_objects.append(model_import)

model_config = ConformerCTCConfig(conformer_cfg=ConformerEncoderV1Config(num_layers=12, ...), ...)
call, imports = build_config_constructor_serializers(model_config, "cfg")
serializer_objects.extend([*imports, call])
serializer_objects.append(
    PyTorchModel(model_class_name=model_import.object_name, model_kwargs: {"cfg": CodeWrapper("cfg")}
)
...
returnn_config = ReturnnConfig(..., python_epilog=Collection(serializer_objects))
```
After serialization this results in the following code inside the `returnn.config` file:
```
...
from i6_experiments.users.berger.pytorch.models.conformer_ctc import ConformerCTCModel
from i6_experiments.users.berger.pytorch.models.conformer_ctc import ConformerCTCConfig
from i6_models.assemblies.conformer.conformer_v1 import ConformerEncoderV1Config
...
cfg = ConformerCTCConfig(
    conformer_cfg=ConformerEncoderV1Config(
        num_layers=12,
...
)
model_kwargs = {"cfg": cfg}

def get_model(epoch, step, **kwargs):
    return ConformerCTCModel(epoch=epoch, step=step, **model_kwargs, **kwargs)
```

### DIFF
--- a/common/setups/returnn_pytorch/serialization.py
+++ b/common/setups/returnn_pytorch/serialization.py
@@ -1,18 +1,22 @@
 from __future__ import annotations
-from typing import Any, Dict, List, Optional, Set, Union
+
 import os
 import pathlib
 import shutil
 import string
 import textwrap
+from dataclasses import fields
+from inspect import isfunction
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
-from sisyphus import gs, tk
-from sisyphus.hash import sis_hash_helper
-from sisyphus.delayed_ops import DelayedBase
-
+import torch
 from i6_core.util import instanciate_delayed
+from i6_models.config import ModelConfiguration, ModuleFactoryV1
+from sisyphus import gs, tk
+from sisyphus.delayed_ops import DelayedBase
+from sisyphus.hash import sis_hash_helper
 
-from ..serialization import SerializerObject
+from ..serialization import Call, Import, SerializerObject
 
 
 class PyTorchModel(SerializerObject):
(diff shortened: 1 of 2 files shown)


Atticus1806:
Would it be somehow possible to detect duplicate imports? E.g. for me now it seems `ModuleFactoryV1` will be imported quite a lot if this is used automatically.

JackTemaki:
@SimBe195 I would add the `Import` code together with the `PartialImport` that replaces the `PyTorchModel` in a separate PR today.

curufinwe (approved):
Looks good to me.

curufinwe:
Hash check fails because one of the packages has been updated and the old version is no longer present on the mirror. https://github.com/rwth-i6/i6_experiments/pull/175 fixes that.

common/setups/serialization.py
```diff
@@ -348,6 +349,62 @@ def _sis_hash(self):
...
+    Call(callable_name="range", args=[1, 10], kwargs=[("step", 2)], return_assign_variables="number_range")
+    ->
+    number_range = range(1, 10, step=2)
+    """
+
+    def __init__(
+        self,
+        callable_name: str,
+        args: Optional[List[Union[str, DelayedBase]]] = None,
+        kwargs: Optional[List[Tuple[str, Union[str, DelayedBase]]]] = None,
```
JackTemaki, line 366: I would allow only for kwargs, but add an extra parameter for non-hashed kwargs in case you want to have that.

common/setups/returnn_pytorch/serialization.py
```diff
@@ -1,18 +1,22 @@
...
 import shutil
 import string
 import textwrap
+from dataclasses import fields
+from inspect import isfunction
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
+import torch
+from i6_core.util import instanciate_delayed
+from i6_models.config import ModelConfiguration, ModuleFactoryV1
```
JackTemaki, line 14: ~~I would move this import locally into the build function, otherwise this will cause issues with existing setups.~~ Ah this does not work because it is needed for the typing... hmm maybe this works?
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from i6_models.config import ModelConfiguration, ModuleFactoryV1
```

And then import it again within the function.

JackTemaki (changes requested):

christophmluscher (approved):

common/setups/returnn_pytorch/serialization.py
curufinwe: No need to import again within the function. But what needs to be done is converting the type-annotations to strings, this way the concrete type is not required here.

JackTemaki: > No need to import again within the function
But if you do not do that, it will not exist, so then the `isinstance` checks will fail.

SimBe195: Moved import inside the function and annotated `cfg: Any` instead paired with an `assert isinstance(cfg, ModelConfiguration)` inside the function.

SimBe195:
> Would it be somehow possible to detect duplicate imports? E.g. for me now it seems `ModuleFactoryV1` will be imported quite a lot if this is used automatically.

Added some deduplication logic for the imports.

common/setups/returnn_pytorch/serialization.py
```diff
@@ -137,3 +141,112 @@ def _sis_hash(self) -> bytes:
             "delayed_objects": [obj for obj in self.serializer_objects if obj.use_for_hash],
         }
         return sis_hash_helper(h)
+
+
+def build_config_constructor_serializers(
+    cfg: Any, variable_name: Optional[str] = None, unhashed_package_root: Optional[str] = None
```
curufinwe, line 147:
```suggestion
    cfg: 'ModelConfiguration', variable_name: Optional[str] = None, unhashed_package_root: Optional[str] = None
```

JackTemaki (approved):
@SimBe195 if you say this all works as expected I am fine with merging

PR #157, opened 2023-07-31
add unhashed_package_root and PartialImport

JackTemaki:
`unhashed_package_root` allows to create "movable" configs that do not change their hash when moved to another location, which is problematic behavior of the current "Import" class.  Then, this also adds a `PartialImport` object which allows to import a callable with a set of pre-defined and hashed parameters. This can be used to for passing extra arguments to e.g. `get_model` or another hook function.

I do not like that `unhashed_package_root` is Optional, but I also do not see a way around it without breaking already existing things.


JackTemaki:
Prerequisite for #156

common/setups/serialization.py
```diff
@@ -71,17 +72,21 @@ class Import(SerializerObject):
...
         import_as: Optional[str] = None,
-        *,
         use_for_hash: bool = True,
         ignore_import_as_for_hash: bool = False,
     ):
         """
         :param code_object_path: e.g. `i6_experiments.users.username.my_rc_files.SomeNiceASRModel`.
             This can be the object itself, e.g. a function or a class. Then it will use __qualname__ and __module__.
+        :param unhashed_package_root: The root path to a package, from where relatives paths will be hashed.
+            Recommended is to use the root folder of an experiment module.
```
albertz, line 86: It would be good to have an example.

common/setups/serialization.py
```diff
@@ -71,17 +72,21 @@ class Import(SerializerObject):
...
         :param import_as: if given, the code object will be imported as this name
-        :param use_for_hash:
+        :param use_for_hash: if false, this module is not hashed when passed to a Collection/Serializer
```
albertz, line 88:
```suggestion
        :param use_for_hash: if false, this import is not hashed when passed to a Collection/Serializer
```

common/setups/serialization.py
```diff
@@ -71,17 +72,21 @@ class Import(SerializerObject):
...
+        :Param ignore_import_as_for_hash: do not hash `import_as` if set
```
albertz, line 89:
```suggestion
        :param ignore_import_as_for_hash: do not hash `import_as` if set
```

common/setups/serialization.py
```diff
@@ -112,6 +124,72 @@ def _sis_hash(self):
...
+
+    TEMPLATE = textwrap.dedent(
+        """\
+            import functools
+            kwargs = ${KWARGS}
+    
+            from ${IMPORT_PATH} import ${IMPORT_NAME} as _${IMPORT_NAME}
+            ${OBJECT_NAME} = functools.partial(_${IMPORT_NAME}, **kwargs)
+    
+        """
```
albertz, line 140: There are a lot of problems with this code. It exposes and leaks many symbols to the global namespace (`functools`, `kwargs`, `_${IMPORT_NAME}`). This could potentially break other code (admittedly unlikely, but still).

I just want to raise this. There are certainly easy ways to reduce this (without completely avoiding it) and there are also ways to completely avoid this (but making the code quite ugly). For example, you can use some combinations of `__import__` calls.

albertz: This should be possible via `__import__` (or `importlib` if you like that more).

common/setups/serialization.py
```diff
@@ -71,17 +72,21 @@ class Import(SerializerObject):
 
     def __init__(
         self,
+        *,
         code_object_path: Union[str, FunctionType, Any],
+        unhashed_package_root: Optional[str] = None,
```
albertz, line 77: I wonder if there maybe should be sth additionally like `extra_hash`, which the user can use to add explicitly to the hash, to make it unique and avoid possible collisions.
```suggestion
        unhashed_package_root: Optional[str] = None,
        extra_hash: Optional[str] = None,
```
This would be added in `_sis_hash` then.

common/setups/serialization.py
JackTemaki: What about:
```python
"""
            import functools
            ${OBJECT_NAME} = functools.partial(
                __import__("${IMPORT_PATH}", globals(), locals(), ["${IMPORT_NAME}"], 0).${IMPORT_NAME},
                **${KWARGS}
            )
"""
```
?

albertz: `functools` itself can also be done this way.
Also `globals`, `locals` and `level` don't need to be passed.
Also `fromlist` better via kwarg.
So:
```python
${OBJECT_NAME} = __import__("functools").partial(
    __import__("${IMPORT_PATH}", fromlist=["${IMPORT_NAME}"]).${IMPORT_NAME},
    **${KWARGS}
)
```

common/setups/serialization.py
albertz: In principle, yes, this would have the same effect. But I'm not sure people are aware of that, and I thought that such extra arg here makes it more visible. But maybe instead we can also hint to this in the docstring of `unhashed_package_root`.

JackTemaki:
Okay somehow a rebase did something wrong here, let me figure it out.

common/setups/serialization.py
JackTemaki: Opinions from other people?

JackTemaki:
Reviews please!

common/setups/serialization.py
```diff
@@ -72,16 +73,20 @@ class Import(SerializerObject):
...
         ignore_import_as_for_hash: bool = False,
     ):
         """
         :param code_object_path: e.g. `i6_experiments.users.username.my_rc_files.SomeNiceASRModel`.
             This can be the object itself, e.g. a function or a class. Then it will use __qualname__ and __module__.
+        :param unhashed_package_root: The root path to a package, from where relatives paths will be hashed.
+            Recommended is to use the root folder of an experiment module.
         :param import_as: if given, the code object will be imported as this name
-        :param use_for_hash:
+        :param use_for_hash: if false, this import is not hashed when passed to a Collection/Serializer
```
Atticus1806, line 88:
```suggestion
        :param use_for_hash: if False, this import is not hashed when passed to a Collection/Serializer
```

common/setups/serialization.py
```diff
@@ -112,6 +124,70 @@ def _sis_hash(self):
...
+        hashed_arguments: Dict[str, Any],
+        unhashed_arguments: Dict[str, Any],
+        import_as: Optional[str] = None,
+        use_for_hash: bool = True,
+        ignore_import_as_for_hash: bool = False,
+    ):
+        """
+        :param code_object_path: e.g. `i6_experiments.users.username.my_rc_files.SomeNiceASRModel`.
+            This can be the object itself, e.g. a function or a class. Then it will use __qualname__ and __module__.
+        :param unhashed_package_root: The root path to a package, from where relatives paths will be hashed.
```
Atticus1806, line 155: Maybe stay in the example? So e.g.: "For above example would be i6_experiments.users.username" or so

common/setups/serialization.py
```diff
@@ -112,6 +124,70 @@ def _sis_hash(self):
...
+            Recommended is to use the root folder of an experiment module.
+        :param hashed_arguments: argument dictionary for addition partial arguments to set to the callable.
+            Will be serialized as dict into the config, so make sure to use only serializable/parseable content
+        :param unhashed_arguments: same as above, but does not influence the hash
+        :param import_as: if given, the code object will be imported as this name
+        :param use_for_hash: if false, this module is not hashed when passed to a Collection/Serializer
```
Atticus1806, line 161:
```suggestion
        :param use_for_hash: if False, this module is not hashed when passed to a Collection/Serializer
```

Atticus1806 (approved):

common/setups/serialization.py
```diff
@@ -95,7 +102,14 @@ def __init__(
 
         self.object_name = self.code_object.split(".")[-1]
         self.module = ".".join(self.code_object.split(".")[:-1])
-        self.package = ".".join(self.code_object.split(".")[:-2])
```
michelwi, line 98: You are completely removing `self.package` here. In my setup I was using a [Collection](https://github.com/rwth-i6/i6_experiments/blob/8f65f607a9a159cfa0b98df2f13be889760d30a3/common/setups/returnn_pytorch/serialization.py#L69). The docstring said about the `package` argument:

https://github.com/rwth-i6/i6_experiments/blob/8f65f607a9a159cfa0b98df2f13be889760d30a3/common/setups/returnn_pytorch/serialization.py#L96

Which turns out was a lie. So instead I was giving the `some_import.package` of all my `Import`s. 

What is your advice to do instead? would giving `some_import.module` work? The code seems to expect to copy a folder and not a file. But anyway if I just need one module I would prefer to only copy one file..

common/setups/serialization.py
```diff
@@ -112,6 +126,72 @@ def _sis_hash(self):
         return sis_hash_helper(self.code_object)
 
 
+class PartialImport(Import):
+    """
+    Like Import, but for partial callables where certain parameters are given fixed and are hashed.
```
michelwi, line 131: > where certain parameters are given fixed 

Well.. this is not how `partial` works. Afaik the parameters that are given (and hashed) here are not fixed and can be overridden at runtime. Then we possibly have parameters influencing the hash that are not actually used or the hash changes without any different behavior (because the changed `hashed_arguments` from here are overwritten when actually calling the partial function/object.

Would this be a concern for us?

common/setups/serialization.py
michelwi: I don't think it would be necessary, but I have nothing against adding an argument that I will never use.

common/setups/serialization.py
JackTemaki: The code worked liked that in the beginning, but I found it to be a suboptimal solution. Thus I started copying folders instead, so that I could share some common code. Maybe we do not want that? 

> So instead I was giving the some_import.package of all my Imports.

This one I do not understand...

common/setups/serialization.py
JackTemaki: Does partial not thrown an error if a parameter is given both "fixed" and "dynamically"? This would indeed be a problem.

JackTemaki: Then I am out of ideas.

common/setups/serialization.py
JackTemaki: Would just keeping `self.package` be okay for you?

common/setups/serialization.py
albertz: I didn't fully followed your discussion, but what exactly is the problem actually? It makes sense to me that `partial` behaves this way.

It's also very easy to have things influencing the hash, which would have no effect. E.g. put some random stuff into a `ReturnnConfig`. You cannot really avoid that the user does that. I would say it's up to the user.

It's also very easy to have things not influencing the hash, which would have an effect. Again, up to the user.

The more I work with Sisyphus, the more the feeling I get that 90% of our work is just because we care about the hashing logic so much, and quite often, it's wrong in the end anyway. And in a lot of cases, we are not fixing things now, or not cleaning things up, because it would change the hash, and that leads to a lot of messy legacy code just because of that... (@mmz33 can tell a few stories...)

Anyway, I don't want to get off topic here. My point is just that we cannot avoid such problems, and I don't really see a problem here.

JackTemaki: > Anyway, I don't want to get off topic here.

I like going off-topic here for a second, this can be used for tomorrow :) 

> The more I work with Sisyphus, the more the feeling I get that 90% of our work is just because we care about the hashing logic so much, and quite often, it's wrong in the end anyway

This only began to happen when we started sharing code. In my first 4 years of Sisyphus, everybody using it in MT had their own recipes, shared only what was needed and modified it to their own ideas of how their individual hashing should work. Of course there were tons of errors, but then it were your own errors and you would know about that sooner or later.  Still, Sisyphus allowed us to do experiments no other group was able to do easily at that time (looking at Miguels iterative backtranslation), and I was able to re-run a full setup for JTP some years after creating it and deleting all the data, by just activating the Sisyphus manager once.

Now when we share things, and all own setups are created in a way that you can actually update your recipes with other peoples code, there is no good solution to fixing/handling errros. RETURNN suffers a little (definitely not as much as here) from the same problem, people used wrong behavior (e.g. Batchnorm, ZoneoutLSTM) and now you can not remove it easily because setups depend on that. For i6_core we took over a lot of years-old untested and possibly buggy code just because there was no time to write it all new.

Nowadays I wonder if hasheable/controlled reproducible and "perfectly" shareable pipelines are even desired, because the needed engineering work distracts so much from actual research. If you get your work done, and the code avoids the biggest errors, we should consider good enough (different viewpoints may apply for I6 and AppTek). 

Long story short: multiple people are already using this (or previous versions) of this code to run successful experiments, so the interest in discussing this PR is understandably limited. So I agree with Alberts statement: There is no direct problem here. I think any further "safeguards" for the user would take unnecessary time, effort and increase code complexity.

michelwi: -- Off Topic Rant --
I remember my own time at the start of using sisyphus when every time you updated the recipes there was a solid chance that the manager would tell you either "Error in line XXX" when some interface was broken or "Jobs Runable: 10k" when some hash in the beginning of your pipeline changed and all your jobs would now be rerun. That was unpleasant and lead to people not pulling frequently which led to even higher chances of everything breaking when you finally needed a new feature. I do not want to go back there.

Then we decided to preserve the old behavior. That was good, because you could update now without loosing halve a day debugging the changes. But it was bad because we carried around a lot of broken legacy behavior. Then we decided to once in a year do a hash breaking change where all bugs would be fixed in a branch and then once you start a new set of experiments you switch over.
Well, Markus started a new set of experiments that moth, everyone else did not. Then we had a diverged set of branches and nobody could do adaptation. That was also not the solution.

With the switch to i6_core most of the (hash + compatibility) problems were resolved. Now even we can collaborate and use each others setups and `import_work_directory` is actually usable. (Tried that recently with a setup from Mohammad)
I really prefer this way and would never like to go back to the old days.

-- On topic --
> but what exactly is the problem actually?

No problem, actually. Just from the doc string I had the impression that @JackTemaki had a misunderstanding how partial works, which was confirmed later in this thread. We have clarified this misunderstanding and with the full knowledge we can answer the question that I posed:

> Would this be a concern for us?

and "No, this is the users problem" is a very fine answer that I do in fact agree with.
I might not use this class, but others are very welcome to if they want.

-- Off Topic Rant 2 --
I am a bit burned with partials, as I spent a significant amount of a day debugging an annoying problem that involved partials in combination with sets as default arguments - and then trying to find a solution that is backwards compatible.

common/setups/serialization.py
michelwi: >Would just keeping self.package be okay for you?

This would work for me. 

Or I just copy the line 
```
package = ".".join(my_import.code_object.split(".")[:-2])
```
in my recipe collection if you think `self.package` should not me here.

common/setups/serialization.py
JackTemaki: > here was a solid chance that the manager would tell you
Oh yes this was horrible, but I compared only the extremes to each other, having independent recipes compared to fully sharing them. Sharing them without keeping behavior was a nightmare.

> Now even we can collaborate and use each others setups
Did the same with my Hiwis, huge benefit.

> I really prefer this way and would never like to go back to the old days.
I do as well, this is not what I wanted to say. I just wanted to say it can never be "perfect". You can not have all benefits of constantly updating your code and keeping it clean together with full cooperation and stability. So we have to live with hash bugs and keeping old code as long as we are still more happy with the overall solution ;) 

> I might not use this class, but others are very welcome to if they want.
So you prefer the very explicit `PyTorchModel` solution?

SimBe195 (approved):

common/setups/serialization.py
JackTemaki: I put a comment

michelwi (approved):

JackTemaki:
As @curufinwe wanted to move it later to i6_core, I will just merge it now. If there are changes requested we can do this as follow-up.

PR #160, opened 2023-08-10
fix ce loss opts

michelwi:
As discussed: the CE scale was overwritten nd therefore always set to 1.0


users/zeineldeen/experiments/conformer_att_2022/librispeech_960/seq_train_helpers.py
```diff
@@ -4,7 +4,7 @@
 def get_trans_lstm_lm():
     """
     Returns subnetwork of a LM trained only on transcripts. The LM architecture is close to ASR decoder architecture.
-    PPL = ?
+    PPL = 76 (train) ; 112 (dev clean+other)
```
michelwi, line 7: I took the liberty of filling out the `?` ;)

mmz33: Thanks.

mmz33 (approved):

PR #161, opened 2023-08-10
add local norm to ExternalLMDecoder

michelwi:
this adds the keys `"am_scale"` (default None) and `"local_norm"` (default False) to `ext_lm_opts`.

If `local_norm` is activated, it will locally norm over AM, LM and prior/iLM (if given)

### DIFF
--- a/users/zeineldeen/models/lm/external_lm_decoder.py
+++ b/users/zeineldeen/models/lm/external_lm_decoder.py
@@ -369,7 +369,7 @@ class ExternalLMDecoder:
       fusion_source += [ilm_decoder.output_prob_name]
 
     if self.ext_lm_opts.get("local_norm", False):
-      fusion_str = fusion_str + '- tf.math.reduce_logsumexp(' + fusion_str + ", axis=-1, keepdims=True)"
+      fusion_str = f"{fusion_str} - tf.math.reduce_logsumexp({fusion_str}, axis=-1, keepdims=True)"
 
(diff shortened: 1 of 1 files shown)


users/zeineldeen/models/lm/external_lm_decoder.py
```diff
@@ -365,6 +368,9 @@ def _create_external_lm_net(self) -> dict:
       fusion_str += ' - {} * safe_log(source(2))'.format(self.prior_lm_opts['scale'])
       fusion_source += [ilm_decoder.output_prob_name]
 
+    if self.ext_lm_opts.get("local_norm", False):
+      fusion_str = fusion_str + '- tf.math.reduce_logsumexp(' + fusion_str + ", axis=-1, keepdims=True)"
```
mmz33, line 372: maybe this can be simpler like this:
```
fusion_str = fusion_str + f" - tf.math.reduce_logsumexp({fusion_str}, axis=-1, keepdims=True)"
```
(there was also no space after `-`)

mmz33 (changes requested):

users/zeineldeen/models/lm/external_lm_decoder.py
michelwi: sure. or even
```suggestion
      fusion_str = f"{fusion_str} - tf.math.reduce_logsumexp({fusion_str}, axis=-1, keepdims=True)"
```

PR #163, opened 2023-08-15
fix missing prev: in miniLSTM att input

michelwi:
and also removing the whole `is_first_frame` logic.

Running now.. I'll report if this changes the WER.


mmz33:
Sorry i did black formatting so the change is not clear anymore. [here](https://github.com/rwth-i6/i6_experiments/pull/163/files#diff-02d18d8a4b982a52425630fe64d4058d68f1bde341cca02019ba6bae1f9b64e5R79) is the added if condition.

PR #164, opened 2023-08-15
generalize jobs for fairseq labeled data

AndreasPlt:
@vieting I noticed that the CreateFairseqLabeledDataJob was not well suited for using dedicated dev sets (dev-clean, dev-other in the case of LibriSpeech). I therefore changed some things and corrected some errors in the documentation as well. The changes were applied to all setups using that job (though it is not necessary, all existing setups should be able to work with the fixes as well).

### DIFF
--- a/users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/config_01_finetune.py
+++ b/users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/config_01_finetune.py
@@ -20,6 +20,25 @@ from recipe.i6_experiments.users.vieting.experiments.librispeech.librispeech_960
     import SetupFairseqJob
 
 
+def get_labels(
+    dest_name: str,
+    corpus_paths: Union[List[tk.Path], tk.Path],
+):
+    """
+    :param dest_name: name of the output file
+    :param corpus_paths: path to the corpora
+    """
+    if isinstance(corpus_paths, tk.Path):
+        corpus_paths = [corpus_paths]
+
+    label_data_job = CreateFairseqLabeledDataJob(
+        corpus_paths=corpus_paths,
+        dest_name=dest_name,
+    )
+    
+    return label_data_job.out_labels_path
+
+
 def get_task_dev_sampled(
     corpus_name: str,
     valid_percent: float = 0.01, 
@@ -141,23 +160,6 @@ def get_task_dev_separate(
     task = merge_job.out_task_path
     return task
 
-def get_labels(
-    dest_name: str,
-    corpus_paths: Union[List[tk.Path], tk.Path],
-):
-    """
-    :param dest_name: name of the output file
-    :param corpus_paths: path to the corpora
-    """
-    if isinstance(corpus_paths, tk.Path):
-        corpus_paths = [corpus_paths]
-
-    label_data_job = CreateFairseqLabeledDataJob(
-        corpus_paths=corpus_paths,
-        dest_name=dest_name,
-    )
-    
-    return label_data_job.out_labels_path
 
 def get_fairseq_root(fairseq_python_exe: Optional[tk.Path] = None):
     """
--- a/users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/config_01_finetune.py
+++ b/users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/config_01_finetune.py
@@ -271,34 +271,34 @@ def get_pretrained_model(model_path: Optional[Union[str, tk.Path]] = None):
 
 def main():
     # defines
-    PREFIX_NAME = "experiments/librispeech/librispeech_100_ctc/fairseq/"
-    EXP_NAME = "base"
-    NUM_GPUS = 1
-    GPU_MEM_RQMT = 24
-    CORPUS_NAME = "train-clean-100"
-    FAIRSEQ_PYTHON_EXE = tk.Path("/work/asr3/vieting/hiwis/pletschko/miniconda3/envs/fairseq_python38/bin/python")
-    USE_SAMPLED_DEV = False
+    prefix_name = "experiments/librispeech/librispeech_100_ctc/fairseq/"
+    exp_name = "base"
+    num_gpus = 1
+    gpu_mem_rqmt = 24
+    corpus_name = "train-clean-100"
+    fairseq_python_exe = tk.Path("/work/asr3/vieting/hiwis/pletschko/miniconda3/envs/fairseq_python38/bin/python")
+    use_sampled_dev = False
 
     # get pretrained model
     w2v_path = get_pretrained_model()
 
     # create wav2vec labeled data
-    if USE_SAMPLED_DEV:
+    if use_sampled_dev:
         dev_set = "dev"
-        labeled_data = get_task_dev_sampled(corpus_name=CORPUS_NAME, valid_percent=0.01)
+        labeled_data = get_task_dev_sampled(corpus_name=corpus_name, valid_percent=0.01)
     else:
         dev_set = "dev-other"
-        labeled_data = get_task_dev_separate(train_corpus_name=CORPUS_NAME, dev_corpus_name=dev_set)
+        labeled_data = get_task_dev_separate(train_corpus_name=corpus_name, dev_corpus_name=dev_set)
 
     # create fairseq config for finetuning
     fairseq_args = get_fairseq_args(
         labeled_data=labeled_data,
         w2v_path=w2v_path,
         dev_file_name=dev_set,
-        num_gpus=NUM_GPUS,
+        num_gpus=num_gpus,
     )
     fairseq_config = FairseqHydraConfig(fairseq_args)
-    fairseq_root = get_fairseq_root(fairseq_python_exe=FAIRSEQ_PYTHON_EXE)
+    fairseq_root = get_fairseq_root(fairseq_python_exe=fairseq_python_exe)
 
     # run finetuning
     job = FairseqHydraTrainingJob(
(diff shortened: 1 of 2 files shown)


vieting (changes requested):

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/config_01_finetune.py
```diff
@@ -43,8 +45,7 @@ def get_task(
     ), f"unknown corpus names: {corpus_names}"
 
     if valid_percent <= 0:
-        assert "dev-clean" in corpus_names or "dev-other" in corpus_names, \
-            "validation set is required if valid_percent <= 0"
+        logging.info("Not sampling validation set from training data. Create a separate dev set instead.")
```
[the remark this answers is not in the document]
AndreasPlt: Tried to fix that with the new rework of the job!

users/vieting/jobs/fairseq.py
```diff
@@ -15,55 +16,84 @@ class CreateFairseqLabeledDataJob(Job):
...
-        self.valid_percent = valid_percent
-        assert 0.0 <= self.valid_percent <= 1.0
+        self.valid_percent = sample_valid_percent
+        assert 0.0 <= self.valid_percent <= 1.0, "sample_valid_percent must be between 0 and 1."
         self.seed = seed
         self.path_must_contain = path_must_contain
+        self.create_letter_dict = create_letter_dict
+
+        if not valid_corpus_paths and self.valid_percent == 0:
+            logging.warning("No validation set given and sample_valid_percent is 0. No validation set will be created.")
```
vieting, line 83: Long line

vieting (changes requested):
I don't quite like the very specific handling of train and validation here. I think it would be more generic to have this job to convert bliss to the fairseq labeled data format (no matter if train or validation) and a separate job that copies multiple files together to create a specific combination of train and validation sets.

Also, there are changes to completely unrelated files. Please be careful with that.

AndreasPlt:
> I don't quite like the very specific handling of train and validation here. I think it would be more generic to have this job to convert bliss to the fairseq labeled data format (no matter if train or validation) and a separate job that copies multiple files together to create a specific combination of train and validation sets.

In this case, random sampling for validation data would not be possible anymore. Or should I then leave the random sampling as it is?

vieting:
> In this case, random sampling for validation data would not be possible anymore. Or should I then leave the random sampling as it is?

It is not possible within this job, but you can use the `ShuffleAndSplitSegmentsJob` to get train and validation segments and then the `FilterCorpusBySegmentsJob` to get the corresponding bliss corpora. If you use each of them as input to this job here, you get the respective sets.

users/vieting/jobs/fairseq.py
```diff
@@ -109,38 +103,95 @@ def create_tsv_and_labels(self):
...
+            labeled_data_paths = [labeled_data_paths]
+        self.labeled_data_paths = labeled_data_paths
+        assert all([isinstance(path, tk.Path) for path in self.labeled_data_paths])
+
+        self.create_letter_dict = create_letter_dict
+
+        self.out_task_path = self.output_path("task", directory=True)
+        self.out_dict_ltr_path = self.output_path("task/dict.ltr.txt")
+
+        self.rqmt = {"time": 6, "mem": 8, "cpu": 1}
```
vieting, line 165:
```suggestion
        self.rqmt = {"time": 1, "mem": 8, "cpu": 1}
```
One hour should be much more than enough I guess.

users/vieting/jobs/fairseq.py
```diff
@@ -109,38 +103,95 @@ def create_tsv_and_labels(self):
...
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt, mini_task=True)
```
vieting, line 168: If you use this as a mini task (which is probably a good idea), you don't need to specify `rqmt`. I think you can delete it completely because the job only uses minimal resources.

users/vieting/jobs/fairseq.py
```diff
@@ -176,23 +227,4 @@ def create_dict_ltr(self):
...
-            corpus_object.load(corpus_path.get())
-            for segment in corpus_object.segments():
-                audio_path = segment.recording.audio
-                assert os.path.exists(audio_path), f"Path {audio_path} does not exist."
-                if common_dir is None:
-                    common_dir = os.path.dirname(audio_path)
-                else:
-                    common_dir = os.path.commonpath([common_dir, os.path.dirname(audio_path)])
-        return common_dir
+            f.write(dict_ltr_content)
```
vieting, line 230: Please add an empty line at the file end.

users/vieting/jobs/fairseq.py
```diff
@@ -109,38 +103,95 @@ def create_tsv_and_labels(self):
...
+            for file in os.listdir(path_dir):
+                filename = os.fsdecode(file)
+                if filename.endswith(".tsv") or filename.endswith(".ltr") or filename.endswith(".wrd"):
+                    subprocess.run(
+                        [
+                            "cp",
+                            os.path.join(path.get(), filename),
+                            os.path.join(self.out_task_path, filename),
+                        ]
+                    )
```
vieting, line 190: `shutil.copyfile(src, dst)` seems to be more straight-forward.

users/vieting/jobs/fairseq.py
```diff
@@ -109,38 +103,95 @@ def create_tsv_and_labels(self):
...
+            self.create_dict_ltr()
+
+    def merge(self):
+        for path in self.labeled_data_paths:
+            assert os.path.exists(path.get()), f"Path {path} does not exist."
+
+            # iterature through directory "path" and copy all files to self.out_task_path directory
+            path_dir = os.fsencode(path.get())
+            for file in os.listdir(path_dir):
+                filename = os.fsdecode(file)
```
vieting, line 182: Why is this needed?

users/vieting/jobs/fairseq.py
```diff
@@ -109,38 +103,95 @@ def create_tsv_and_labels(self):
...
+        if self.create_letter_dict:
+            self.create_dict_ltr()
+
+    def merge(self):
+        for path in self.labeled_data_paths:
+            assert os.path.exists(path.get()), f"Path {path} does not exist."
+
+            # iterature through directory "path" and copy all files to self.out_task_path directory
+            path_dir = os.fsencode(path.get())
+            for file in os.listdir(path_dir):
```
vieting, line 181:
```suggestion
            for file in os.listdir(path.get()):
```
Is this not possible?

vieting (changes requested):

users/vieting/jobs/fairseq.py
AndreasPlt: Apparently, transforming a Python string to a byte string via `fsencode` is useful when working with invalid UTF-8 characters (if I understand correctly). However, assuming that we won't encounter any such characters, we could omit this `fsencode` and `fsdecode` process for easier readability, as suggested by you below.

users/vieting/jobs/fairseq.py
```diff
@@ -109,38 +103,95 @@ def create_tsv_and_labels(self):
...
+        for corpus_path in self.corpus_paths:
+            corpus_object = corpus.Corpus()
+            corpus_object.load(corpus_path.get())
+            for segment in corpus_object.segments():
+                audio_path = segment.recording.audio
+                assert os.path.exists(audio_path), f"Path {audio_path} does not exist."
+                if common_dir is None:
+                    common_dir = os.path.dirname(audio_path)
+                else:
+                    common_dir = os.path.commonpath([common_dir, os.path.dirname(audio_path)])
```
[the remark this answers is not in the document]
AndreasPlt: This should yield the same result, however requires probably more memory (since we would need to store every `audio_path` from the corpus, which might be quite many). I don't know how neglectable this effect is, especially when possibly using larger corpora later

users/vieting/jobs/fairseq.py
```diff
@@ -12,58 +14,45 @@
...
-    - train.tsv
-    - train.ltr
-    - train.wrd
-    - valid.tsv
-    - valid.ltr
-    - valid.wrd
-    - dict.ltr.txt
+    - dest.tsv
+    - dest.ltr
+    - dest.wrd
```
vieting, line 19:
```suggestion
    - <dest_name>.tsv
    - <dest_name>.ltr
    - <dest_name>.wrd
```

users/vieting/jobs/fairseq.py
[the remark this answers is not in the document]
vieting: I don't think it's too critical. You have to load each corpus individually anyway and just keep some string for each recording (which is a small fraction of the total memory). I think it's not realistic that the number of corpora would be so large that this would have an effect here.

users/vieting/jobs/fairseq.py
```diff
@@ -109,37 +88,86 @@ def create_tsv_and_labels(self):
...
+    def merge(self):
+        for path in self.labeled_data_paths:
+            assert os.path.exists(path.get()), f"Path {path} does not exist."
+
+            # iterature through directory "path" and copy all files to self.out_task_path directory
+            for file in os.listdir(path.get()):
+                if file.endswith(".tsv") or file.endswith(".ltr") or file.endswith(".wrd"):
+                    src = os.path.join(path.get(), file)
+                    dst = os.path.join(self.out_task_path.get(), file)
+                    shutil.copyfile(src, dst)
```
vieting, line 167: You might check that `dst` does not exist yet because it would be unexpected behavior if two input folders have files with the same name. I think it would be better to crash in that case than to ignore one of them silently.
```suggestion
                    assert not os.path.exists(dst), f"'{dst}' exists, two inputs have the same names"
                    shutil.copyfile(src, dst)
```

vieting (changes requested):

vieting (approved):

PR #166, opened 2023-08-22
added codecs (only ulaw) to audio perturbation

Max-Ryujin:
- added codecs, but only for ulaw 
- added an experiment with ulaw

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/experiments.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/experiments.py
@@ -599,6 +599,35 @@ def run_mel_audio_perturbation():
                     "preemphasis": "0.9_0.9_1.0",
                 },
             ),
+            "lgm80_conf-wei-oldspecaug-audio_perturbation_v2": dict(
+                returnn_args={
+                    "conformer_type": "wei",
+                    "specaug_old": {"max_feature": 8},
+                    "audio_perturbation": True,
+                    "extra_args": {
+                        "audio_perturb_args": {  # v2
+                            "speed": {"prob": 0.6, "minimum": 0.88, "maximum": 1.12},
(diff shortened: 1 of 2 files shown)


users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -599,6 +599,35 @@ def run_mel_audio_perturbation():
...
+                    "specaug_old": {"max_feature": 8},
+                    "audio_perturbation": True,
+                    "extra_args": {
+                        "audio_perturb_args": {  # v2
+                            "speed": {"prob": 0.6, "minimum": 0.88, "maximum": 1.12},
+                            "tempo": {"prob": 0.6, "minimum": 0.83, "maximum": 1.17},
+                            "preemphasis": {"prob": 0.9, "minimum": 0.9, "maximum": 1.0},
+                            "codecs": [
+                            {"format": "wav", "encoding": "ULAW", "prob": 0.4},
+                            ],
```
vieting, line 614:
```suggestion
                            "codecs": [{"format": "wav", "encoding": "ULAW", "prob": 0.4}],
```

vieting: Otherwise, use additional indentation.

vieting: Additionally, `"format"` is not used by your implementation. I think I had it in the example before because it was required for applying effects in `torchaudio`, but now it should be removed I guess.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/experiments.py
Max-Ryujin: Done

vieting (approved):

PR #168, opened 2023-08-27
implement mmi training

michelwi:
Implementation of MMI sequence training type.

### DIFF
--- a/users/zeineldeen/experiments/conformer_att_2022/librispeech_960/seq_train_helpers.py
+++ b/users/zeineldeen/experiments/conformer_att_2022/librispeech_960/seq_train_helpers.py
@@ -168,3 +168,79 @@ def add_min_wer(net, loss_scale, ce_scale, am_scale, lm_scale, beam_size, ce_lab
         },
         "target": "bpe_labels",
     }
+
+
+def add_mmi(net, loss_scale, ce_scale, am_scale, lm_scale, beam_size, ce_label_smoothing=0.0):
+    """
+    Add Maximum Mutual Information training criterion.
+    """
+    assert ce_label_smoothing==0.0, "label smoothing not implemented correctly"
+    subnet = net["output"]["unit"]
+
+    # add LM
+    subnet["lm_output"] = get_trans_lstm_lm()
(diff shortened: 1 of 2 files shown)


users/zeineldeen/experiments/conformer_att_2022/librispeech_960/seq_train_helpers.py
```diff
@@ -168,3 +168,79 @@ def add_min_wer(net, loss_scale, ce_scale, am_scale, lm_scale, beam_size, ce_lab
...
+    }
+    extra_search_subnet["output"]["from"] = "combined_log_prob"  # asr + lm
+    extra_search_subnet["output"]["length_normalization"] = False
+    extra_search_subnet["output"]["cheating"] = "exclusive"
+    extra_search_subnet["output"]["beam_size"] = beam_size
+    extra_search_subnet["output"]["input_type"] = "log_prob"
+
+    # CE loss
+    if "loss_opts" in subnet["output_prob"]:
+        subnet["output_prob"]["loss_opts"]["scale"] = loss_scale + ce_scale
```
mmz33, line 230: so this is correct because the AM part of the MMI numerator is same as CE loss right?

mmz33 (approved):

users/zeineldeen/experiments/conformer_att_2022/librispeech_960/seq_train_helpers.py
michelwi: Correct, cf https://arxiv.org/abs/2005.10049 Equation 7.
The `loss_scale` is for the numerator part (with the \alpha)
The `ce_scale` is if some additional cross entropy loss should be added.

Note that because of this being part of the numerator, I require `ce_label_smoothing==0.0`, as this would otherwise interfere with MMI loss. we could duplicate this output layer and then have the MMI numerator without label smoothing and the CE with optional label smoothing.

michelwi:
I mark this ready for review. I tried the code and got reasonable WERs after training.

mmz33:
> I mark this ready for review. I tried the code and got reasonable WERs after training.

Good to know.

PR #169, opened 2023-09-05
Added scf trainings to the ctc experiments 

Max-Ryujin:
- applied black to experiments.py
- renamed `run_mel_nn_args` to better reflect the new code
- added `run_scf_baseline` which is identical to `run_mel_baseline` except
-  `feature_args = {"class": "ScfNetwork", "size_tf": 256 // 2, "stride_tf": 10 // 2}` which is based on the hybrid implementation of SCF
- Epochs and batch size are also based on the hybrid implementation of SCF


vieting:
What about this PR? Is it still needed?

Max-Ryujin:
This PR is outdated and no longer needed. It got replaced by `https://github.com/rwth-i6/i6_experiments/pull/176`

PR #170, opened 2023-09-05
Added SCF trainings to the CTC experiments 

Max-Ryujin:
- renamed `run_mel_nn_args` to `run_nn_args`
- added `run_scf_baseline` which is identical to `run_mel_baseline` except `feature_args = {"class": "ScfNetwork", "size_tf": 256 // 2, "stride_tf": 10 // 2}`
- batch size is kept at 10k but split into 2x5k, because of memory limits


vieting (approved):
Looks good.

PR #172, opened 2023-09-13
Split nn.py

Atticus1806:
Prequesite for #151 

Refactoring was done by @christophmluscher


common/setups/rasr/util/nn/decode.py
```diff
@@ -0,0 +1,88 @@
...
+    prior_scales: List[float]
+    pronunciation_scales: List[float]
+    returnn_config: Optional[returnn.ReturnnConfig]
+    rtf: int
+    search_parameters: Optional[SearchParameters]
+    use_gpu: Optional[bool]
+
+
+@dataclass()
+class NnRecogArgs:
```
JackTemaki, line 65: The same class appearing twice? This can not be correct. I assume we can just delete the first version then?

JackTemaki (changes requested):

common/setups/rasr/util/nn/decode.py
Atticus1806: This is marked Todo by @christophmluscher so maybe he can give more information on the thoughts here / how to proceed

christophmluscher:
I created my own PR. Syncing with @Atticus1806 PR was annoying...

PR #173, opened 2023-09-13
Add alignment alias and step, more state tying dumping

Atticus1806:
Prequesite for #151 

Adds an alias to the dump alignment jobs, adds a new method for forced align and state tying dump for the final state tying.


common/setups/rasr/gmm_system.py
```diff
@@ -1237,6 +1244,23 @@ def run_vtln_sat_step(self, step_args):
                     **step_args.sdm_args,
                 )
 
+    def run_forced_align_step(self, step_args):
```
JackTemaki, line 1247: I wonder if it would be cleaner to add this as `_v2` and the old one as `run_forced_align_step`. Otherwise it is extremely unclear why it does not match.

christophmluscher:
I will replace this with another PR.

PR #174, opened 2023-09-13
Refactoring of Hybrid and Rasr System

Atticus1806:
Prerequisite for #151, has #172 as prerequisite

Introduces some typing updates as well as and update to training

Again mainly taken over from @christophmluscher


christophmluscher:
I am breaking this down into smaller commits.

PR #175, opened 2023-09-18
Perform update before installing libsndfile in hash check action

curufinwe:
Version stored in apt cache can be out of date and no longer present in the mirror, causing for example https://github.com/rwth-i6/i6_experiments/actions/runs/6219531486/job/16877777086?pr=156 to fail.


curufinwe:
It seems the hash check is anyway broken because there is some invalid python code present. I will not debug this further as it's not related to this PR.

Atticus1806 (approved):

JackTemaki (approved):

PR #179, opened 2023-10-04
Adds forced alignment for GMM system

christophmluscher:


common/setups/rasr/gmm_system.py
```diff
@@ -1237,6 +1237,23 @@ def run_vtln_sat_step(self, step_args):
                     **step_args.sdm_args,
                 )
 
+    def run_forced_align_step(self, step_args):
+        train_corpus_keys = step_args.pop("train_corpus_keys", self.train_corpora)
```
Atticus1806, line 1241: This will break old setups right? Since it is renamed from `corpus_keys` to `train_corpus_keys`. This is why it was not included in the other PR like this.

Atticus1806 (changes requested):

common/setups/rasr/gmm_system.py
christophmluscher: what do you mean break? the hash should not be influenced or is it?

Atticus1806: further down you call `self.run_forced_align_step(step_args)` and just pass the args. But in the old code the args should contain: `step_args.pop("corpus_keys", self.train_corpora)`, but in the new code that key is called `"train_corpus_keys"`

Atticus1806: unless I am missing something I think old code using that function would now not work anymore.

common/setups/rasr/gmm_system.py
```diff
@@ -1390,12 +1407,7 @@ def run(self, steps: Union[List[str], RasrSteps]):
 
             # ---------- Forced Alignment ----------
             if step_name.startswith("forced_align"):
-                corpus_keys = step_args.pop("corpus_keys", self.train_corpora)
-                for corpus in corpus_keys:
-                    self.forced_align(
-                        feature_scorer_corpus_key=corpus,
-                        **step_args,
-                    )
+                self.run_forced_align_step(step_args)
```
Atticus1806, line 1410:
```suggestion
                corpus_keys = step_args.pop("corpus_keys", None)
                assert "corpus_keys" not in step_args.keys() or "train_corpus_keys" not in step_args.keys(), "Please define either corpus_keys or train_corpus_keys, but not both."
                if corpus_keys:
                    step_args["train_corpus_keys"] = corpus_keys
                self.run_forced_align_step(step_args)
```

Atticus1806 (approved):

PR #180, opened 2023-10-04
PyTorch ONNX Hybrid System Class

christophmluscher:
IIRC, this class was added by @vieting and extended by @Atticus1806 ?!

Since there are dependencies on user stuff and there might be a few open questions I ll add this for now as a draft PR so that people are aware that this exists..

### DIFF
--- a/common/setups/rasr/pytorch_onnx_hybrid_system.py
+++ b/common/setups/rasr/pytorch_onnx_hybrid_system.py
@@ -1,0 +1,182 @@
+import copy
+import itertools
+from typing import Dict, List, Optional, Tuple, Union
+
+# -------------------- Sisyphus --------------------
+
+from sisyphus import tk
+
+# -------------------- Recipes --------------------
+
+import i6_core.rasr as rasr
+import i6_core.returnn as returnn
+
+from i6_core.returnn.flow import (
+    make_precomputed_hybrid_tf_feature_flow,
+    add_tf_flow_to_base_flow,
+)
+
+
+# -------------------- Init --------------------
+
+Path = tk.setup_path(__package__)
+from i6_core.mm import CreateDummyMixturesJob
+
+from i6_experiments.common.setups.rasr.hybrid_system import HybridSystem
+
+from i6_experiments.users.hilmes.tools.onnx import ExportPyTorchModelToOnnxJob
+
+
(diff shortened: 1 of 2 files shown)


common/setups/rasr/pytorch_onnx_hybrid_system.py
```diff
@@ -0,0 +1,157 @@
...
+# -------------------- Init --------------------
+
+Path = tk.setup_path(__package__)
+
+from i6_core.mm import CreateDummyMixturesJob
+from i6_core.rasr import OnnxFeatureScorer
+
+from i6_experiments.common.setups.rasr.hybrid_system import HybridSystem
+
+from i6_experiments.users.hilmes.tools.onnx import ExportPyTorchModelToOnnxJob
```
JackTemaki, line 29: It is not allowed to reference user code in common, also this way of export is deprecated. This should better be discussed...

christophmluscher: agreed :D

> Since there are dependencies on user stuff and there might be a few open questions I ll add this for now as a draft PR so that people are aware that this exists..

:D

JackTemaki: reading would help :)

Atticus1806:
I don't think this makes sense to merge, the code in my userfolder regarding that part is highly experimental, not fully pushed and far from clean. 
If we want to merge this, the jobs should be move to `i6_core` and then merged there as general jobs. Since as @JackTemaki says this is deprecated anyways I am not sure if all this effort is worth the time.

vieting:
Yes, I also think we should add all necessary jobs to i6_core and then use them here. What is the current status on that?

JackTemaki:
> Yes, I also think we should add all necessary jobs to i6_core and then use them here. What is the current status on that?

I am not in favor of having two conflicting solutions to the same problem...

We already have the "precomputed" system as reference, so if someone wants the system with the feature scorer instead he/she can re-write the code here using the jobs that are already part of `i6_core` and make a PR for that.

vieting:
> We already have the "precomputed" system as reference

Oh right. I didn't see that this was about the feature scorer. Then I also never used this.

Atticus1806:
After offline discussion.

PR #181, opened 2023-10-04
Split `nn.py` file

christophmluscher:
THe file was getting quite large, split into several files. Imports should still work


Atticus1806 (changes requested):
There are no `NnRecogArgs` anymore? This will potentially break setups.

JackTemaki:
> There are no `NnRecogArgs` anymore? This will potentially break setups.

But they are here: https://github.com/rwth-i6/i6_experiments/pull/181/files#diff-2c01418d940f8adb8ebc1306dbe6881188bf8bcb6e15a50cc70a0c9df64e3145R41

or not?

Atticus1806:
Huh, seems like I overlooked that file for some reason.

JackTemaki (approved):
I can say my existing setups do not break, for everything else @Atticus1806 you have to decide if it is to your liking.

Atticus1806 (approved):

JackTemaki:
I really hope we can have a reference Hybrid pipeline that is included in the (hash-)tests soon.

PR #191, opened 2023-10-26
Datastream helpers

JackTemaki:
This is the last remaining component of my usual setups which is not pushed yet. Some parts are maybe not too relevant, now that we are often not doing feature extraction with librosa but mostly in the graph. Still, it can be used to compute features statistics that exactly match the ones computed by the PyTorch feature extraction. 

It also makes it transparent in the recipes what features we actually support in RETURNN. 

Mostly relevant for @Atticus1806 but I marked other that also could find this interesting.

### DIFF
--- a/common/setups/returnn/datastreams/audio.py
+++ b/common/setups/returnn/datastreams/audio.py
@@ -8,8 +8,8 @@ from typing import *
 from i6_core.returnn import ReturnnConfig
 from i6_core.returnn.dataset import ExtractDatasetMeanStddevJob
 
-from i6_experiments.users.rossenbach.common_setups.returnn.datastreams.base import Datastream
-from i6_experiments.users.rossenbach.common_setups.returnn.datasets.audio import OggZipDataset
+from .base import Datastream
+from ..datasets.audio import OggZipDataset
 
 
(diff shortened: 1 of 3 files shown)


common/setups/returnn/datastreams/audio.py
```diff
@@ -0,0 +1,255 @@
...
+        :param kwargs: additional options that are passed manually
+        """
+        super().__init__(available_for_inference)
+        self.options = options
+        self.additional_options = kwargs.copy()
+
+        if options.features not in KNOWN_FEATURES:
+            print("Warning: %s is not a known feature type" % options.features)
+
+        if type(options.feature_options) not in KNOWN_FEATURES.get(options.features, [type(None)]):
```
Atticus1806, line 141: Since all types include `None` we can move the check down here and have `KNOWN_FEATURES` a bit cleaner.
If in the future we have features where this is not the case we can change this.

Atticus1806 (approved):
Looks good to me. Would change the one comment but since this a bit personal flavor we can keep as is, up to you.

mmz33 (approved):

PR #193, opened 2023-12-11
Add config for LM lookahead w/ 4gram

NeoLegends:
The existing LM configuration is already okay, I think the only missing parts are the LM lookahead w/ 4gram for the LFR models.

@Marvin84 Could you generate me the decoding (RASR) configs with your setups for this change, so I can diff them and see if something is missing?

### DIFF
--- a/users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
+++ b/users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
@@ -931,7 +931,7 @@ class BASEFactoredHybridDecoder:
         if search_parameters.lm_lookahead_scale is not None:
             # Use 4gram for lookahead. The lookahead LM must not be too good.
             #
-            # Half the normal LM scale is a good value here.
+            # Half the normal LM scale is a good starting value.
 
             # To validate the assumption the original LM is a 4gram
             assert original_lm_config.language_model_config.type.lower() == "arpa"
--- a/users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
+++ b/users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
@@ -933,6 +933,9 @@ class BASEFactoredHybridDecoder:
             #
             # Half the normal LM scale is a good value here.
 
+            # To validate the assumption the original LM is a 4gram
+            assert original_lm_config.language_model_config.type.lower() == "arpa"
+
             adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.separate_lookahead_lm = True
             adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.lm_lookahead.lm_lookahead_scale = (
                 search_parameters.lm_lookahead_scale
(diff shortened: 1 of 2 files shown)


users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
```diff
@@ -917,13 +919,30 @@ def recognize(
...
+            adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.separate_lookahead_lm = True
+            adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.lm_lookahead.lm_lookahead_scale = (
+                search_parameters.lm_lookahead_scale
+            )
+            adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.lookahead_lm.file = (
+                original_lm_config.language_model_config.file
+            )
+            # TODO(future): Add LM image instead of file here.
+            adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.lookahead_lm.scale = 1.0
+            adv_search_extra_config.flf_lattice_tool.network.recognizer.recognizer.lookahead_lm.type = "ARPA"
```
Marvin84, line 948: So you confirm that the "real" lm lookahead scale in this case is under `lm_lookahead.lookahead_scale` and not here?

Marvin84 (approved):
I will first merge this, test it, and in case there is a problem re-open the PR

NeoLegends:
You could just have left the PR open and checked out the branch?

users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
NeoLegends: I believe the lm_lookahead_scale scales the lookahead during decoding, using LM scores scaled by the `lookahead_lm.scale`. I'm not entirely sure with that though, but from testing I discovered the scales do not have the same effect. For 4gram lookahead this configuration works well.

users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
```diff
@@ -917,13 +919,33 @@ def recognize(
...
             )
-        else:
-            adv_search_extra_config = None
+        if search_parameters.lm_lookahead_scale is not None:
+            # Use 4gram for lookahead. The lookahead LM must not be too good.
+            #
+            # Half the normal LM scale is a good starting value.
+
+            # To validate the assumption the original LM is a 4gram
+            assert original_lm_config.language_model_config.type.lower() == "arpa"
```
Marvin84, line 937: @NeoLegends JFYI, here you should do: ` assert original_lm_config.type.lower() == "arpa"` instead

users/raissi/setups/common/decoder/BASE_factored_hybrid_search.py
curufinwe: I don't know by heart, this is unfortunately quite complicated due to historic reasons. I'd need to look into this. Maybe @SimBe195 remembers.

SimBe195: Yeah there was some very confusing logic with these scales. I looked at this a while ago but can't remember the details. I believe this was even handled a bit inconsistently in the code, like sometimes the `lookahead-lm.scale` and `lm-lookahead-scale` would be multiplied together and sometimes only the lm-lookahead-scale would be applied. It gets even more confusing when you have a `CombineLM` with sub-lm's that also have their own scale in addition.

PR #195, opened 2023-12-12
Perturbations: non_linearity and fix source code export

Max-Ryujin:
- added code for nonlinear perturbation
- fixed `get_code_for_perturbation` to correctly return the entire code 
- changed tempo to stretch for slightly better performance 
- black formatting

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -139,7 +139,7 @@ class WaveformPerturbation:
 
         if random_state.random() < factor.prob:
             alpha = random_state.random() * (factor.max - factor.min) + factor.min
-            audio = np.sign(audio) * np.abs(audio) ** (alpha)
+            audio = np.sign(audio) * np.abs(audio) ** alpha
             audio = audio.astype(np.float32)
         return audio
 
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -1,4 +1,5 @@
 from typing import List, Dict, Optional, Any
+import inspect
 
 
 class PerturbationFactor:
@@ -27,6 +28,7 @@ class WaveformPerturbation:
         sox_effects: Optional[List[List[str]]] = None,
         codecs: Optional[List[Dict[str, Any]]] = None,
         preemphasis: Optional[Dict[str, Any]] = None,
+        non_linearities: Optional[Dict[str, Any]] = None,
     ):
         """
         Initializes an instance of a class.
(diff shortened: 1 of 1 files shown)


users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -122,12 +131,25 @@ def apply_codecs(audio, sample_rate, random_state, codecs):
...
 
 def get_code_for_perturbation():
-    classes = ["from typing import List, Dict, Any, Optional"]
-    for cls_name, cls in list(globals().items()):
-        if isinstance(cls, type):
-            classes.append(cls)
-    return classes
+    class_sources = ["from typing import List, Dict, Any, Optional"]
+    for name, obj in globals().items():
+        if inspect.isclass(obj) and obj.__module__ == __name__:
```
vieting, line 151: Why do you need to change this?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -122,12 +131,25 @@ def apply_codecs(audio, sample_rate, random_state, codecs):
                     tfm.set_output_format(encoding="u-law")
                 else:
                     raise NotImplementedError(f"Codec {codec} not implemented.")
+        return tfm.build_array(input_array=audio, sample_rate_in=sample_rate)
```
vieting, line 134: Was that not implemented before?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -88,7 +97,7 @@ def sox(self, audio, sample_rate, random_state):
             speed = True
         if self._tempo is not None and random_state.random() < self._tempo.prob and not speed:
             factor = random_state.random() * (self._tempo.max - self._tempo.min) + self._tempo.min
-            tfm.tempo(factor)
+            tfm.stretch(factor)
```
vieting, line 100: Did you verify this does exactly the same?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -55,16 +57,23 @@ def __init__(
             - 'minimum' (float): The minimum preemphasis factor.
             - 'maximum' (float): The maximum preemphasis factor.
             Example: {"prob": 0.9, "minimum": 0.9, "maximum": 1.0}
+        :param non_linearity: A dictionary containing parameters for the non-linearity filter.
+            - 'prob' (float): The probability of applying the non-linearity filter.
+            - 'alpha' (float): The alpha value for the non-linearity filter.
```
vieting, line 62: Like for other perturbations, please define alpha via a minimum and maximum and sample it randomly from that range. To have a fixed alpha, you can then still set both minimum and maximum to that value.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: yes, it does. the only difference is small performance gains depending on the factor value

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: No, it was not.

vieting: As I mentioned earlier, it is generally desirable to change only things that are related to the PR (so here as per your title the non linearity and fixed source code export). In this case, I'm okay with keeping it in here, but please keep this in mind for the future.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -122,6 +131,16 @@ def apply_codecs(audio, sample_rate, random_state, codecs):
...
                     raise NotImplementedError(f"Codec {codec} not implemented.")
+        return tfm.build_array(input_array=audio, sample_rate_in=sample_rate)
+
+    @staticmethod
+    def non_linearity_function(audio, sample_rate, random_state, factor):
+        import numpy as np
+
+        if random_state.random() < factor.prob:
+            alpha = random_state.random() * (factor.max - factor.min) + factor.min
+            audio = np.sign(audio) * np.abs(audio) ** (alpha)
```
vieting, line 142:
```suggestion
            audio = np.sign(audio) * np.abs(audio) ** alpha
```

vieting (approved):

PR #202, opened 2024-02-07
Implemented ulaw for audio perturbation

Max-Ryujin:
The implementation follows the standard u-law encoding and sounds identical to the torchaudio implementation. 
The values however are not identical to the torchaudio output. When normalized. the values differ up to 0.02.

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -73,7 +73,9 @@ class WaveformPerturbation:
         if codecs:
             self._perturbations.append(functools.partial(self.apply_codecs, codecs=codecs))
         if non_linearity:
-            self._perturbations.append(functools.partial(self.non_linearity_function, factor=PerturbationFactor(**non_linearity)))
+            self._perturbations.append(
+                functools.partial(self.non_linearity_function, factor=PerturbationFactor(**non_linearity))
+            )
 
     def run(self, audio, sample_rate, random_state):
         import numpy as np
@@ -126,8 +128,8 @@ class WaveformPerturbation:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    #standard values for µ-law encoding
-                    quantization_bits=8
+                    # standard values for µ-law encoding
+                    quantization_bits = 8
                     MU = 255.0
 
                     # ensure audio is normalised
@@ -138,9 +140,11 @@ class WaveformPerturbation:
                     encoded_audio = np.sign(normalized_audio) * np.log1p(MU * np.abs(normalized_audio)) / np.log1p(MU)
                     encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
 
-                    #normalise encoded audio
-                    encoded_normalized_audio = (encoded_audio - encoded_audio.min()) / (encoded_audio.max() - encoded_audio.min())
-    
+                    # normalise encoded audio
+                    encoded_normalized_audio = (encoded_audio - encoded_audio.min()) / (
+                        encoded_audio.max() - encoded_audio.min()
+                    )
+
                     output_audio = encoded_normalized_audio
                 else:
                     raise NotImplementedError(f"Codec {codec} not implemented.")
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -130,7 +130,7 @@ class WaveformPerturbation:
                 if codec.get("encoding") == "ULAW":
                     # standard values for µ-law encoding
                     quantization_bits = 8
-                    MU = 255.0
+                    mu = 255.0
 
                     # ensure audio is normalised
                     max_amplitude = np.max(np.abs(output_audio))
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -123,7 +123,6 @@ class WaveformPerturbation:
 
     @staticmethod
     def apply_codecs(audio, sample_rate, random_state, codecs):
-        output_audio = audio
         for codec in codecs:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
@@ -132,12 +131,8 @@ class WaveformPerturbation:
                     quantization_bits = 8
                     mu = 255.0
 
-                    # ensure audio is normalised
-                    max_amplitude = np.max(np.abs(output_audio))
-                    normalized_audio = audio / max_amplitude
-
-                    # µ-law encoding formula
-                    encoded_audio = np.sign(normalized_audio) * np.log1p(MU * np.abs(normalized_audio)) / np.log1p(MU)
+                    # mu-law encoding formula
+                    encoded_audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
                     encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
 
                     # normalize encoded audio
@@ -145,7 +140,7 @@ class WaveformPerturbation:
                         encoded_audio.max() - encoded_audio.min()
                     )
 
-                    output_audio = encoded_normalized_audio
+                    audio = encoded_normalized_audio
                 else:
                     raise NotImplementedError(f"Codec {codec} not implemented.")
         return output_audio
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -127,7 +127,7 @@ class WaveformPerturbation:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    # check if youdio is in the right range for mu-law encoding
+                    # check if audio is in the right range for mu-law encoding
                     if np.max(np.abs(audio)) > 1.0:
                         raise ValueError("Audio must be in the range [-1, 1] for mu-law encoding.")
                     # standard value for mu-law encoding
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -140,7 +140,7 @@ class WaveformPerturbation:
                     encoded_audio = np.sign(normalized_audio) * np.log1p(MU * np.abs(normalized_audio)) / np.log1p(MU)
                     encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
 
-                    # normalise encoded audio
+                    # normalize encoded audio
                     encoded_normalized_audio = (encoded_audio - encoded_audio.min()) / (
                         encoded_audio.max() - encoded_audio.min()
                     )
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
@@ -130,13 +130,10 @@ class WaveformPerturbation:
                     # check if youdio is in the right range for mu-law encoding
                     if np.max(np.abs(audio)) > 1.0:
                         raise ValueError("Audio must be in the range [-1, 1] for mu-law encoding.")
-
                     # standard value for mu-law encoding
                     mu = 255.0
-
                     # mu-law encoding formula
-                    encoded_audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
-                    audio = encoded_audio
+                    audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
                 else:
                     raise NotImplementedError(f"Codec {codec} not implemented.")
         return audio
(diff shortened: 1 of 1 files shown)


Max-Ryujin:
Mu and quantization_bits are currently hardcoded into the code. If necessary I could add them to the codecs dictionary.

vieting:
> When normalized. the values differ up to 0.02.

How exactly normalized? What is min and max in that case?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
 
     @staticmethod
     def apply_codecs(audio, sample_rate, random_state, codecs):
-        import sox
-
-        tfm = sox.Transformer()
+        output_audio = audio
```
vieting, line 126: Why introducing `output_audio`? It seems like you could just keep using `audio` instead.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
         for codec in codecs:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # standard values for µ-law encoding
+                    quantization_bits = 8
+                    MU = 255.0
```
vieting, line 133:
```suggestion
                    mu = 255.0
```
Please also change below.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+
+                    # ensure audio is normalised
```
vieting, line 135:
```suggestion
                    # ensure audio is normalized
```
We usually use US spelling.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+                    max_amplitude = np.max(np.abs(output_audio))
+                    normalized_audio = audio / max_amplitude
+
+                    # µ-law encoding formula
```
vieting, line 139:
```suggestion
                    # mu-law encoding formula
```
Probably fine, but sometimes special characters can break things, so I'd rather be on the safe side.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # standard values for µ-law encoding
+                    quantization_bits = 8
+                    MU = 255.0
+
+                    # ensure audio is normalised
+                    max_amplitude = np.max(np.abs(output_audio))
+                    normalized_audio = audio / max_amplitude
```
vieting, line 137: Is this typically done as part of mu-law encoding? Otherwise, this could lead to unexpected behavior.

Max-Ryujin:
`encoded_normalized_audio = (encoded_audio - encoded_audio.min()) / (encoded_audio.max() - encoded_audio.min())`
Min and max are the lowest and highest values after encoding. I moved the value range from 0-255 to 0-1

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+
+                    # µ-law encoding formula
+                    encoded_audio = np.sign(normalized_audio) * np.log1p(MU * np.abs(normalized_audio)) / np.log1p(MU)
+                    encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
+
+                    # normalise encoded audio
+                    encoded_normalized_audio = (encoded_audio - encoded_audio.min()) / (
+                        encoded_audio.max() - encoded_audio.min()
+                    )
```
vieting, line 146: Same here, is it always supposed to be like this?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,32 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+
+                    # ensure audio is normalised
+                    max_amplitude = np.max(np.abs(output_audio))
+                    normalized_audio = audio / max_amplitude
+
+                    # µ-law encoding formula
+                    encoded_audio = np.sign(normalized_audio) * np.log1p(MU * np.abs(normalized_audio)) / np.log1p(MU)
+                    encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
+
+                    # normalise encoded audio
```
vieting, line 143:
```suggestion
                    # normalize encoded audio
```

vieting (changes requested):

vieting:
> I moved the value range from 0-255 to 0-1.

That's probably not something that someone would expect when using this codec.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: This is indeed not necessary and removing it also fixes the small numerical difference to torchaudio

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: I added this. I think it is necessary to put the perturbated audio in the same value range as the unperturbed audio to allow for the training of the features. Two different value ranges in the same training data would be hard to learn

Max-Ryujin:
It is not really possible to move this outside, since the perturbation is called directly from the network layer.

vieting:
> It is not really possible to move this outside, since the perturbation is called directly from the network layer.

I don't understand what you mean here.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,25 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # standard values for µ-law encoding
+                    quantization_bits = 8
+                    mu = 255.0
+
+                    # mu-law encoding formula
+                    encoded_audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
+                    encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
```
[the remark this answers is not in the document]
Max-Ryujin: No, but I can change that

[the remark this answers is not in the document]
vieting: Yes, please. See also below.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,25 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+
+                    # normalize encoded audio to -1 to 1
+                    encoded_normalized_audio = (encoded_audio / (2**quantization_bits - 1) * 2) - 1
```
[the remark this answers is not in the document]
vieting: Yes, in fact that is a requirement for mu-law encoding. Please add a check for this so that it fails if the input is outside that range. Then, normalizing the output accordingly is fine.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,25 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
-        import sox
-
-        tfm = sox.Transformer()
         for codec in codecs:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # standard values for mu-law encoding
+                    quantization_bits = 8
```
vieting, line 131: Why do you introduce this quantization aspect here? Is it needed for what we aim to do?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,25 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+                    mu = 255.0
+
+                    # mu-law encoding formula
+                    encoded_audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
+                    encoded_audio = ((encoded_audio + 1) / 2 * (2**quantization_bits - 1)).astype(np.float64)
```
vieting, line 136: We usually use `np.float32`. Is there a particular reason to change that?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,21 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # standard value for mu-law encoding
+                    mu = 255.0
+
+                    # mu-law encoding formula
+                    encoded_audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
+                    encoded_audio = ((encoded_audio + 1) / 2 * 255).astype(np.float32)
```
vieting, line 135: Why is this needed?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,21 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
-        tfm = sox.Transformer()
         for codec in codecs:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # standard value for mu-law encoding
+                    mu = 255.0
+
+                    # mu-law encoding formula
```
vieting, line 133: Didn't you previously check that the audio signal is in the range [-1, 1]?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,21 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
+                    encoded_audio = np.sign(audio) * np.log1p(mu * np.abs(audio)) / np.log1p(mu)
+                    encoded_audio = ((encoded_audio + 1) / 2 * 255).astype(np.float32)
+
+                    audio = encoded_audio
```
vieting, line 137: Is the `encoded_audio` variable really necessary?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: what exactly? This is the mu-law encoding step.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: added back in, but it should not be necessary as Switchboard is loaded in the correct format.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: No, but it makes the code more readable especially if more encodings are added in the future.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
vieting: The first line (currently 138) corresponds to the encoding equation e.g. on Wikipedia. I'm not sure where you take the second from. That's what my question was about.

Max-Ryujin: To move the value range to the same as the  `torchaudio.transforms.MuLawEncoding()` 0-255

Max-Ryujin: I wanted to stay consistent with the torchaudio implementation. The signal will be normalized by the wave_norm layer later anyway.

vieting: This code is supposed to be stand-alone. It should not depend on or make assumptions about what might happen somewhere else. I think if the input is in [-1, 1], it makes more sense to stay in that range for the output as well.

Max-Ryujin: Done

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
vieting: Do you still think it improves the readability? :thinking:

Max-Ryujin: I removed it for now. As long as there is no other encoding it should be clear.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -121,17 +123,20 @@ def preemphasis_numpy(waveform, coeff=0.97):
...
     def apply_codecs(audio, sample_rate, random_state, codecs):
-        import sox
-
-        tfm = sox.Transformer()
         for codec in codecs:
             prob = codec.pop("prob", 1.0)
             if random_state.random() < prob:
                 if codec.get("encoding") == "ULAW":
-                    tfm.set_output_format(encoding="u-law")
+                    # check if youdio is in the right range for mu-law encoding
```
vieting, line 130:
```suggestion
                    # check if audio is in the right range for mu-law encoding
```

vieting (approved):

PR #204, opened 2024-02-07
Allow unfreezing features in staged network dict

Max-Ryujin:
- added the ability to unfreeze features 
- added experiments that include 
  - starting with features from partition epoch 16 and freezing them for the first 16 partition epochs
  - freezing features after partition epoch 256 until the end of the training

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/experiments.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/experiments.py
@@ -1,100 +1,33 @@
 import copy
 import os.path
 from sisyphus import tk, gs
-from sisyphus.delayed_ops import DelayedFormat
 from typing import Any, Dict
 
-from i6_core.meta.system import CorpusObject
-from i6_core.lexicon.modification import AddEowPhonemesToLexiconJob
 from i6_core.returnn.config import CodeWrapper
 from i6_core.recognition import Hub5ScoreJob
 from i6_core.tools import CloneGitRepositoryJob
-from i6_experiments.common.datasets.switchboard.corpus_eval import get_hub5e00
-from i6_experiments.common.setups.rasr.util import RasrDataInput
-from i6_experiments.users.berger.recipe.lexicon.modification import DeleteEmptyOrthJob, MakeBlankLexiconJob
 from i6_experiments.users.vieting.tools.report import Report
-
-# TODO: run_gmm_system_from_common might be copied here for stability
-from i6_experiments.users.vieting.experiments.switchboard.hybrid.feat.experiments import run_gmm_system_from_common
+from i6_experiments.users.vieting.experiments.switchboard.transducer.feat.data import (
+    get_switchboard_data,
+    get_returnn_ogg_datasets,
+)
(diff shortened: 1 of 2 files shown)


users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -791,6 +791,128 @@ def run_scf_baseline():
...
+                    "staged_opts": {
+                        1: "freeze_features",
+                        16: "unfreeze_features",
+                    },
+                    "extra_args": {
+                        "accum_grad_multiple_step": 2,
+                        "conv_pad_seq_len_to_power": 1.5,
+                        "preload_from_files": {
+                            "existing-model": {
+                                "filename": "/u/maximilian.kannen/setups/20230406_feat/work/i6_core/returnn/training/ReturnnTrainingJob.xUBplWqSsRII/output/models/epoch.016.index",
```
vieting, line 835: You should never use absolute paths as strings in a setup like this. The clean solution would be to get the checkpoint from a previous training. In case this is just experimental for you to test this, I'd rather wait for you to check the results and only clean and merge this if it has shown to be helpful.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -791,6 +791,128 @@ def run_scf_baseline():
...
+                },
+                feature_args=feature_args,
+                lr_args={
+                    "peak_lr": 4e-4,
+                    "start_lr": 1.325e-05,
+                    "end_lr": 1e-5,
+                    "increase_epochs": 180,
+                    "decrease_epochs": 180,
+                    "final_epochs": 0,
+                },
```
vieting, line 857: If this is always the same, please declare a variable and re-use that.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -791,6 +791,128 @@ def run_scf_baseline():
...
+                            "existing-model": {
+                                "filename": "/u/maximilian.kannen/setups/20230406_feat/work/i6_core/returnn/training/ReturnnTrainingJob.xUBplWqSsRII/output/models/epoch.016.index",
+                                "init_for_train": True,
+                                "var_name_mapping": {
+                                    "/conv_h_filter/conv_h_filter": "features/conv_h_filter/conv_h_filter",
+                                    "/conv_l/W": "features/conv_l/W",
+                                    "/conv_l_act/bias": "features/conv_l_act/bias",
+                                    "/conv_l_act/scale": "features/conv_l_act/scale",
+                                },
+                                "prefix": "features",
```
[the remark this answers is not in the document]
Max-Ryujin: It is only possible to ignore certain variables. As far as I know this is the easiest way to do this

Max-Ryujin:
Since the freezing experiments did not show any improvement, I will remove most of these changes and only add the ability to unfreeze features

vieting (approved):

PR #205, opened 2024-02-07
Fix preemphasis and wave_norm configuration in feature_net

Max-Ryujin:
Fixes the correct layer structure for preemphasis with wave_norm. 
the line `if layer_config.get('class') != 'variable':` is needed for SCF networks.

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/baseline_args.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/baseline_args.py
@@ -81,14 +81,18 @@ def get_nn_args_single(
     feature_net = feature_network_class(**feature_args).get_as_subnetwork()
     if preemphasis:
         for layer in feature_net["subnetwork"]:
-            if feature_net["subnetwork"][layer].get("from", "data") == "data":
-                feature_net["subnetwork"][layer]["from"] = "preemphasis"
+            layer_config = feature_net["subnetwork"][layer]
+            if layer_config.get('class') != 'variable':
+                if feature_net["subnetwork"][layer].get("from") == "data":
+                    feature_net["subnetwork"][layer]["from"] = "preemphasis"
         feature_net["subnetwork"]["preemphasis"] = PreemphasisNetwork(alpha=preemphasis).get_as_subnetwork()
     if wave_norm:
         for layer in feature_net["subnetwork"]:
             if feature_net["subnetwork"][layer].get("from") == "data":
                 feature_net["subnetwork"][layer]["from"] = "wave_norm"
         feature_net["subnetwork"]["wave_norm"] = {"axes": "T", "class": "norm", "from": "data"}
+        if preemphasis:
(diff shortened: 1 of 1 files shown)


users/vieting/experiments/switchboard/ctc/feat/baseline_args.py
```diff
@@ -81,14 +81,18 @@ def get_nn_args_single(
...
+                if feature_net["subnetwork"][layer].get("from") == "data":
+                    feature_net["subnetwork"][layer]["from"] = "preemphasis"
         feature_net["subnetwork"]["preemphasis"] = PreemphasisNetwork(alpha=preemphasis).get_as_subnetwork()
     if wave_norm:
         for layer in feature_net["subnetwork"]:
             if feature_net["subnetwork"][layer].get("from") == "data":
                 feature_net["subnetwork"][layer]["from"] = "wave_norm"
         feature_net["subnetwork"]["wave_norm"] = {"axes": "T", "class": "norm", "from": "data"}
+        if preemphasis:
+            feature_net["subnetwork"]["preemphasis"]["from"] = "wave_norm"
```
vieting, line 95: Instead of adding very specific checks, I would prefer introducing a variable `source_layer` or so. Then you change the order of wave norm and preemphasis. Before the wave norm, you set it to `"data"` and for wave norm, you use that variable as `"from"` and then set it to `"wave_norm"`. Then for preemphasis, you again use that variable as `"from"` and then set it to `"preemphasis"`. That way, it would also be easier to extend then nested specific checks.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/baseline_args.py
```diff
@@ -78,17 +78,21 @@ def get_nn_args_single(
...
-        feature_net["subnetwork"]["wave_norm"] = {"axes": "T", "class": "norm", "from": "data"}
+        feature_net["subnetwork"]["wave_norm"] = {"axes": "T", "class": "norm", "from": source_layer}
+        source_layer = "wave_norm"
+    if preemphasis:
+        feature_net["subnetwork"]["preemphasis"] = PreemphasisNetwork(alpha=preemphasis).get_as_subnetwork(source=source_layer)
+        source_layer = "preemphasis"
+    for layer in feature_net["subnetwork"]:
+        if layer not in ["wave_norm", "preemphasis"]:
+            layer_config = feature_net["subnetwork"][layer]
+            if layer_config.get("class") != "variable" and layer_config.get("from") == "data":
```
vieting, line 94:
```suggestion
            if layer_config.get("class") != "variable" and layer_config.get("from", "data") == "data":
```
There might be layers which have no `"from"` defined and then RETURNN defaults to `"data"`.

vieting (changes requested):

vieting (approved):

PR #211, opened 2024-03-13
Experimental librispeech hybrid setup

Max-Ryujin:
[WIP]


common/setups/rasr/hybrid_system.py
```diff
@@ -228,10 +228,10 @@ def returnn_training(
         cv_corpus_key,
         devtrain_corpus_key=None,
     ) -> returnn.ReturnnTrainingJob:
-        if nn_train_args.returnn_root is None:
-            nn_train_args.returnn_root = self.returnn_root
-        if nn_train_args.returnn_python_exe is None:
-            nn_train_args.returnn_python_exe = self.returnn_python_exe
+        # if 'returnn_root' not in nn_train_args:
```
Atticus1806, line 231: Careful, this is not user code, but the common code. You must not change this. 
You probably commented this out, because it caused some crash in your setup. As long as the main branch here is not broken, I assume its some wrong configuration? 
If you think its not please make a separate PR with only this change for open review.

common/setups/rasr/hybrid_system.py
```diff
@@ -468,7 +469,6 @@ def nn_recog(
                     checkpoints=checkpoints,
                     train_job=train_job,
                     recognition_corpus_key=dev_c,
-                    acoustic_mixture_path=self.train_input_data[train_corpus_key].acoustic_mixtures,
```
Atticus1806, line 471: I think removing this is actually correct, this seems like a bug to me.
But please open a separate PR for this.

common/setups/rasr/hybrid_system.py
```diff
@@ -368,6 +368,7 @@ def nn_recognition(
         use_epoch_for_compile=False,
         forward_output_layer="output",
         native_ops: Optional[List[str]] = None,
+        acoustic_mixture_path: Optional[tk.Path] = None,
```
Atticus1806, line 371: This is not used afterwards right? This was removed. You probably had to add it because there is some call which includes this (c.f. other comment further below)
Lets better fix the wrong call.

Atticus1806 (changes requested):

PR #212, opened 2024-03-21
LatticeToNBestListJob

vieting:
Add a job that creates n-best lists from a lattice. This is currently intended to be used as a first step for creating a dataset with n-best data for transducer MBR training. The code is heavily based on the `LatticeToCtmJob`. I added it in my experimental folder for now, we might also directly go for i6_core but I thought we can test here and if it works move to i6_core.

### DIFF
--- a/users/vieting/jobs/nbest.py
+++ b/users/vieting/jobs/nbest.py
@@ -89,8 +89,6 @@ class LatticeToNBestListJob(rasr.RasrCommand, Job):
         config.flf_lattice_tool.network.archive_reader.links = "reduce-scores"
         config.flf_lattice_tool.network.archive_reader.format = "flf"
         config.flf_lattice_tool.network.archive_reader.path = lattice_cache
-        config.flf_lattice_tool.network.archive_reader.flf.append.keys = "confidence"
-        config.flf_lattice_tool.network.archive_reader.flf.append.confidence.scale = 0.0
 
         # dump n-best
         config.flf_lattice_tool.network.reduce_scores.type = "reduce"
--- a/users/vieting/jobs/nbest.py
+++ b/users/vieting/jobs/nbest.py
@@ -1,0 +1,117 @@
+__all__ = ["NBestListsFromLatticeJob"]
+
+import shutil
+import os
+
+from sisyphus import *
+
+import i6_core.rasr as rasr
+
+
+class NBestListsFromLatticeJob(rasr.RasrCommand, Job):
+    def __init__(
+        self,
+        crp,
+        lattice_cache,
+        n,
+        *,
+        parallelize=False,
+        extra_config=None,
+        extra_post_config=None,
+    ):
+        kwargs = locals()
+        del kwargs["self"]
+
+        self.config, self.post_config = self.create_config(**kwargs)
+        self.concurrent = crp.concurrent if parallelize else 1
+        self.exe = self.select_exe(crp.flf_tool_exe, "flf-tool")
+        self.lattice_cache = lattice_cache
+
+        self.out_log_file = self.log_file_output_path("nbest_from_lattice", crp, self.concurrent > 1)
+        self.out_nbest_file = self.output_path("nbest")
+
+        self.rqmt = {
+            "time": max(crp.corpus_duration / (5.0 * self.concurrent), 0.5),
+            "cpu": 1,
+            "mem": 4,
+        }
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt, args=range(1, self.concurrent + 1))
+        if self.concurrent > 1:
+            yield Task("merge", mini_task=True, tries=2)  # 2 tries to catch fs problems crashing the job
+
+    def create_files(self):
+        self.write_config(self.config, self.post_config, "nbest_from_lattice.config")
+        self.write_run_script(self.exe, "nbest_from_lattice.config")
+
+    def run(self, task_id):
+        log_file = self.out_log_file if self.concurrent <= 1 else self.out_log_file[task_id]
+        self.run_script(task_id, log_file)
+        if self.concurrent <= 1:
+            shutil.move("nbest.1", self.out_nbest_file.get_path())
+
+    def merge(self):
+        with open(self.out_nbest_file.get_path(), "wt") as out:
+            for t in range(1, self.concurrent + 1):
+                assert os.path.getsize("nbest.%d" % t) > 0, "Empty File nbest.%d, maybe restart merge" % t
+                with open("nbest.%d" % t, "rt") as f:
+                    shutil.copyfileobj(f, out)
+
+    @classmethod
+    def create_config(
+        cls,
+        crp,
+        lattice_cache,
+        n,
+        parallelize,
+        extra_config,
+        extra_post_config,
+        **kwargs,
+    ):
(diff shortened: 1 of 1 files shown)


users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
+class NBestListsFromLatticeJob(rasr.RasrCommand, Job):
+    def __init__(
+        self,
+        crp,
+        lattice_cache,
+        n,
+        *,
+        parallelize=False,
+        extra_config=None,
+        extra_post_config=None,
```
SimBe195, line 20: Proper typehints should be added.

users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
+        config.flf_lattice_tool.network.segment.type = "speech-segment"
+        config.flf_lattice_tool.network.segment.links = "0->archive-reader:1 0->dump-n-best:1"
+
+        # read lattice
+        config.flf_lattice_tool.network.archive_reader.type = "archive-reader"
+        config.flf_lattice_tool.network.archive_reader.links = "reduce-scores"
+        config.flf_lattice_tool.network.archive_reader.format = "flf"
+        config.flf_lattice_tool.network.archive_reader.path = lattice_cache
+        config.flf_lattice_tool.network.archive_reader.flf.append.keys = "confidence"
+        config.flf_lattice_tool.network.archive_reader.flf.append.confidence.scale = 0.0
```
michelwi, line 92: I see you appending a key "confidence" here, but it does not seem to be used later..?
```suggestion
```

users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
```
SimBe195, line 92: I believe this is not needed since confidence is never used afterwards.

users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # dump n-best
+        config.flf_lattice_tool.network.reduce_scores.type = "reduce"
+        config.flf_lattice_tool.network.reduce_scores.keys = "am lm"
+        config.flf_lattice_tool.network.reduce_scores.links = "n-best"
```
SimBe195, line 97: Maybe the reduce node in between should be made optional?

users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
```
michelwi, line 97: what is the intention behind reducing all scores to the "am" key?

users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
+
+        config.flf_lattice_tool.network.n_best.type = "n-best"
+        config.flf_lattice_tool.network.n_best.n = n
+        config.flf_lattice_tool.network.n_best.remove_duplicates = True
+        config.flf_lattice_tool.network.n_best.ignore_non_words = True
```
SimBe195, line 102: One could also make these configurable.

SimBe195 (changes requested):

users/vieting/jobs/nbest.py
```diff
@@ -0,0 +1,117 @@
...
```
michelwi, line 102: I see the "n-best" node takes a parameter score-key. Maybe you wanted to set it to "am" here after reducing all scores to am?

If it is unset, will it use the entire semiring score? Then we could still remove the "reduce" node..
edit, nevermind, this is to store the score: https://github.com/rwth-i6/rasr/blob/1c027984c3f53ff4f0f5f6476f4dce00cc03cdf2/src/Flf/NBest.cc#L1166-L1169

users/vieting/jobs/nbest.py
michelwi: are the non-words even dumped by the "dump-n-best" node? otherwise it would just be duplicate word sequences in the output file.

SimBe195:
It might also make sense to have an option for the level of the n-best list (word, phoneme, etc.). This could be done by inserting a `map-alphabet` node after the archive reader like it's also done in the `LatticeToCtmJob`:
```
config.flf_lattice_tool.network.to_lemma.type = "map-alphabet"
config.flf_lattice_tool.network.to_lemma.map_input = "to-lemma"|"to-phoneme"|...
config.flf_lattice_tool.network.to_lemma.project_input = True
config.flf_lattice_tool.network.to_lemma.links = ...
```

users/vieting/jobs/nbest.py
vieting: See also @SimBe195's comment
> Maybe the reduce node in between should be made optional?

I took the `LatticeToCtmJob` as a template and modified things based on an example config from Simon. You guys are more familiar with the flf tool than I am. What would you suggest here?

vieting:
> It might also make sense to have an option for the level of the n-best list (word, phoneme, etc.). This could be done by inserting a `map-alphabet` node after the archive reader like it's also done in the `LatticeToCtmJob`:
> 
> ```
> config.flf_lattice_tool.network.map_alphabet.type = "map-alphabet"
> config.flf_lattice_tool.network.map_alphabet.map_input = "to-lemma"|"to-phoneme"|...
> config.flf_lattice_tool.network.map_alphabet.project_input = True
> config.flf_lattice_tool.network.map_alphabet.links = ...
> ```

Should the `map_input` then be configurable with default `"to-lemma"`?

users/vieting/jobs/nbest.py
SimBe195: From the documentation of `n-best`:

> If the "ignore-non-word" option is activated, then two hypotheses only differing in non-words are considered duplicates.

So I would guess that the dump node includes non-words and that the `ignore-non-words` option only comes into play when `remove-duplicates` is active.

users/vieting/jobs/nbest.py
SimBe195: Right now this reduce node relies on the fact that `am` and `lm` are exactly the (relevant) score keys though which doesn't generalize well. So doing score summation separately when parsing might actually be preferable.

SimBe195:
> > It might also make sense to have an option for the level of the n-best list (word, phoneme, etc.). This could be done by inserting a `map-alphabet` node after the archive reader like it's also done in the `LatticeToCtmJob`:
> > ```
> > config.flf_lattice_tool.network.map_alphabet.type = "map-alphabet"
> > config.flf_lattice_tool.network.map_alphabet.map_input = "to-lemma"|"to-phoneme"|...
> > config.flf_lattice_tool.network.map_alphabet.project_input = True
> > config.flf_lattice_tool.network.map_alphabet.links = ...
> > ```
> 
> Should the `map_input` then be configurable with default `"to-lemma"`?

I just tested a bit and besides "to-lemma", "to-lemma-pron" and "to-preferred-lemma-pron" the other mapping options don't work straightforwardly. So maybe just a boolean flag for `word-level` that maps `to-lemma` and otherwise the pronunciation is dumped?

users/vieting/jobs/nbest.py
vieting: So just keep both configurable as it is now?

users/vieting/jobs/nbest.py
vieting: That would mean we remove the reduce node entirely?

SimBe195 (approved):

PR #216, opened 2024-04-22
add RNN-T init from CTC baseline

JackTemaki:
i6_models + i6_native_ops based pure PyTorch RNN-T with torchaudio beam search

PR #217, opened 2024-05-06
returnn setup for wav2vec2 ctc

AndreasPlt:
We moved the training of CTC wav2vec2 from fairseq to returnn. The setup is based on Nick's RNN-T example setup (see [here](https://github.com/rwth-i6/i6_experiments/pull/216))


users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/experiments/ctc_phon/baseline.py
```diff
@@ -0,0 +1,201 @@
...
+    tune_and_evaluate_helper(
+        training_name, asr_model, default_decoder_config, lm_scales=[2.3, 2.5, 2.7], prior_scales=[0.2, 0.3, 0.4]
+    )
+
+    # No improvement, just as example
+    # asr_model_best4 = prepare_asr_model(
+    #     training_name+ "/best4", train_job, train_args, with_prior=True, datasets=train_data, get_best_averaged_checkpoint=(4, "dev_loss_ctc")
+    # )
+    # tune_and_evaluate_helper(training_name + "/best4", asr_model_best4, default_decoder_config, lm_scales=[2.3, 2.5, 2.7], prior_scales=[0.2, 0.3, 0.4])
+
```
vieting, line 201: I guess that can be removed?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
```diff
@@ -0,0 +1,80 @@
...
+                target_dict.add_symbol(f'{i}')
+            return target_dict
+
+        task.state.add_factory('target_dictionary', _build_dummy_target_dict)
+        return task
+    
+    @classmethod
+    def from_dict(cls, cfg: dict):
+        ret = cls(**cfg)
+        print(ret)
```
vieting, line 77: Is this intended or was it just for debugging?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        Build a dummy task with a target dictionary of size `label_target_size`.
+        This is needed to instantiate the Wav2Vec2Ctc model.
+        """
+        task_cfg = AudioFinetuningConfig()
+        task = AudioFinetuningTask(task_cfg)
+
+        def _build_dummy_target_dict() -> Dictionary:
+            target_dict = Dictionary()
+            for i in range(self.label_target_size - 3): # -3 to account for the special tokens in fairseq dictionary
```
vieting, line 67: This is pretty hacky. Is there an easy nicer alternative?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
```diff
@@ -0,0 +1,80 @@
...
+        Build a full FairseqConfig object from the model and task configurations updates.
+        Non-specified fields will be filled with the default values.
+
+        FairseqConfig is needed to instantiate a FairseqModel.
+        """
+        fairseq_model_config = self._build_model_config()
+        fairseq_task_config = self._build_task_config()
+        fairseq_conf = FairseqConfig()
+        setattr(fairseq_conf, 'model', fairseq_model_config)
+        setattr(fairseq_conf, 'task', fairseq_task_config)
```
vieting, line 39: Can this be written as
```suggestion
        fairseq_conf = FairseqConfig(
            model=fairseq_model_config,
            task=fairseq_task_config,
        )
```

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
```diff
@@ -0,0 +1,80 @@
...
+        # convert to OmegaConf to resolve interpolations in the default config fields
+        return OmegaConf.structured(fairseq_conf)
+
+    def _build_model_config(self) -> Wav2Vec2CtcConfig:
+        model_cfg = Wav2Vec2CtcConfig()
+        return self._update_config(model_cfg, self.model_config_updates)
+
+    def _build_task_config(self) -> AudioFinetuningConfig:
+        task_cfg = AudioFinetuningConfig()
+        return self._update_config(task_cfg, self.task_config_updates)
```
vieting, line 50: I think the code might be easier to read if you just do this in the `build_full_config()` method.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
```diff
@@ -0,0 +1,80 @@
...
+            (see fairseq.models.wav2vec.wav2vec2_asr.Wav2Vec2CtcConfig)
+        task_config_updates: dict: Updates to the AudioFinetuningConfig
+            (see fairseq.tasks.audio_finetuning.AudioFinetuningConfig)
+        label_target_size: int: Size of the target dictionary.
+    """
+    model_config_updates: dict = dataclasses.field(default_factory=dict)
+    task_config_updates: dict = dataclasses.field(default_factory=dict)
+    label_target_size: int = 0
+
+    def build_full_config(self) -> FairseqConfig:
```
vieting, line 28: Wouldn't this name be more descriptive?
```suggestion
    def build_fairseq_config(self) -> FairseqConfig:
```

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+        Get the number of gradient updates that have been performed so far
+        """
+        return self.num_update
+    
+    def set_num_updates(self, num_update):
+        """
+        Set the number of gradient updates that have been performed so far
+        """
+        self.num_update = num_update
+        self.w2v_model.w2v_encoder.set_num_updates(num_update)
```
vieting, line 35: Is this needed for the w2v model to work properly?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+        **kwargs
+    ):
+        """
+        Args:
+            raw_audio: [B, T', 1] tensor (B: batch size, T': raw audio length)
+            raw_audio_len: [B] tensor
+            corpus_key: str
+        Returns:
+            logprobs: [B, T, C] tensor (C: vocab size, T: audio feature length)
+            input_lengths: [B] tensor
```
vieting, line 50: Isn't it rather output lengths?

vieting (changes requested):
I guess the only parts we really need to check is `w2v_ctc_wrapper_cfg.py`, `w2v_ctc_wrapper.py` and `experiments/ctc_phon/baseline.py`, right?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
AndreasPlt: It was just for debugging reasons, I will remove it (assuming that you are referring to `print(ret)`)

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
AndreasPlt: An alternative would be to remove the special tokens in the Dictionary object itself. IMO this would be even more hacky since it is required to change three Python lists/dicts (`symbols`, `count`, `indices`, see [here](https://github.com/facebookresearch/fairseq/blob/main/fairseq/data/dictionary.py)). Although the length of the dict is computed based only on `symbols`, I am not sure if it could lead to any kind of weird behavior if only this is changed without changing the other datastructures accordingly.

We could also change the `label_target_size` during instantiation of the config object, but this would lead to different values of the actual size of the target labels and the corresponding attribute in the config object. This would also require an additional `__post_init__` function since (afaik) this is the only way to add further initialization in a Python DataClass object.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
AndreasPlt: Yes, this is required in order to freeze the transformer for the first `freeze_finetune_updates` updates automatically. Otherwise, this would be needed to be done manually in our forward function.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
AndreasPlt: Good point, `output_lengths` would be a better fitting name.

AndreasPlt: Yes, but it corresponds to the parameter `input_lengths` for the CTC loss. IMO it's better to stay consistent and just name it the same way, otherwise we would call CTC loss with `input_lengths=output_lengths`, which would be equally confusing.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
AndreasPlt: Yes, this way the w2v_model can automatically freeze the transformer for the first `freeze_finetune_updates` updates. Otherwise, we would need to do this manually.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
AndreasPlt: A better possible name could be `input_feature_lengths` or just `feature_lengths,` what do you think?

AndreasPlt:
> I guess the only parts we really need to check is `w2v_ctc_wrapper_cfg.py`, `w2v_ctc_wrapper.py` and `experiments/ctc_phon/baseline.py`, right?

Yes. The only other change from Nick's example setup is in the `configs.py`, where I changed `max_seqs` to 1 in the `get_forward_config()`, as discussed

AndreasPlt:
I just added fairseq into the default tools, i.e. to use fairseq by locally installing it instead of installing it in the apptainer that executes the job. Therefore, besides `w2v_ctc_wrapper.py` and `w2v_ctc_wrapper_cfg.py`, `default_tools.py` and `serializer.py` are also changed, so might be worth to quickly have a look at

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper_cfg.py
vieting: Okay, I'm fine with leaving it at that.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
vieting: But that is outside of this function. You can use whatever variable name where you call this function, but I don't think the outside usage should define how we call the variables in here. `feature_lengths` seems alright here though.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/default_tools.py
```diff
@@ -0,0 +1,53 @@
...
+FAIRSEQ_PATH = SetupFairseqJob(
+    fairseq_root = CloneGitRepositoryJob(
+        url="git@github.com:vieting/fairseq_phoneme.git",
+        checkout_folder_name="fairseq",
+        commit="e4a2e4e93efbcbaaae52a17ae6600beb2083fb33",
+    ).out_repository.copy(),
+    python_exe = "/usr/bin/python3",
+).out_fairseq_root
+
+FAIRSEQ_PATH.hash_overwrite = "LIBRISPEECH_DEFAULT_FAIRSEQ"
```
vieting, line 39: We might need more different versions when experimenting with different pre-training variants, so I think it could be easier to get the fairseq path from a helper in the config file. Similar to the config for pre-training.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/serializer.py
```diff
@@ -0,0 +1,130 @@
...
+        unhashed_package_root=PACKAGE,
+        hashed_arguments=net_args,
+        unhashed_arguments=unhashed_net_args or {},
+        import_as="get_model",
+    )
+    pytorch_train_step = Import(
+        code_object_path=package + ".%s.train_step" % network_module, unhashed_package_root=PACKAGE
+    )
+    i6_models = ExternalImport(import_path=I6_MODELS_REPO_PATH)
+    fairseq = ExternalImport(import_path=FAIRSEQ_PATH)
```
vieting, line 42: I wonder whether there is a more flexible way to add this to the config if we might need to switch between fairseq versions for different fine-tunings.

AndreasPlt: I think `default_tools.py` is the right place to store this fairseq path, as it can be accessed globally that way. Maybe we could add a `set_fairseq_path` function in that file, that can be called from our function in `baseline.py` to use a specified fairseq version. What do you think about that?

vieting: Yes we can do that. But don't forget to serialize it for the config.

AndreasPlt: I implemented that with the newest commit (b4744ce)

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/default_tools.py
AndreasPlt: I guess this is also resolved with the newest commit?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/serializer.py
```diff
@@ -38,6 +37,8 @@ def serialize_training(
     pytorch_train_step = Import(
         code_object_path=package + ".%s.train_step" % network_module, unhashed_package_root=PACKAGE
     )
+    # paths might have been set differently in the mean time
+    from .default_tools import I6_MODELS_REPO_PATH, FAIRSEQ_PATH
```
vieting, line 41: Ah I see this is why you wanted to do it via the global variable. But can't you modify `get_training_config()` to pass the fairseq path to the serialization? I personally think it's cleaner not to always change the global variable, especially since the hash depends on it. It's easy to introduce strange behavior this way.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/sisyphus_configs/config_main_pretrain_finetune.py
```diff
@@ -0,0 +1,22 @@
...
+    "apply_mask": True,
+    "mask_prob": 0.65,
+    "mask_channel_prob": 0.5,
+    "mask_channel_length": 64,
+    "layerdrop": 0.1,
+    "activation_dropout": 0.1,
+    "feature_grad_mult": 0.0,
+    "freeze_finetune_updates": 10000 # was 0 in fairseq config
+}
+eow_phon_ls100_ctc_base(model_conf_w2v=model_conf_w2v, train_name_suffix=f"w2v_vanilla_cp_{checkpoint}_pipeline")
```
vieting, line 22:
```suggestion
eow_phon_ls100_ctc_base(model_conf_w2v=model_conf_w2v, train_name_suffix=f"w2v_vanilla_cp_{checkpoint}_pipeline")

```

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/sisyphus_configs/config_main_pretrain_finetune.py
```diff
@@ -0,0 +1,22 @@
+from i6_experiments.users.vieting.experiments.librispeech.\
+    librispeech_100_ctc.fairseq_finetuning.ctc_standalone.experiments.ctc_phon.baseline import eow_phon_ls100_ctc_base
```
vieting, line 2: Line too long? You can also apply `black` to your files which will take care of this automatically.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/sisyphus_configs/config_main_pretrain_finetune.py
```diff
@@ -0,0 +1,22 @@
...
+    "_name": "wav2vec_ctc",
+    "w2v_path": pretrain_job.out_models[checkpoint].model,
+    "apply_mask": True,
+    "mask_prob": 0.65,
+    "mask_channel_prob": 0.5,
+    "mask_channel_length": 64,
+    "layerdrop": 0.1,
+    "activation_dropout": 0.1,
+    "feature_grad_mult": 0.0,
+    "freeze_finetune_updates": 10000 # was 0 in fairseq config
```
vieting, line 20:
```suggestion
    "freeze_finetune_updates": 10000,  # was 0 in fairseq config
```
Again: I suggest applying black.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/sisyphus_configs/config_phoneme_pretrain_finetune.py
```diff
@@ -0,0 +1,28 @@
+from i6_experiments.users.vieting.experiments.librispeech.\
+    librispeech_100_ctc.fairseq_finetuning.ctc_standalone.experiments.ctc_phon.baseline import eow_phon_ls100_ctc_base
```
vieting, line 2: See above, please apply `black` to this file.

vieting (changes requested):

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/serializer.py
AndreasPlt: Done with the newest commit :)

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/default_tools.py
```diff
@@ -29,15 +29,26 @@
...
+    FAIRSEQ_PATH = SetupFairseqJob(
+        fairseq_root = CloneGitRepositoryJob(
+            url=url,
+            branch=branch,
+            checkout_folder_name="fairseq",
+            commit=commit,
+        ).out_repository,
+        python_exe = "/usr/bin/python3",
+    ).out_fairseq_root
+    FAIRSEQ_PATH.hash_overwrite = f"{url}::{branch}::{commit}"
```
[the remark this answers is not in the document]
AndreasPlt: Done with the newest commit :)

vieting (changes requested):
Some nitpicks.

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+                )
+
+        # make logprobs [B, T, C]
+        logprobs = logprobs.transpose(0, 1)
+
+        return logprobs, feature_lengths
+
+def train_step(*, model: Model, data, run_ctx, **kwargs):
+    raw_audio = data["raw_audio"]  # [B, T', 1]
+    raw_audio_len = data["raw_audio:size1"].to("cuda")  # [B] #TODO: check if this is correct
```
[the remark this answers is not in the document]
vieting: Was this really resolved?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/experiments/ctc_phon/baseline.py
```diff
@@ -0,0 +1,205 @@
...
+            )
+            pick_optimal_params_job.add_alias(training_name + f"/pick_best_{key}")
+            decoder_config = copy.deepcopy(base_decoder_config)
+            decoder_config.lm_weight = pick_optimal_params_job.out_optimal_parameters[0]
+            decoder_config.prior_scale = pick_optimal_params_job.out_optimal_parameters[1]
+            search_jobs, wers = search(
+                training_name,
+                forward_config={},
+                asr_model=asr_model,
+                decoder_module="decoder.flashlight_ctc_v1", # TODO: maybe need to change file
```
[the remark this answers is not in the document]
vieting: Was this really resolved?

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/pytorch_networks/wav2vec/w2v_ctc_wrapper.py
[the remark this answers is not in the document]
AndreasPlt: My bad, I forgot to push the corresponding commit

users/vieting/experiments/librispeech/librispeech_100_ctc/fairseq_finetuning/ctc_standalone/experiments/ctc_phon/baseline.py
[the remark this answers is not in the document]
AndreasPlt: My bad, I forgot to push the corresponding commit

vieting (approved):

PR #225, opened 2024-06-17
DeleteLemmataFromLexiconJob

vieting:
Hi @SimBe195,
We recently discussed about the sentence boundary symbol in my lexicon that does not work well together with the specific LMs at hand. I added this job to remove given lemmata from a lexicon. I added it into your collection of lexicon modification jobs initially and now that I have jobs depending on it, I did not want to move it because that would change the hashes. Therefore, I opened this PR so we can just merge it into your user folder if you agree.

### DIFF
--- a/users/berger/recipe/lexicon/modification.py
+++ b/users/berger/recipe/lexicon/modification.py
@@ -124,7 +124,6 @@ class DeleteLemmataFromLexiconJob(Job):
         out_lexicon.phonemes = in_lexicon.phonemes
         out_lexicon.lemmata = []
 
-        silence_phon = None
         for lemma in in_lexicon.lemmata:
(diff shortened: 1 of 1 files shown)


users/berger/recipe/lexicon/modification.py
```diff
@@ -101,6 +102,36 @@ def run(self):
...
+
+    def run(self):
+        in_lexicon = Lexicon()
+        in_lexicon.load(self.bliss_lexicon.get_path())
+
+        out_lexicon = Lexicon()
+        out_lexicon.phonemes = in_lexicon.phonemes
+        out_lexicon.lemmata = []
+
+        silence_phon = None
```
vieting, line 127:
```suggestion
```

SimBe195 (approved):

PR #235, opened 2024-07-31
Swb SCF transducer fix preloading

Max-Ryujin:
Fixed chunking and added wave norm layer mapping.


vieting:
I pushed a commit regarding the chunking into main. Please rebase and remove your changes on the chunking.

vieting (approved):

PR #236, opened 2024-07-31
Added stage 2 for swb SCF transducer

Max-Ryujin:
Nearly identical to log mel stage 2. Only changed names, feature args, specaug, and checkpoint

### DIFF
--- a/users/vieting/experiments/switchboard/transducer/feat/experiments.py
+++ b/users/vieting/experiments/switchboard/transducer/feat/experiments.py
@@ -486,8 +486,6 @@ def run_scf_stage2():
         "rasr_loss_args": {"transducer_training_stage": "fullsum"},
         "conformer_args": {"dropout": 0.25, "batch_norm_freeze": True},
     }
-    returnn_args_keep_hash = copy.deepcopy(returnn_args)
-    returnn_args_keep_hash["extra_args"]["learning_rate_control_error_measure"] = "sum_dev_score"
     feature_args = {
         "class": "ScfNetwork",
         "size_tf": 256 // 2,
--- a/users/vieting/experiments/switchboard/transducer/feat/experiments.py
+++ b/users/vieting/experiments/switchboard/transducer/feat/experiments.py
@@ -869,6 +869,8 @@ def run_scf_stage2():
         returnn_root=RETURNN_ROOT_FULLSUM,
         recog_args=recog_args,
     )
+    for training_name in nn_system.train_jobs:
+        nn_system.train_jobs[training_name].rqmt["gpu_mem"] = 24
     return nn_system, report
 
 
--- a/users/vieting/experiments/switchboard/transducer/feat/experiments.py
+++ b/users/vieting/experiments/switchboard/transducer/feat/experiments.py
@@ -818,7 +818,8 @@ def run_scf_stage2():
                 "label-scorer-type": "tf-ffnn-transducer",
                 "max-batch-size": 256,
                 "reduction-factors": 80 * 4,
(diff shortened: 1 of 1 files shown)


users/vieting/experiments/switchboard/transducer/feat/experiments.py
```diff
@@ -467,6 +467,97 @@ def run_mel_stage2():
...
+            "gradient_clip": 0.0,
+            "min_learning_rate": 1e-6,
+            "max_seq_length": {"classes": 600},
+        },
+        "specaug_old": {"max_feature": 15},
+        "rasr_loss_args": {"transducer_training_stage": "fullsum"},
+        "conformer_args": {"dropout": 0.25, "batch_norm_freeze": True},
+    }
+    returnn_args_keep_hash = copy.deepcopy(returnn_args)
+    returnn_args_keep_hash["extra_args"]["learning_rate_control_error_measure"] = "sum_dev_score"
```
vieting, line 490: This is not needed I assume

users/vieting/experiments/switchboard/transducer/feat/experiments.py
```diff
@@ -467,6 +467,97 @@ def run_mel_stage2():
     )
     return nn_system, report
 
+def run_scf_stage2():
```
vieting, line 470: Please move the function below `run_scf_stage1()`

users/vieting/experiments/switchboard/transducer/feat/experiments.py
```diff
@@ -467,6 +467,97 @@ def run_mel_stage2():
...
+    recog_args = {
+        "lm_scales": [0.3, 0.35, 0.4, 0.45],
+        "label_scorer_args": {
+            "extra_args": {
+                "blank-label-index": 0,
+                "context-size": 1,
+                "label-scorer-type": "tf-ffnn-transducer",
+                "max-batch-size": 256,
+                "reduction-factors": 80 * 4,
+                "reduction-subtrahend": 200 - 1,  # STFT window size - 1
```
vieting, line 508: This is log Mel specific and needs to be adapted. See my comment for `min_chunk_size` in stage 1 with SCF. Here, the subtrahend can be computed as (40 - 1) * 5 + 128 iirc. However, the RASR implementation is not based on how convolutions actually work but only allows for a single "subtrahend" not one for each layer. So we have to additionally subtract 1 to obtain 322 which should work for SCF. Please fix this and make a comment about this.

vieting (changes requested):

users/vieting/experiments/switchboard/transducer/feat/experiments.py
```diff
@@ -782,6 +782,96 @@ def run_scf_stage1():
...
+    recog_args = {
+        "lm_scales": [0.3, 0.35, 0.4, 0.45],
+        "label_scorer_args": {
+            "extra_args": {
+                "blank-label-index": 0,
+                "context-size": 1,
+                "label-scorer-type": "tf-ffnn-transducer",
+                "max-batch-size": 256,
+                "reduction-factors": 80 * 4,
+                # There have to be (40 - 1) * 5 + 128 samples to create one feature frame. RETURNN needs -1.
```
vieting, line 821:
```suggestion
                # There have to be (40 - 1) * 5 + 128 samples to create one feature frame. RASR needs -1.
```

users/vieting/experiments/switchboard/transducer/feat/experiments.py
vieting: Yes, since both features operate on 10ms level.

PR #241, opened 2024-08-18
Adding a new more customisable SpecAugment Implementation

Max-Ryujin:
- Added a new SpecAugment Implementation that supports custom scheduling for time and frequency masks and sorting.
- Added a first set of experiments with the new SpecAugment


users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -798,40 +797,65 @@ def run_scf_specaug_sort():
...
+                returnn_args={**base_returnn_args, "specaug_new": {"enable_sorting": True, "mask_growth_strategy": "step", "max_feature": 15, "steps_per_epoch": 4100}},
+                feature_args=feature_args,
+                lr_args=lr_args,
+                report_args={"batch_size": "2x5k", "experiment": "nonlinear_growth_with_sorting"},
             ),
         },
         num_epochs=450,
         evaluation_epochs=[350, 390, 400, 410, 450],
-        prefix="conformer_",
+        prefix="conformer_specaug_",
```
vieting, line 858:
```suggestion
        prefix="conformer_bs2x5k_scf_specaug_",
```

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -798,40 +797,65 @@ def run_scf_specaug_sort():
...
-        nn_base_args={
-            "bs2x5k_scf_specaugsortlayer2": dict(
-                returnn_args={**returnn_args, "specaug_old": {"max_feature": 15, "sort_layer2": True}},
+        nn_base_args = {
+            "baseline_with_sorting": dict(
+                returnn_args={**base_returnn_args, "specaug_new": {"enable_sorting": True, "max_feature": 15, "steps_per_epoch": 4100}},
                 feature_args=feature_args,
                 lr_args=lr_args,
-                report_args={"batch_size": "2x5k"},
+                report_args={"batch_size": "2x5k", "experiment": "baseline_with_sorting"},
```
vieting, line 805: `"experiment"` does not like a suitable descriptive header for what you are describing here. I should somehow represent that this is about SpecAugment.

users/vieting/experiments/switchboard/ctc/feat/fullsum_ctc_raw_samples.py
```diff
@@ -167,6 +171,8 @@ def make_conformer_fullsum_ctc_model(
                 **specaug_old,
             }
             from_list, python_code = specaug_func(network, from_list=from_list, **specaug_old_args)
+        elif specaug_new is not None:
```
vieting, line 174: Maybe add similar check as above:
```suggestion
        elif specaug_new is not None:
            assert specaug_old is None
```

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+        name (str, optional): The name of the SpecAugment layer. Defaults to "specaug".
+        from_list (Optional[Union[str, List[str]]], optional): The input layer(s) for the SpecAugment layer.
+            Default is ["data"].
+        config (Dict, optional): A dictionary containing configuration options for the SpecAugment layer.
+            Config Dict Parameters:
+                max_time_num (int): The beginning maximum number of time masks to be applied. Default is 3.
+                max_time (int): The beginning maximum size (in time frames) of each time mask. Default is 10.
+                max_feature_num (int): The beginning maximum number of frequency masks to be applied. Default is 4.
+                max_feature (int): The beginning maximum size (in frequency bins) of each frequency mask. Default is 5.
+                enable_sorting (bool): Whether to sort filters by their center frequency before applying masks. Default is True.
```
vieting, line 28: I would set the default to `False`.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+                steps_per_epoch (int): The number of steps per epoch. Make sure this parameter is accurate since the all the scheduling depends on it. Default is 2080.
```
vieting, line 29: This is dangerous, it can easily be overlooked and that would introduce unexpected behaviour. I suggest to force configuring this.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+                sorting_start_epoch (int): The subepoch number to start sorting filters. Default is 0.
+                mask_growth_strategy (str): The strategy for increasing the mask sizes over epochs (linear or step). Default is "linear".
+                time_mask_num_schedule (Dict[int, float]): A dictionary mapping subepoch numbers to the multiplicator for the time mask number.
+                    Default is {0: 1, 10: 1.5, 20: 2}.
+                time_mask_size_schedule (Dict[int, float]): A dictionary mapping subepoch numbers to the multiplicator for the time mask size.
+                    Default is {0: 1, 15: 1.5, 25: 2}.
+                freq_mask_num_schedule (Dict[int, float]): A dictionary mapping subepoch numbers to the multiplicator for the frequency mask number.
+                    Default is {0: 1, 5: 1.2, 15: 1.5}.
+                freq_mask_size_schedule (Dict[int, float]): A dictionary mapping subepoch numbers to the multiplicator for the frequency mask size.
+                    Default is {0: 1, 10: 1.2, 20: 1.5}.
```
vieting, line 39: I would not set such defaults. They are highly tuning-dependent. They should be in the `experiments.py` file, not here. The baseline behaviour should be the regular SpecAugment behaviour imho.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+
+    Returns:
+        tuple: A tuple containing the name of the SpecAugment layer and the functions required for it.
+    """
+    from .specaug_param_helper import generate_specaug_params
+
+    if from_list is None:
+        from_list = ["data"]
+
+    default_config = {
```
vieting, line 52: See above

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+    network[name] = {
+        "class": "eval",
+        "from": from_list,
+        "eval": f"self.network.get_config().typed_value('transform')(source(0, as_data=True), network=self.network, **{full_config})",
+    }
+
+    return [name], get_specaug_funcs()
+
+
+def sort_filters_by_center_freq(x):
```
vieting, line 108: You can import functions from the existing file if they are identical.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+    sorted_indices = tf.cond(use_sorting, get_sorted_indices, get_unsorted_indices)
+
+    def get_masked():
+        x_masked = x
+        time_mask_max_size = tf.gather(specaug_params["time_mask_max_size"], current_epoch)
+        time_mask_max_num = tf.gather(specaug_params["time_mask_max_num"], current_epoch)
+        freq_mask_max_num = tf.gather(specaug_params["freq_mask_max_num"], current_epoch)
+        freq_mask_max_size = tf.gather(specaug_params["freq_mask_max_size"], current_epoch)
+        time_mask_max_proportion = tf.cast(
+            tf.math.floor(config["time_mask_max_proportion"] * tf.cast(tf.shape(x)[data.time_dim_axis], tf.float32)),
```
vieting, line 263: Why is this cast needed?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+        # check for the limits
+        actual_time_mask_max_num = tf.minimum(time_mask_max_num, time_mask_max_proportion // time_mask_max_size)
+        actual_freq_mask_max_num = tf.minimum(freq_mask_max_num, freq_mask_max_proportion // freq_mask_max_size)
+
+        x_masked = random_mask(
+            x_masked,
+            batch_axis=data.batch_dim_axis,
+            axis=data.time_dim_axis,
+            min_num=0,
+            max_num=actual_time_mask_max_num,
```
vieting, line 280: This is different from the existing implementation which is
```
max_num=tf.maximum(
    tf.shape(x)[data.time_dim_axis] // int(1.0 / 0.7 * max_time),
    max_time_num,
) // (1 + increase_flag)
```
You don't have this `increase_flag` to reduce the number at the beginning of the training (need to check for stability). Also the first part of the maximum in the existing variant is missing. Did you think about what effect this had and how this changes in your case? This might be critical.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+        x_masked = x
+        time_mask_max_size = tf.gather(specaug_params["time_mask_max_size"], current_epoch)
+        time_mask_max_num = tf.gather(specaug_params["time_mask_max_num"], current_epoch)
+        freq_mask_max_num = tf.gather(specaug_params["freq_mask_max_num"], current_epoch)
+        freq_mask_max_size = tf.gather(specaug_params["freq_mask_max_size"], current_epoch)
+        time_mask_max_proportion = tf.cast(
+            tf.math.floor(config["time_mask_max_proportion"] * tf.cast(tf.shape(x)[data.time_dim_axis], tf.float32)),
+            tf.int32,
+        )
+        freq_mask_max_proportion = tf.cast(
```
vieting, line 266: See above

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+
+    sorted_indices = tf.cond(use_sorting, get_sorted_indices, get_unsorted_indices)
+
+    def get_masked():
+        x_masked = x
+        time_mask_max_size = tf.gather(specaug_params["time_mask_max_size"], current_epoch)
+        time_mask_max_num = tf.gather(specaug_params["time_mask_max_num"], current_epoch)
+        freq_mask_max_num = tf.gather(specaug_params["freq_mask_max_num"], current_epoch)
+        freq_mask_max_size = tf.gather(specaug_params["freq_mask_max_size"], current_epoch)
+        time_mask_max_proportion = tf.cast(
```
vieting, line 262: This variable name does not seem to be well chosen. Once you multiply with the sequence length, it's not a proportion any more, is it?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+            tf.math.floor(config["time_mask_max_proportion"] * tf.cast(tf.shape(x)[data.time_dim_axis], tf.float32)),
+            tf.int32,
+        )
+        freq_mask_max_proportion = tf.cast(
+            tf.math.floor(config["freq_mask_max_proportion"] * tf.cast(tf.shape(x)[data.feature_dim_axis], tf.float32)),
+            tf.int32,
+        )
+
+        # check for the limits
+        actual_time_mask_max_num = tf.minimum(time_mask_max_num, time_mask_max_proportion // time_mask_max_size)
```
vieting, line 272: I wonder if this makes sense. If you have max proportion of 0.5 and a sequence length of 100, you want to mask at most 50 frames. The default max mask size is 10. However, you cannot tell the individual mask sizes here. You would get an actual max number of 5. This is somewhat pessimistic since the actual masks could be smaller resulting in less masked frames than intended.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_param_helper.py
```diff
@@ -0,0 +1,65 @@
+
```
vieting, line 1: Please apply black to this file.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_param_helper.py
```diff
@@ -0,0 +1,65 @@
...
+        growth_strategy (str): Strategy for mask growth over epochs.
+
+    Returns:
+        Dict[str, List[int]]: Pre-generated parameter values for each epoch.
+    """
+    params = {key: [] for key in base_values.keys()}
+
+    for epoch in range(num_epochs):
+        for key, base_value in base_values.items():
+            schedule = schedules[key]
```
vieting, line 26: Maybe you should assert that 0 (or 1, depending on where you start) is among the keys?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -0,0 +1,300 @@
...
+    default_config = {
+        "max_time_num": 3,
+        "max_time": 10,
+        "max_feature_num": 4,
+        "max_feature": 5,
+        "enable_sorting": True,
+        "sorting_start_epoch": 0,
+        "steps_per_epoch": 2080,
+        "mask_growth_strategy": "linear",
+        "time_mask_num_schedule": {0: 1, 10: 1.5, 20: 2},
```
vieting, line 61: Wouldn't it feel more natural to start with 1 instead of 0?

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: I do not know if an how it would be possible to force an entry in a dict.
I could add an assertion or remove the parameter from the dict and add it as a separate parameter.
An assertion seems better because it would be unintuitive if just one of the specaugment parameter needs to be set separately from the others.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: Does this also work with 
```
def get_specaug_funcs() -> list:
    return [sort_filters_by_center_freq, _mask, random_mask, transform]
```
I thought the function has to in the same file so I can write it into the config.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: I got type errors without it. I think because the output of math.floor has the same type as the input and since `config["freq_mask_max_proportion"]` is a float the type would be wrong

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: Yes, I tried to remove as much hard-coded things as possible from the SpecAugment implementation. This gets replaced by the scheduling and `freq_mask_max_proportion` and  `time_mask_max_proportion` replace the maximum. This way I can set a upper limit for the masks and change the mask growths as well. I thought about making the randomness of the mask sizes and numbers also configurable but decided against it for simplicity.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: time_mask_max_proportion is intended as a safety check to ensure the scheduling is not to aggressive. It should not be reached if scheduling,, base_values, steps per epoch, num_epochs, and the number of features is all known and set correctly. I tried to do it as a assertion first, but it is not possible to build assertion into the tf graph.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: Yes, just remove the default and assert that the key exists.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: I don't think so, but you can just give it a try.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: I see that you have similar options for configuration. I just feat that your current version results in very different behaviour compared to the existing version. This would introduce significant tuning effort for your version. Previously, `max_num` was the maximum of `max_time_num` and a term that grows with the sequence length. In your case, you take the minimum.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -786,6 +786,13 @@ def run_scf_specaug():
         },
         "conformer_type": "wei",
     }
+    base_specaug_config = {
+        "steps_per_epoch": 4100,
+        "time_mask_num_schedule": {0:1, 10: 1.5, 20: 2},
```
vieting, line 791: Please apply black.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -25,18 +25,18 @@ def add_specaug_layer(
...
                 max_feature_num (int): The beginning maximum number of frequency masks to be applied. Default is 4.
                 max_feature (int): The beginning maximum size (in frequency bins) of each frequency mask. Default is 5.
-                enable_sorting (bool): Whether to sort filters by their center frequency before applying masks. Default is True.
+                enable_sorting (bool): Whether to sort filters by their center frequency before applying masks. Default is False.
                 steps_per_epoch (int): The number of steps per epoch. Make sure this parameter is accurate since the all the scheduling depends on it. Default is 2080.
                 sorting_start_epoch (int): The subepoch number to start sorting filters. Default is 0.
                 mask_growth_strategy (str): The strategy for increasing the mask sizes over epochs (linear or step). Default is "linear".
                 time_mask_num_schedule (Dict[int, float]): A dictionary mapping subepoch numbers to the multiplicator for the time mask number.
-                    Default is {0: 1, 10: 1.5, 20: 2}.
+                    Default is None.
```
vieting, line 33: Maybe add the information that this will result in a multiplier of 1 for all epochs.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -50,18 +50,14 @@ def add_specaug_layer(
...
     default_config = {
-        "max_time_num": 3,
-        "max_time": 10,
-        "max_feature_num": 4,
-        "max_feature": 5,
-        "enable_sorting": True,
+        "max_time_num": 1,
+        "max_time": 15,
+        "max_feature_num": 5,
+        "max_feature": 4,
```
vieting, line 56: Is this consistent with the values we used previously?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -50,18 +50,14 @@ def add_specaug_layer(
...
         "sorting_start_epoch": 0,
         "steps_per_epoch": 2080,
-        "mask_growth_strategy": "linear",
-        "time_mask_num_schedule": {0: 1, 10: 1.5, 20: 2},
-        "time_mask_size_schedule": {0: 1, 15: 1.5, 25: 2},
-        "freq_mask_num_schedule": {0: 1, 5: 1.2, 15: 1.5},
-        "freq_mask_size_schedule": {0: 1, 10: 1.2, 20: 1.5},
+        "mask_growth_strategy": "linear", 
         "time_mask_max_proportion": 0.7,
         "freq_mask_max_proportion": 0.7,
```
vieting, line 62: I would set the default such that this has no influence.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_param_helper.py
```diff
@@ -19,11 +16,15 @@ def generate_specaug_params(
     Returns:
         Dict[str, List[int]]: Pre-generated parameter values for each epoch.
     """
+
+    if schedules is None:
```
vieting, line 20: According to the typing information above, it should not be `None`.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: I really want to avoid hidden scheduling in the mask sizes when I have an explicit scheduling now. This will need some new tuning, but tuning the masks for SCF was the goal of this implementation anyway.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: Yes, these are the default values set for specaug_old in `fullsum_ctc_raw_samples.py`

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_param_helper.py
Max-Ryujin: Right this check is useless

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -27,18 +28,18 @@ def add_specaug_layer(
                 max_feature (int): The beginning maximum size (in frequency bins) of each frequency mask. Default is 5.
                 enable_sorting (bool): Whether to sort filters by their center frequency before applying masks. Default is False.
                 steps_per_epoch (int): The number of steps per epoch. Make sure this parameter is accurate since the all the scheduling depends on it.
-                sorting_start_epoch (int): The subepoch number to start sorting filters. Default is 0.
+                sorting_start_epoch (int): The subepoch number to start sorting filters. Default is 1.
                 mask_growth_strategy (str): The strategy for increasing the mask sizes over epochs (linear or step). Default is "linear".
                 time_mask_num_schedule (Dict[int, float]): A dictionary mapping subepoch numbers to the multiplicator for the time mask number.
-                    Default is None.
+                    Default is None (unchanged over the enitre training).
```
vieting, line 34:
```suggestion
                    Default is None (unchanged over the entire training).
```
Typo, also below.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -258,18 +141,18 @@ def get_masked():
         time_mask_max_num = tf.gather(specaug_params["time_mask_max_num"], current_epoch)
         freq_mask_max_num = tf.gather(specaug_params["freq_mask_max_num"], current_epoch)
         freq_mask_max_size = tf.gather(specaug_params["freq_mask_max_size"], current_epoch)
-        time_mask_max_proportion = tf.cast(
+        total_time_masks_max_size = tf.cast(
```
vieting, line 144:
```suggestion
        total_time_masks_max_frames = tf.cast(
```
Wouldn't this reflect it better?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: I see this point. But can you then please provide me with some simulations (like a plot max_num over epochs) on how this develops over the training with the old baseline implementation and your current baseline?

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: plot
Without scheduling it is missing the slow start for the first 2000 steps. 
The same with the current scheduling that I added but here the number of masks also gets bigger after 5 epochs. I could change the default scheduling to go from 2 to 5 after one epoch. That would be the closest I can get to the old behavior with the new Implementation

vieting: Can you please compare original with your implementation without schedule for different sequence lengths? This is where things become interesting.

Max-Ryujin: plot_time
I see the Problem. The old implementation scaled with the sequence length while the new one will stay static (3) independent of the input length. 
- I could add a multiplicator to the number of masks based on the sequence length (fixed or changeable)
- I could make it a new parameter (minimal proportion) similar to the maximum proportion 
- I could make it a fixed minimum percentile like it is in the old implementation

vieting: This is not at all about a minimum. The minimum number of masks is always 0 in the existing implementation. This is about the maximum only. And in the existing case, the maximum is `max(<length_dependent_factor>, max_time_num)`. The simplest option would be to introduce a possibility to allow configuring this accordingly. E.g. by something like `max_time_num_seq_len_divisor` which could be `int(1.0 / 0.7 * max_time)` like in the default case.

Max-Ryujin: But this would be the lower limit for the maximum number of masks. In the old implementation, it was possible to set `max_time_num` higher than this and cover the entire input. That is why I said minimum. I can add a `max_time_num_seq_len_divisor ` but I am not sure why I need the `max_time_num` parameter at all if the mask number is most of the time just dependent on the input length.

vieting: > I am not sure why I need the max_time_num parameter at all if the mask number is most of the time just dependent on the input length.

That's a valid question. I think it would be good to have some statistics on the actual masking that is applied in training. I suggest you allow the implementation here to replicate the existing behaviour. Then you come up with a way to collect the mentioned statistics and then we can draw conclusions on how future experiments should be configured.

vieting: Following up on this: Can you please also change the experiments introduced in this PR to a minimal baseline that is just reproducing the old behavior? This could serve as a sanity check and at the same time, we could collect said statistics. With the statistics available, you can then add a set of experiments varying some of the factors.

Max-Ryujin: I added an experiment that is very close to the old baseline, but I can not replicate the `increase_flag` behavior, since the scheduling can only change the sizes from one epoch to the next and not inside the first epoch.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -811,6 +811,15 @@ def run_scf_specaug():
 
     nn_args, report_args_collection = get_nn_args_baseline(
         nn_base_args={
+            "old_baseline": dict(
```
vieting, line 814:
```suggestion
            "baseline": dict(
```
Please call the existing thing is the baseline, not the new stuff.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: You could run both the current version and a version where you apply the behavior of the old `increase_flag` for the full first subepoch.

vieting (changes requested):
Just smaller comments. Please remove all the new "experimental" trainings though as discussed in one thread. We should first get the mask statistics to carefully design an experimental plan instead of blindly running different versions where the effect on the actual mask sizes is not clear.

vieting:
Also, some lines are too long in `specaug_configurable.py` (mainly comment lines). Can you please also fix that?

vieting (changes requested):
As I said before: Please remove all the new "experimental" trainings though as discussed in one thread. We should first get the mask statistics to carefully design an experimental plan instead of blindly running different versions where the effect on the actual mask sizes is not clear.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -799,39 +804,35 @@ def run_scf_specaug_sort():
...
+                    **base_returnn_args,
+                    "specaug_config": {"enable_sorting": False, "max_feature": 15, "steps_per_epoch": 4100},
+                },
                 feature_args=feature_args,
                 lr_args=lr_args,
-                report_args={"batch_size": "2x5k"},
+                report_args={"batch_size": "2x5k", "enable_sorting": False, "max_feature": 15},
             ),
-            "bs2x5k_scf_specaugsortlayer2frome210": dict(
+            "baseline_increase_flag": dict(
```
vieting, line 816:
```suggestion
            # note this only reduces the masking in frequency dimension, not in time.
            "baseline_increase_flag": dict(
```

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -799,39 +804,35 @@ def run_scf_specaug_sort():
...
-                            },
-                        },
-                        **returnn_args["extra_args"],
+                    **base_returnn_args,
+                    "specaug_config": {
+                        "enable_sorting": False,
+                        "max_feature": 15,
+                        "max_feature_num": 2,
+                        "steps_per_epoch": 4100,
+                        "time_mask_num_schedule": {0: 1, 1: 2},
```
vieting, line 824: This does not make sense, does it? You should reduce it to half in the first epoch, but instead you double it from the second. I think it's anyway not possible to replicate the old behavior with the current implementation because the division by 2 is after the max with the sequence length dependent part. But maybe let's try if it works like this (without doubling like you do here) and only if not replicate in more detail.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/experiments.py
Max-Ryujin: I reduced the start point and double it to the previous value. for the time dimension it is as you say, there is no way to replicate the old behavior. this is the closest that is possible.

vieting: Why is this the closest possible? Why this `time_mask_num_schedule`? Shouldn't it be without? Maybe it does not make a difference, but this factor of two here does not make sense to me.

Max-Ryujin: I did not realize that I still had the second scheduling in there. It should only be the one for the features. I forgot to remove it after I realized that I could not replicate the behavior.

PR #245, opened 2024-08-23
Fixed a datatype bug in specaugment

Max-Ryujin:
Converted `max_time_num_seq_len_divisor` to the correct data type

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
@@ -163,7 +163,7 @@ def transform(data, network, **config):
         max_time_num_seq_len = tf.cast(
             tf.math.floordiv(
                 tf.cast(tf.shape(x)[data.time_dim_axis], tf.float32),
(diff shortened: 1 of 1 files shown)


users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -159,12 +160,18 @@ def get_masked():
             tf.math.floor(config["freq_mask_max_proportion"] * tf.cast(tf.shape(x)[data.feature_dim_axis], tf.float32)),
             tf.int32,
         )
-
+        max_time_num_seq_len = tf.cast(
+            tf.math.floordiv(
+                tf.cast(tf.shape(x)[data.time_dim_axis], tf.float32),
+                max_time_num_seq_len_divisor * tf.cast(time_mask_max_size, tf.float32),
```
vieting, line 166: Shouldn't it be something like 1 / this line?

Max-Ryujin: Yes, I missed that.

vieting (approved):

PR #246, opened 2024-08-26
Adding filter based masks to SpecAugment

Max-Ryujin:
- adding new functions to mask based on variance
- adding needed parameters to the config


users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -79,40 +85,114 @@ def add_specaug_layer(
...
+    """
+    import tensorflow as tf
+
+    n_batch = tf.shape(x)[batch_axis]
+    n_features = tf.shape(x)[axis]
+
+    # Determine the number of masks for this batch
+    num_masks = tf.random.uniform(shape=(), minval=0, maxval=total_max_number_masks + 1, dtype=tf.int32)
+
+    # Calculate probabilities based on variance and variance_factor
```
vieting, line 175: Can you please elaborate the motivation of the exact usage of the `variance_factor`?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -79,40 +85,114 @@ def add_specaug_layer(
...
+
+    # Create a boolean mask using tf.where
+    indices = tf.expand_dims(masked_features, axis=-1)
+    mask_2d = tf.scatter_nd(indices, tf.zeros_like(indices, dtype=tf.float32), [n_features])
+    mask_2d = tf.reshape(mask_2d, [1, -1])
+
+    # Broadcast the mask to match x's shape
+    mask = tf.tile(mask_2d, [n_batch, 1])
+
+    # maybe this is better? and closer to the original code
```
vieting, line 192: This looks like it needs some polishing...

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -49,6 +49,9 @@ def add_specaug_layer(
                 freq_mask_max_proportion (float): The maximum proportion of the frequency axis that can be masked. Default is 1.
                 max_time_num_seq_len_divisor (int):
                     The divisor for the sequence length to determine the maximum number of time masks. Default is 0.7.
+                mask_based_on_variance (bool): Whether to mask based on variance or not. Default is False.
+                variance_factor (float): The factor that determines the probability of masking a feature based on variance. Default is 0.5.
+                total_max_number_masks (int): The total maximum number of masks to be applied if variance-based masking is used. Default is 50.
```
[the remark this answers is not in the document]
Max-Ryujin: Since variance based masking does not use a number of different sizes of masks and instead is masking single features based on variance, I need a parameter for the maximum number of masked features. I could calculate it using the existing ones (max_feature_num * max_feature) but that would overload the variables and make the behavior very hard to understand.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: It allows for control over the influence of the variance. 0 means total random selection of masked features and 1 means the masking is purely based on the softmax over the variance. I expect the best solution to be somewhere in between.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -201,5 +282,55 @@ def get_masked():
     return x
 
 
-def get_specaug_funcs() -> list:
-    return [sort_filters_by_center_freq, _mask, random_mask, transform]
+def transform_with_variance(data, max_time_num, max_time, variance_factor, total_max_number_masks, network):
```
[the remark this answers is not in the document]
Max-Ryujin: It was not clear to me how a combination of both would look like so I separated them. Scheduling could be useful. It will require another schedule parameter in the config for the same reason I needed a new max mask parameter.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -79,40 +85,114 @@ def add_specaug_layer(
...
+
+    x = tf.convert_to_tensor(filter_layer)  # (256, 1, 150) = (N, 1, C)
+    filters = tf.transpose(tf.squeeze(x))  # (C, N)
+    num_freqs = 128  # F
+    w = tf.linspace(0.0, np.pi - np.pi / num_freqs, num_freqs)  # (F,)
+    zm1 = tf.expand_dims(tf.exp(-1j * tf.cast(w, "complex64")), 1)  # (F, 1)
+    exponents = tf.expand_dims(tf.range(tf.shape(filters)[1]), 0)  # (1, N)
+    zm1_pow = tf.pow(zm1, tf.cast(exponents, dtype="complex64"))  # (F, N)
+    f_resp = tf.tensordot(zm1_pow, tf.cast(tf.transpose(filters), dtype="complex64"), axes=1)  # (F, C)
+    f_resp = tf.abs(f_resp)
```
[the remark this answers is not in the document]
Max-Ryujin: The old function in `specaug_sort_layer2.py` returns sorted indices, but I can split it up to share the fist part between both.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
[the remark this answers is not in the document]
vieting: Alright, but then please make the name reflect that this is only for the filter content based masking.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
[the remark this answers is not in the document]
vieting: Right, if you just mask single channels based on their frequency response's variance, the sorting has nothing to do with this. I think more detailed control over the schedule would be helpful, though.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -201,5 +282,55 @@ def get_masked():
...
+            variance_based_mask,
+            transform_with_variance,
+        ]
+    else:
+        return [
+            sort_filters_by_center_freq,
+            _mask,
+            random_mask,
+            transform,
+        ]
```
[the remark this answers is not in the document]
vieting: I don't know, probably "historic reasons" :D Let's just keep the old file like it is, that doesn't affect this PR.

vieting:
I'm wondering if it would be better to have a more generic modification in two aspects:

1. This is specifically tailored for variance based masking. But there are other ways to determine masking based on the filters' frequency responses, like bandwidth-based or so. It would be helpful to make the code here more generic to allow for other types in the future.
2. Right now, it's either the standard random masking or the variance based masking. It could be helpful to have them separate, so that a combination would be possible. This would e.g. allow for a schedule, where the variance-based masking is slowly increased over the training without having to refrain from SpecAugment in the beginning. It would still be possible to have the single behaviour by setting the respective other masking parameters to 0 (or so).

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -176,6 +256,7 @@ def get_masked():
             total_time_masks_max_frames // time_mask_max_size,
         )
         actual_freq_mask_max_num = tf.minimum(freq_mask_max_num, total_freq_masks_max_size // freq_mask_max_size)
+        # Add tf.print statements to track values
```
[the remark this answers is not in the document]
Max-Ryujin: Yes this was a TODO. I still need to add the log functionality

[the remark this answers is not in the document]
Max-Ryujin: But it is in the wrong function. The normal transform already got the logs in the other PR.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_sort_layer2.py
```diff
@@ -46,6 +44,22 @@ def sort_filters_by_center_freq(x):
     # move to log domain, not needed for center frequencies
     # f_resp = 20 * tf.math.log(f_resp) / tf.math.log(tf.constant(10.0))
 
+    return f_resp
+
+
+def sort_filters_by_center_freq(x):
```
vieting, line 50: Please leave the old function unchanged. If this is because of black, maybe just add the new function in `specaug_configurable.py`. I sometimes hashed the function contents and this would break them (maybe not here, but better be safe).

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_sort_layer2.py
```diff
@@ -46,6 +44,22 @@ def sort_filters_by_center_freq(x):
     # move to log domain, not needed for center frequencies
     # f_resp = 20 * tf.math.log(f_resp) / tf.math.log(tf.constant(10.0))
```
vieting, line 45: Since we do not only sort but return the actual frequency response, let's uncomment this.
```suggestion
    f_resp = 20 * tf.math.log(f_resp) / tf.math.log(tf.constant(10.0))
```

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -829,6 +829,25 @@ def run_scf_specaug():
...
                 report_args={"batch_size": "2x5k", "enable_sorting": False, "max_feature": 15},
             ),
+            "variance_based_specaug": dict(
+                returnn_args={
+                    **base_returnn_args,
+                    "specaug_config": {
+                        "enable_sorting": False,
+                        "mask_based_on_variance": True,
+                        "variance_factor": 0.5,
+                        "total_max_number_masks": 50,
```
vieting, line 839: This is less than usual, right? You told me offline that the maximum masking in frequency is 9.4% of the channels on average. This here would be less.

vieting: Also, you need to change the variable name.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +116,90 @@ def add_specaug_layer(
     # Update config with pre-generated parameters
     full_config["specaug_params"] = specaug_params
 
-    network[name] = {
-        "class": "eval",
-        "from": from_list,
-        "eval": f"self.network.get_config().typed_value('transform')(source(0, as_data=True), network=self.network, **{full_config})",
-    }
+    if not full_config["mask_based_on_variance"]:
+
```
vieting, line 120:
```suggestion
```

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +116,90 @@ def add_specaug_layer(
...
+    import tensorflow as tf
+
+    f_resp = get_frequency_response(filter_layer)
+
+    # Calculate variance for each feature
+    variance = tf.math.reduce_variance(f_resp, axis=0)
+    return variance
+
+
+def variance_based_mask(x, batch_axis, axis, variance, variance_factor, max_number_masks_for_variance_specaug):
```
vieting, line 165: In light of my previous general comment on generality: Maybe it would be better to have the probabilities per channel as input here and compute them in the transform function. Then you could just replace the transform function if you want to test another way of computing the probabilities.

vieting (changes requested):
Some more comments. I suggest to rename the new masking `filter_based_masking` and the corresponding parameters should reflect this name. Then you can create `transform_with_filter_variance` which implements this based on the filters' frequency response's variance.

Max-Ryujin:
It is somewhat more general now.  scheduling works and the masking function is independent from variance. I could also now quickly add new filter. It is still not possible to combine normal or sorted masking with filter based masking.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_sort_layer2.py
vieting: You marked this as resolved, but it is not. There are still changes in the old function.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -49,6 +49,12 @@ def add_specaug_layer(
                 freq_mask_max_proportion (float): The maximum proportion of the frequency axis that can be masked. Default is 1.
                 max_time_num_seq_len_divisor (int):
                     The divisor for the sequence length to determine the maximum number of time masks. Default is 0.7.
+                mask_based_on_variance (bool): Whether to mask based on variance or not. Default is False.
+                variance_factor (float): The factor that determines the probability of masking a feature based on variance. Default is 0.5.
+                max_number_masks_for_variance_specaug (int): The total maximum number of masks to be applied if variance-based masking is used. Default is 50.
+                variance_mask_schedule (Dict[int, float]):
```
vieting, line 55: This all still have variance in their name. As I said, I would prefer something like "filter-based".

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -71,6 +77,9 @@ def add_specaug_layer(
         "time_mask_max_proportion": 1.0,
         "freq_mask_max_proportion": 1.0,
         "max_time_num_seq_len_divisor": 0.7,
+        "mask_based_on_variance": False,
```
vieting, line 80: It could help for making this more generic to have something like `filter_based_masking_strategy` instead. By default it would be `None`, then nothing is done in this direction. If it is `"variance"`, then `transform_with_variance()` is used. Similarly, other strategies could be added.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +117,93 @@ def add_specaug_layer(
...
+            get_frequency_response,
+            calculate_frequency_response_variance,
+            _mask,
+            random_mask,
+            filter_based_masking ,
+            transform_with_variance,
+        ]
+
+
+def calculate_frequency_response_variance(filter_layer):
```
vieting, line 148: It isn't necessary to introduce this function imho. Instead of calling this function, you just have two lines that are very easy to read:
```
    f_resp = get_frequency_response(filter_layer)
    f_resp_var = tf.math.reduce_variance(f_resp, axis=0)
```

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +117,93 @@ def add_specaug_layer(
...
+
+    f_resp = get_frequency_response(filter_layer)
 
-    return [name], get_specaug_funcs()
+    # Calculate variance for each feature
+    variance = tf.math.reduce_variance(f_resp, axis=0)
+    return variance
+
+
+def filter_based_masking (x, batch_axis, axis, probability_distribution, max_number_masks_for_probability_specaug):
```
vieting, line 165:
```suggestion
def filter_based_masking(x, batch_axis, axis, probability_distribution, max_number_masks_for_probability_specaug):
```

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +117,93 @@ def add_specaug_layer(
...
+    """
+    :param tf.Tensor x: (batch,time,feature)
+    :param int batch_axis:
+    :param int axis:
+    :param int max_number_masks_for_probability_specaug:
```
vieting, line 170: `probability_distribution` missing.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +117,93 @@ def add_specaug_layer(
...
+    ndim = x.get_shape().ndims
+    n_features = tf.shape(x)[axis]
+    
+    num_masks = tf.random.uniform(
+        shape=[n_batch],
+        minval=0,
+        maxval=max_number_masks_for_probability_specaug + 1,
+        dtype=tf.int32
+    )
+    max_masks = tf.reduce_max(num_masks)
```
vieting, line 183: Why reduce max? You could have different numbers for each sequence in the batch. I think that's also the case for the baseline SpecAugment.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +117,93 @@ def add_specaug_layer(
...
+    :param int batch_axis:
+    :param int axis:
+    :param int max_number_masks_for_probability_specaug:
+    """
+    import tensorflow as tf
+    n_batch = tf.shape(x)[batch_axis]
+    ndim = x.get_shape().ndims
+    n_features = tf.shape(x)[axis]
+    
+    num_masks = tf.random.uniform(
```
vieting, line 177: I suggest you add comments with shapes for each tensor and ideally also with a short info what this does/is used for.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -201,5 +292,90 @@ def get_masked():
     return x
 
 
-def get_specaug_funcs() -> list:
-    return [sort_filters_by_center_freq, _mask, random_mask, transform]
+def transform_with_variance(data, network, **config):
```
vieting, line 295: Please check a diff between this and the existing `transform` function. It does not seem like this is already finished. Parts of the original function are missing and it is not yet possible to use a combination of the existing and the new behavior.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: I don't think this is resolved yet. Please add the shapes for each tensor.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: It is merged now

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -177,29 +243,71 @@ def get_masked():
...
+            x_masked = filter_based_masking(
+                x_masked,
+                batch_axis=data.batch_dim_axis,
+                axis=data.feature_dim_axis,
+                probability_distribution=final_probs,
+                max_number_masks_for_probability_specaug=tf.gather(
+                    config["specaug_params"]["max_number_masks_for_variance_specaug"], current_epoch
+                ),
+            )
+        else:
```
vieting, line 290: It should not be `else`. Just leave the code as it is and only add the `if` for the filter-based masking (then this if does also not contain time masking of course). If the regular masking is not desired, it should be set to 0 via the config options.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_sort_layer2.py
vieting: There is still a line removed.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -177,29 +243,71 @@ def get_masked():
...
+                            actual_time_mask_max_num,
+                            tf.shape(x)[data.time_dim_axis],
+                            variance.shape,
+                            sep=", ",
+                        )
+                    ]
+                ):
+                    return tf.identity(x_masked)
+
+            x_masked = tf.cond(enable_logging, logging_ops, lambda: tf.identity(x_masked))
```
vieting, line 270: Shouldn't this be in `filter_based_masking` where the actual mask size is determined?

vieting: What I meant was you can do a separate PR to change what currently exists. I didn't think of adding the same old behavior here. So I would add it with the actual masks directly here but I don't mind if you keep it like this for now.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +119,66 @@ def add_specaug_layer(
     # Update config with pre-generated parameters
     full_config["specaug_params"] = specaug_params
 
+
```
vieting, line 122:
```suggestion
```

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +119,66 @@ def add_specaug_layer(
...
     network[name] = {
         "class": "eval",
         "from": from_list,
         "eval": f"self.network.get_config().typed_value('transform')(source(0, as_data=True), network=self.network, **{full_config})",
     }
+    return [name], [
+        sort_filters_by_center_freq,
+        get_frequency_response,
+        filter_based_masking,
```
vieting, line 131: Don't you already have experiments running with `specaug_configurable.py`? If you now always add this function, their hashes will break.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +119,66 @@ def add_specaug_layer(
...
+    # Reshape the mask to match the input tensor shape
+    mask_shape = [tf.shape(x)[i] if i in (batch_axis, axis) else 1 for i in range(ndim)] # shape: (batch, time, feature)
+    mask = tf.reshape(mask, mask_shape)                                          # shape: (batch, time, feature)
+
+    x = tf.where(mask, tf.zeros_like(x), x)
+
+    return x
 
 
 def transform(data, network, **config):
```
vieting, line 184: I would add this new version as `transform_with_filter_variance_masking` and leave the existing `transform` function the same. Otherwise, currently/previously running experiments are not reproducible anymore.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -106,13 +119,66 @@ def add_specaug_layer(
...
+        sort_filters_by_center_freq,
+        get_frequency_response,
+        filter_based_masking,
+        _mask,
+        random_mask,
+        transform,
+    ]
+
+
+def filter_based_masking(x, batch_axis, axis, probability_distribution, max_number_masks_for_probability_specaug):
```
vieting, line 138: Did you test this sufficiently well? When I run an example with this function, it does not look like the distribution of masks follows what is given in `probability_distribution`. See the notebook that I provide you offline.

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: This bug is fixed. See commit [b2dbb28](https://github.com/rwth-i6/i6_experiments/pull/246/commits/b2dbb28875fb45a4ea2f47c94a4a58e349a8ff97)

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: This is correct, but not a big Problem, since the entire file is quite new and there are not a lot of experiments that use the new SpecAugment yet.

Max-Ryujin: Changed it

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -166,24 +181,145 @@ def filter_based_masking(x, batch_axis, axis, probability_distribution, max_numb
...
 
     x = tf.where(mask, tf.zeros_like(x), x)
 
     return x
 
 
+def get_peak_to_average_ratio(x):
+    import tensorflow as tf
+
+    x = tf.convert_to_tensor(x)
```
vieting, line 204: Is this needed?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -166,24 +181,145 @@ def filter_based_masking(x, batch_axis, axis, probability_distribution, max_numb
...
+    f_resp = get_frequency_response(x)
+
+    peak = tf.reduce_max(f_resp, axis=0)
+    average = tf.reduce_mean(f_resp, axis=0)
+    peak_to_average_ratio = peak - average
```
vieting, line 209: Do you do minus because this is in log space? I would still use a division.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -201,5 +317,129 @@ def get_masked():
...
+            ),
+            total_time_masks_max_frames // time_mask_max_size,
+        )
+        actual_freq_mask_max_num = tf.minimum(freq_mask_max_num, total_freq_masks_max_size // freq_mask_max_size)
+
+        if config["filter_based_masking_strategy"] == "variance":
+            f_resp = get_frequency_response(filter_layer)
+            variance = tf.math.reduce_variance(f_resp, axis=0)
+            n_features = tf.shape(x)[data.feature_dim_axis]
+            probs = tf.nn.softmax(variance)
```
vieting, line 386: Why do you take the softmax instead of just normalizing to unit sum?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -201,5 +317,129 @@ def get_masked():
...
+            uniform_probs = tf.ones_like(probs) / tf.cast(n_features, tf.float32)
+            final_probs = filter_factor * probs + (1 - filter_factor) * uniform_probs
+        elif config["filter_based_masking_strategy"] == "peakToAverage":
+            ratio = get_peak_to_average_ratio(filter_layer)
+            n_features = tf.shape(x)[data.feature_dim_axis]
+            probs = tf.nn.softmax(ratio)
```
vieting, line 392: See above

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: It is not, because `get_frequency_response` is already converting to a tensor. Or do you mean the entire function?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: I agree, but using the difference could be another interesting filter.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: I think the properties of softmax fit well for this use case. If a filter stands out for a high variance it should be masked more often.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: Just this line.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: Yes, you could try it in a subsequent experiment.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
vieting: Yes, it should be masked more often. But I'm afraid the softmax would flatten the distribution too much. You could do something like `probs = variance / sum(variance)` (pseudo-code). Could you please obtain the actual variance for true filters and compare the resulting `probs`?

Max-Ryujin: image
Softmax indeed looks like the wrong choice.

vieting (approved):
Looks good to me now, I think we can merge and run the experiments, analyze the results and come up with possible improvements.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -223,5 +302,152 @@ def logging_ops():
...
+                [
+                    tf.print(
+                        "Specaug Log: ",
+                        current_epoch,
+                        tf.shape(x)[data.time_dim_axis],
+                        sep=", ",
+                    )
+                ]
+            ):
+                return tf.identity(x_masked)
```
vieting, line 399: The `logging_ops` method is defined twice if I see it correctly. Also, it does not really contain that much useful information, does it?

Max-Ryujin: Yes. I defined it twice in the transform functions so that I do not have to import it into the config as well and and change more hashes. The log does not contain much information, because most info about the mask size should be reported from inside the `_mask` function where the size is known. This type of logging will be added in a future PR.

vieting: I don't understand how defining it twice here is related to hashes. Adding more detailed logging later seems alright.

Max-Ryujin: I see the problem. That was a merge artifact. I removed one of them

PR #247, opened 2024-08-26
Add logging for SpecAugment behaviour

Max-Ryujin:
Add logging to the SpecAugment Code.
It feels a bit like a hack and it fills the returnn log really quick.
TODO: Add a toggle to disable logging before finishing this PR.

### DIFF
--- a/users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
+++ b/users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
@@ -184,26 +184,23 @@ def transform(data, network, **config):
         time_upper_limit_hit = tf.equal(actual_time_mask_max_num, total_time_masks_max_frames // time_mask_max_size)
         freq_limit_hit = tf.equal(actual_freq_mask_max_num, total_freq_masks_max_size // freq_mask_max_size)
 
-        enable_logging = tf.convert_to_tensor(config.get("enable_logging", False), dtype=tf.bool)
+        enable_logging = tf.convert_to_tensor(config["enable_logging"], dtype=tf.bool)
 
         def logging_ops():
+
             with tf.control_dependencies([
-                tf.print("--------------------"),
-                tf.print("Lower limit (time):", max_time_num_seq_len),
-                tf.print("Upper limit (time):", total_time_masks_max_frames // time_mask_max_size),
-                tf.print("actual_time_mask_max_num:", actual_time_mask_max_num),
-                tf.print("actual_freq_mask_max_num:", actual_freq_mask_max_num),
-                tf.print(" The total time frames:", tf.shape(x)[data.time_dim_axis]),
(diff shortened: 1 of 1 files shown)


vieting (commented):
- In order to reduce the text in the log, I would suggest a simpler form of logging, e.g. a prefix for each line followed by same values separated by commas. This would also be easy to parse lateron.
- Also, I'm wondering what information you actually need to log. Wouldn't it be sufficient to log `max_nums` and `max_dims` for both time and features along with maybe the sequence length? Some quantities that you compute here could also be computed offline later, right?
- It would also be conceivable to log the actual mask sizes rather than the limits for the random selection, but I'm not sure if this is necessary.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -49,7 +49,8 @@ def add_specaug_layer(
                 freq_mask_max_proportion (float): The maximum proportion of the frequency axis that can be masked. Default is 1.
                 max_time_num_seq_len_divisor (int):
                     The divisor for the sequence length to determine the maximum number of time masks. Default is 0.7.
-        num_epochs (int, optional): The total number of epochs for which the training will run. default 450
+                enable_logging (bool): Whether to enable detailed logging via print statements. Defaults to False.
+        num_epochs (int, optional): The total number of epochs for which the training will run. default 450.
```
vieting, line 53: Why is the indentation inconsistent with the lines above?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -176,6 +178,29 @@ def get_masked():
...
+        enable_logging = tf.convert_to_tensor(config["enable_logging"], dtype=tf.bool)
+
+        def logging_ops():
+
+            with tf.control_dependencies([
+                tf.print(
+                "Specaug Log: ",
+                current_epoch,
+                actual_time_mask_max_num,
+                actual_freq_mask_max_num,
```
vieting, line 196: What about `time_mask_max_size` and same for frequency?

Max-Ryujin: time_mask_max_size is fixed and does not need to be logged.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: Is it? It looks fine in my editor. it should be in line with the rest of the dict entries.

vieting: Please have a look here in the PR in the browser. Maybe it's about spaces vs. tabs?

vieting (commented):
Except for the indentation, this looks alright to me. Did you test it?

vieting:
The thumbs up to my previous message seems like you did test it. Once you fixed the indentation, I will merge the PR.

vieting (approved):

PR #248, opened 2024-08-27
Allow audio conversion while creating ogg zip from LS960 bliss corpus

NeoLegends:
Adapt to https://github.com/rwth-i6/returnn/commit/3f7bd3b9c712c6b92ed7ccd8dd925a595cf73716. With the flag set to `True`, the assertion triggers.

### DIFF
--- a/common/datasets/librispeech/corpus.py
+++ b/common/datasets/librispeech/corpus.py
@@ -185,7 +185,7 @@ def get_ogg_zip_dict(
     for name, bliss_corpus in bliss_corpus_dict.items():
         ogg_zip_job = BlissToOggZipJob(
             bliss_corpus,
-            no_conversion=True,
+            no_conversion=False,
             returnn_python_exe=returnn_python_exe,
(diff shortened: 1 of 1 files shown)


NeoLegends:
Well, since this is hash-breaking, WDYT about ignoring the flag within the job, and always setting `no_conversion=False`, even if the flag is set otherwise in the job constructor?

JackTemaki:
> Should we maybe just remove this assert in RETURNN again, and leave it to the user to use no_conversion correctly?

Currently the `no_conversion` option is not usable reasonably at all any more if I interpreted the code correctly. Only with the `sprint_cache` flag set, but this is a very special case. So yes, it should be removed. If at all, the check should be for "one recording has one segment covering the whole recording". This is the case for LibriSpeech, and the reason for using this flag and avoiding converting to ogg twice.

albertz:
Basically, reduce the assert to `assert not start_time` would do this check, right? So we ignore the duration.

PR #249, opened 2024-08-30
Updates to Perturbation experiments in CTC

Max-Ryujin:
- added pitch perturbation and corresponding experiments
- added the ability to decide if the codec should be applied to segments that are not perturbed 
- added softer mu law codec experiments
- removed now unnecessary code for recognition with mu law
  -  Doing codec for recognition does not make sense with the new way of doing mu law


users/vieting/experiments/switchboard/ctc/feat/baseline_args.py
```diff
@@ -260,41 +257,6 @@ def get_returnn_config(
             feature_net["subnetwork"]["preemphasis"] = PreemphasisNetwork(alpha=alpha).get_as_subnetwork(
                 source=source_layer
             )
-        if "codecs" in audio_perturb_args:
```
vieting, line 263: Do you remove this because the assumption is that with smaller mu it should not be applied in recognition at all?

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -617,16 +617,20 @@ def run_scf_audio_perturbation_from_checkpoint():
...
         {"tempo": {"prob": 0.3, "minimum": 0.83, "maximum": 1.17}},
         {"tempo": {"prob": 0.8, "minimum": 0.83, "maximum": 1.17}},
         {"tempo": {"prob": 0.3, "minimum": 0.7, "maximum": 1.3}},
         {"tempo": {"prob": 0.8, "minimum": 0.7, "maximum": 1.3}},
         {"tempo": {"prob": 1, "minimum": 0.83, "maximum": 1.17}},
         {"tempo": {"prob": 1, "minimum": 0.7, "maximum": 1.3}},
+        {"pitch": {"prob": 0.3, "minimum": -2, "maximum": 2}},
+        {"pitch": {"prob": 0.3, "minimum": -3, "maximum": 3}},
+        {"pitch": {"prob": 0.7, "minimum": -2, "maximum": 2}},
+        {"pitch": {"prob": 0.7, "minimum": -3, "maximum": 3}},
```
vieting, line 633: Can you shortly explain how you chose the min and max values?

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -131,16 +147,19 @@ def apply_codecs(audio, sample_rate, random_state, codecs):
         for codec in codecs:
             if codec.get("encoding") == "ULAW":
                 prob = codec.pop("prob", 1.0)
-                min_value = codec.pop("minimum", 255)
-                max_value = codec.pop("maximum", 255)
+                min_value = codec.pop("minimum", 1)
+                max_value = codec.pop("maximum", 1)
```
vieting, line 151: What is the motivation for changing the default to 1?

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/baseline_args.py
Max-Ryujin: Yes, and because mu law is no longer applied to the non perturbed segments by default.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
Max-Ryujin: mostly based on listening to sound samples. Everything outside of this range sounded unrealistic. If the bigger ranges result in better WERs I can add more ranges later

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
Max-Ryujin: 1 does not change the audio and is in my opinion the expected default if no value is given.

vieting: I think it does, right?

Max-Ryujin: My bad. The change was not hear able or visible. I guess the correct value would be 0 but this is not a viable value. I am not sure what other value I should choose.

vieting: I think 255 makes sense since it is the default value used for mu-law encoding. It is maybe not a good value for us, but we can overwrite it in any case.

vieting: It seems the most obvious value to expect from a default imho.

Max-Ryujin: Just as the default for maximum or for minimum as well?
1 seems like the best option for minimum as the smallest possible change with an integer.

vieting: I would also do 255 for minimum. Because then you get the default mu-law encoding if you don't specify anything.

vieting: It won't matter too much because you'll specify the values anyway for all experiments I guess. But this seems like the most natural default.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -975,16 +979,20 @@ def run_mel_audio_perturbation_from_checkpoint():
...
 
     perturbation_args = [
-        {"codecs": [{"encoding": "ULAW", "prob": 0.3, "minimum": 250, "maximum": 260}]},
-        {"codecs": [{"encoding": "ULAW", "prob": 0.7, "minimum": 250, "maximum": 260}]},
-        {"codecs": [{"encoding": "ULAW", "prob": 0.7, "minimum": 240, "maximum": 270}]},
-        {"codecs": [{"encoding": "ULAW", "prob": 1, "minimum": 250, "maximum": 260}]},
+        {"codecs": [{"encoding": "ULAW", "prob": 0.3, "minimum": 1, "maximum": 10, "default": None}]},
+        {"codecs": [{"encoding": "ULAW", "prob": 0.7, "minimum": 1, "maximum": 10, "default": None}]},
+        {"codecs": [{"encoding": "ULAW", "prob": 0.7, "minimum": 1, "maximum": 20, "default": None}]},
+        {"codecs": [{"encoding": "ULAW", "prob": 1, "minimum": 1, "maximum": 10, "default": None}]},
```
vieting, line 985: Since the mu-law experiments are very unclear w.r.t. their outcome, maybe let's just run them for SCF first to see if it the hyperparameters are in the right ballpark and once that is the case, you can run the corresponding experiments for log Mel, too.
```suggestion
```

users/vieting/experiments/switchboard/ctc/feat/network_helpers/perturbation.py
```diff
@@ -139,8 +155,11 @@ def apply_codecs(audio, sample_rate, random_state, codecs):
                 if random_state.random() < prob:
                     mu = random_state.random() * (max_value - min_value) + min_value
                 else:
-                    # standard value for mu-law encoding
-                    mu = 255
+                    if codec.get("default") is None:
+                       continue
+                    else:
+                        # standard value for mu-law encoding
+                        mu = codec.get("default")
```
vieting, line 162: Slightly nicer imho.
```suggestion
                    # standard value for mu-law encoding
                    mu = codec.get("default")
                    if mu is None:
                       continue
```

vieting (changes requested):
Just two nitpicks, otherwise looks good to me.

vieting (approved):

PR #250, opened 2024-08-31
Adding Transducer training stage 3 for SCF

Max-Ryujin:
- added stage 3 for SCf
 - should be nearly identical to lgm stage 3 with name changes and a different `reduction-subtrahend`


vieting (commented):
Looks good to me. Did you run the training to test everything?

Max-Ryujin:
RASR path needs some updates to include all needed libraries. Right now `libpython3.8.so.1.0` is missing for multiple jobs.

vieting (commented):
Now you reduced the reduction subtrahend by 1 for the recognition on the training set, but not for the final recognition. That seems inconsistent.

Max-Ryujin:
Is there a way to make sure it is also needed for the final recognition without waiting for the training to finish?

vieting:
> Is there a way to make sure it is also needed for the final recognition without waiting for the training to finish?

Yes, but I'm not sure if the effort is needed. I'd just reduce it for the final recognition as well and if that causes issues we might further investigate.

PR #255, opened 2024-10-22
add new fairseq_pretraining function for  starting from checkpoint

AndreasPlt:
For doing scheduled hard_negatives training, I added `run_fairseq_pretraining_from_checkpoint` function that takes an additional `checkpoint` parameter and runs the pretraining job from that checkpoint

### DIFF
--- a/users/vieting/experiments/librispeech/librispeech_960_pretraining/wav2vec2/config_02_fairseq_phoneme.py
+++ b/users/vieting/experiments/librispeech/librispeech_960_pretraining/wav2vec2/config_02_fairseq_phoneme.py
@@ -134,7 +134,7 @@ def run_fairseq_pretraining_from_checkpoint(exp_name, commit, checkpoint, python
     # generate config
     fairseq_args = get_fairseq_args(num_gpus=num_gpus)
     fairseq_args["task"]["alignment"] = alignment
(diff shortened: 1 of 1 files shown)


AndreasPlt:
I noticed that this, when using the `checkpoint` parameter, it is not possible to resume training from the last checkpoint when interrupted. 

The reason for that is that, when `checkpoint` is given, fairseq always starts training from that checkpoint and not from `checkpoint_last`. On the other hand, when removing the `checkpoint` parameter in order to make fairseq train from the last checkpoint, the hash changes and `checkpoint_last` is in fact to available anymore. Any ideas, how to solve that elegantly?

vieting (changes requested):
The function `run_fairseq_pretraining_from_checkpoint` that you introduce has huge overlap with the existing `run_fairseq_pretraining`. Instead of copying the whole function and only changing a small part of it, please add `checkpoint: Optional[tk.Path] = None` as an argument to the existing function and do the modifications you need if it is given, e.g.

```
if checkpoint is not None:
    fairseq_args["checkpoint"]["continue_once"] = checkpoint
```

AndreasPlt:
Done. Instead of `tk.Path` however, I added `str` as a typehint since I am unsure if fairseq accepts a `tk.Path`. Feel free to correct me if I'm wrong

vieting (approved):

PR #262, opened 2025-03-18
Adding combination experiments in CTC

Max-Ryujin:
- Adds trainings that combine STFT-based SpecAugment and audio perturbation.


users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -1320,6 +1331,190 @@ def run_specaug_stft_experiments():
...
+                        "existing-model": {
+                            "filename": nn_system_stft_specaug.train_jobs[
+                                "conformer_bs2x5k_scf_stft20ms_fmask_5_8of512"
+                            ].out_checkpoints[24],
+                            "init_for_train": True,
+                        }
+                    },
+                },
+            },
+            feature_args={
```
vieting, line 1398: Looks like it could maybe fit in one line. Did you apply `black`?

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -1320,6 +1331,190 @@ def run_specaug_stft_experiments():
...
+        "peak_lr": 4e-4,
+        "start_lr": 1.325e-05,
+        "end_lr": 1e-5,
+        "increase_epochs": 180,
+        "decrease_epochs": 180,
+        "final_epochs": 0,
+    }
+
+    nn_base_args = {
+        "scf_bs2x5k_stft_specaug_tempo": dict(
```
vieting, line 1376: I suggest to use the same naming convention as for the experiments in `run_specaug_stft_experiments()` regarding the STFT and mask size settings.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -1298,6 +1298,17 @@ def run_specaug_stft_experiments():
...
                 report_args={"batch_size": "10k"},
             ),
+            "bs10k_scf_stft10ms_fmask_5_8of256_pre1": dict(
+                returnn_args={
+                    **returnn_args,
+                    "specaug_stft": {"max_feature": 8},
+                    "batch_size": 10000,
+                    "extra_args": {},
+                },
+                feature_args={"class": "ScfNetwork", "size_tf": 256 // 2, "stride_tf": 10 // 2, "preemphasis": 1.0},
```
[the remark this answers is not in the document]
vieting: Wouldn't this be the same? This is easier to read imho because only the different preemphasis matters here.
```suggestion
                feature_args={**feature_args_scf, "preemphasis": 1.0},
```

users/vieting/experiments/switchboard/ctc/feat/experiments.py
vieting: But anyway, I think it would be easier to have these args in a dict defined above as e.g. in `run_specaug_stft_experiments()`.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -1320,6 +1331,190 @@ def run_specaug_stft_experiments():
...
+                    **returnn_args["extra_args"],
+                    "audio_perturb_args": {"tempo": {"prob": 1, "minimum": 0.7, "maximum": 1.3}},
+                    "preload_from_files": {
+                        "existing-model": {
+                            "filename": nn_system_stft_specaug.train_jobs[
+                                "conformer_bs2x5k_scf_stft20ms_fmask_5_8of512"
+                            ].out_checkpoints[24],
+                            "init_for_train": True,
+                        }
+                    },
```
vieting, line 1395: It could maybe improve the readability if you define a helper function that creates this dict. E.g. above `nn_base_args` something like
```
def _preload_dict_helper(train_name, epoch):
    return {
        "existing-model": {
            "filename": nn_system_stft_specaug.train_jobs[train_name].out_checkpoints[epoch],
            "init_for_train": True,
        },
    }
```

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/experiments.py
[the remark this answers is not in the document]
Max-Ryujin: I added this experiment, since all other audio perturbation experiments used alpha=1. For the paper you requested to test this to make sure there is no difference. The result was the same with both alphas.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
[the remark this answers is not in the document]
Max-Ryujin: It is not relevant for the paper, but it was the final result in my Thesis.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -1320,6 +1331,160 @@ def run_specaug_stft_experiments():
...
+                    "frame_size": 400,
+                    "frame_shift": 160,
+                    "fft_size": 512,
+                },
+                "extra_args": {
+                    **returnn_args["extra_args"],
+                    "audio_perturb_args": {"tempo": {"prob": 1, "minimum": 0.7, "maximum": 1.3}},
+                    "preload_from_files": {
+                        **_preload_dict_helper("conformer_bs2x5k_scf_stft20ms_fmask_5_8of512", 24)
+                    },
```
vieting, line 1400: Wouldn't this be the same and more concise?
```suggestion
                    "preload_from_files": _preload_dict_helper("conformer_bs2x5k_scf_stft20ms_fmask_5_8of512", 24),
```

vieting (changes requested):
Looks mostly good to me. I would appreciate if you could add comments to the experiments where they were used (e.g. "final result in my thesis", "paper table X", "comparison experiment for the paper" or similar).

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -1298,6 +1298,17 @@ def run_specaug_stft_experiments():
                 lr_args=lr_args,
                 report_args={"batch_size": "10k"},
             ),
+            "bs10k_scf_stft10ms_fmask_5_8of256_pre1": dict(
```
vieting, line 1301:
```suggestion
            # tested for paper since perturbation exps used alpha=1, result same as with 0.97
            "bs10k_scf_stft10ms_fmask_5_8of256_pre1": dict(
```

vieting (approved):

PR #265, opened 2025-04-11
Adding updated sorted SpecAugment 

Max-Ryujin:
- Added the experiment described in the paper.
- Added the conservative step back to SpecAugment.  
 - It was initially removed to make the the entire scheduling customizable and remove any hard coded values.
 - I added it back late last year while investigating a convergence problem with the new SpecAugment
 - Since the paper and thesis results where archived with the conservative step included, I add it here to main.


users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -243,33 +245,48 @@ def get_masked():
...
         def logging_ops():
             with tf.control_dependencies(
                 [
                     tf.print(
-                        "Specaug Log: ",
-                        current_epoch,
+                        "step",
+                        step,
+                        "Comparison Time: ",
+                        data.time_dim_axis,
```
vieting, line 284: This is the index of the time dim and probably not what you mean, right?

vieting (changes requested):

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -200,8 +200,10 @@ def filter_based_masking(x, batch_axis, axis, probability_distribution, max_numb
 def transform(data, network, **config):
     import tensorflow as tf
 
+    conservative_step = 2000
```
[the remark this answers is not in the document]
Max-Ryujin: Yes. This was used for all experiments in the paper and is in line with the original SpecAugment implementation.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -861,6 +861,22 @@ def run_scf_specaug():
...
                 },
             ),
+            "bs10k_sorting": dict(
+                returnn_args={
+                    **base_returnn_args,
+                    "batch_size": 10000,
+                    "extra_args": {
+                        "conv_pad_seq_len_to_power": 1.5,
+                    },
+                    "specaug_config": {"enable_sorting": True, "max_feature": 15, "steps_per_epoch": 4170},
```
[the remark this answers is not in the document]
Max-Ryujin: Yes. I tuned it to match the average epoch size as close as possible.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -861,6 +861,22 @@ def run_scf_specaug():
                     "max_number_masks_for_filter_based_specaug": 75,
                 },
             ),
+            "bs10k_sorting": dict(
```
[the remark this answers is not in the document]
Max-Ryujin: Yes it was. It was not used in the paper, but I can add it here.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
```diff
@@ -243,33 +245,48 @@ def get_masked():
         )
         max_time_num_seq_len = tf.cast(
             tf.math.floordiv(
-                tf.cast(tf.shape(x)[data.time_dim_axis], tf.float32),
-                tf.cast(1.0, tf.float32) / max_time_num_seq_len_divisor * tf.cast(time_mask_max_size, tf.float32),
+                tf.cast(tf.shape(x)[data.time_dim_axis], tf.int32),
```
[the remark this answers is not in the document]
Max-Ryujin: While investigating the convergence problem last year, I changed it to ensure that it always rounds down. It was not the real problem and likely has no effect, but since all the experiments used this version, I included it in this PR.

users/vieting/experiments/switchboard/ctc/feat/network_helpers/specaug_configurable.py
Max-Ryujin: It was interesting at the time, but probably not useful for a general log message. I removed it

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -861,6 +861,36 @@ def run_scf_specaug():
                     "max_number_masks_for_filter_based_specaug": 75,
                 },
             ),
+            "bs2x5k_sorting": dict(
+                returnn_args={
+                    **base_returnn_args,
+                    "specaug_config": {"enable_sorting": True, "max_feature": 15, "steps_per_epoch": 4170},
```
vieting, line 867: Where does the 4170 come from? Above you used 4100.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
```diff
@@ -861,6 +861,36 @@ def run_scf_specaug():
...
+                },
+                feature_args=feature_args,
+                lr_args=lr_args,
+                report_args={
+                    "batch_size": "10k",
```
vieting, line 872:
```suggestion
                    "batch_size": "2x5k",
```

vieting (changes requested):
Only minor comments.

users/vieting/experiments/switchboard/ctc/feat/experiments.py
Max-Ryujin: Yes. 4100 was a bit to small leading to cases where the mask size scheduling ended before the training was done. So the train loss would drop in the last partition epoch. I can update the other trainings as well in this PR or a separate one.

Max-Ryujin: Added a comment

vieting (approved):

